PR #5, opened 2022-07-19
FLIP: Automated Slot Assignment

pgebheim:
Add FLIP describing a method of automated slot assignment. This is a DRAFT FLIP, and I am looking for feedback on the structure of the document as well as the system it describes. A POC implementation will follow and the specification may change as we proceed with discovering any engineering challenges within.


pgebheim:
@markedconfidential -- Please review this content for readability and structure. Also, a FLIP number will need to be assigned at the time when this is merged.

markedconfidential (approved):

PR #11, opened 2022-09-29
FLIP for Attachments

dsainati1:
This FLIP is intended as an alternative to https://github.com/onflow/flow/pull/1101, and should be considered in the context of the discussion on that FLIP, as well as in https://forum.onflow.org/t/extensibility/622 and https://github.com/onflow/cadence/issues/357.

Please direct discussion on this FLIP to https://forum.onflow.org/t/flip-cadence-extensions-attachments/3645

### DIFF
--- a/cadence/2022-09-21-attachments.md
+++ b/cadence/2022-09-21-attachments.md
@@ -258,6 +258,10 @@ This is backwards compatible, as it does not invalidate any existing Cadence cod
 implementing interfaces. Is this a valuable feature to have, or is it not necessary? Support for it could be added
 in the future, but would require separate discussion. 
 
+* This proposal does not include a static type to express the type of a resource with attachments, despite this existing
+in the previous proposal. This could be added later, along with a method to get a non-optional reference to an attachment
+on resources that we can statically guarantee have that attachment. 
+
 ## Alternatives
 
 In a previous [FLIP](https://github.com/onflow/flow/pull/1101), a solution to the extensibility
--- a/cadence/2022-09-21-attachments.md
+++ b/cadence/2022-09-21-attachments.md
@@ -44,7 +44,7 @@ priv attachment Bar for MyResource {
 ```
 
 Specifying the kind (struct or resource) of an attachment is not necessary, as its kind will necessarily be the same as the type it is extending. At this time,
-extensions can only be defined for a resource or struct composite type. 
+extensions can only be defined for a resource composite type. 
 
 The access modifier defines the scope in which the `attachment` can be used: a `pub attachment` can be attached to its original type anywhere that imports it, 
 while an `access(contract) attachment` can only be used within the contract that defines it. Note that this access is different than the access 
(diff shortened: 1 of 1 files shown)


dreamsmasher (approved):

cadence/2022-09-21-attachments.md
```diff
@@ -0,0 +1,286 @@
...
+    pub fun foo() {
+        super.foo()
+    }
+}
+
+```
+
+Any fields that are declared in an attachment must be initialized, just as any fields declared in a composite must be. An attachment
+that declares fields must declare an initializer, which is run when the attachment is created. The `init` function on an attachment is run after
+the attachment is attached to the base type, so `super` will have a non-`nil` value and the fields and methods of the base types that are accessible to
```
turbolent, line 88: The paragraph above states that the type of `super` is `&T`, where `T` is the extended type. Such a type is non-optional so could not be `nil` in any case

cadence/2022-09-21-attachments.md
```diff
@@ -0,0 +1,286 @@
...
+the base type will be present on that reference. 
+
+The same checks on normal initializers apply to attachment initializers; namely that all the fields declared in the attachment must receive a value in the
+extension's initializer. So, the following would be a legal attachment:
```
turbolent, line 92:
```suggestion
The same checks on normal initializers apply to attachment initializers; namely that all the fields declared in the attachment must be initialized with a value in the extension's initializer. 
So, the following would be a legal attachment:
```

cadence/2022-09-21-attachments.md
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    pub fun foo() { ... }
+}
+
+pub attachment A for R {
+    pub let derivedX: Int 
+
+    init (_ scalar: Int) {
+        self.derivedX = super.x * scalar
```
turbolent, line 76: This initializer uses `super` - is this legal?

turbolent (commented):
Looks good! It would be great to comment more on the differences to the extensions FLIP, and what the advantages and disadvantages this proposal has over it.

cadence/2022-09-21-attachments.md
dsainati1: Yea; the initializer is run when the attachment is added, so super is safe to use here.

cadence/2022-09-21-attachments.md
```diff
@@ -0,0 +1,286 @@
...
+        self.y = y
+    }
+}
+
+let sa = attach A(y: 3) to S(x: "foo")
+```
+
+### Removing Attachments from a Type
+
+Attachments can be removed with a new statement: `remove t from e`. The `t` value here is a type name, rather than a value,
```
joshuahannan, line 176: Can a composite type have multiple attachments of the same type?

joshuahannan (commented):
This looks good to me! I like it better than extensions too. I just had one question about multiple attachments of the same type

cadence/2022-09-21-attachments.md
dsainati1: Ah this is a good point; implicitly it would have to be no since the attachments are reference/looked-up by type.

joshuahannan: I feel like that is going to be something that people will want.

dsainati1: Hmm, in any situation where someone would want to have two different attachments of type `T` on a resource, they could also make a `TManager` type attachment that contains a `[T]` field, and achieve the same effect with a little indirection.

joshuahannan: my prediction is that most attachments will probably want to support multiple per resource, and I think it would just be better for the get attachments method to just return an array of attachments of the requested type, even if it is only a one-element array. I think requiring there be a wrapper type around multiple attachments of the same type just makes things more difficult, because then if you want to find a specific attachment for a resource, you can't just request its type any more. you also have to know the type of the wrapper

dsainati1: Hmm, I think I am perhaps not following; can you give me an example of where resources would want to support multiple copies of the same attachment?

Keep in mind that attachments cannot be removed from resources without destroying them; i.e. they are not objects in and of themselves. In this iteration of the proposal, for example, the `Hat` would not be an attachment for the `Kitty` because the `Hat` should be an independently valuable resource that can be bought and sold independently of the `Kitty` itself. Hence, the attachment for `Kitty` would be a `HatManager` that keeps track of all the `Hat`s present on that `Kitty`, while the `Hat`s are just normal resources.

joshuahannan: okay, that makes more sense

PR #14, opened 2022-10-12
Avoid oppressive terminology

turbolent:


pgebheim (approved):
Thanks for catching this!

PR #40, opened 2022-10-24
FLIP: interface inheritance in Cadence

SupunS:
Discussion: https://forum.onflow.org/t/flip-interface-inheritance-in-cadence/3750

Reference implementation: https://github.com/onflow/cadence/pull/2112


cadence/2022-10-24-interface-inheritance.md
```diff
@@ -0,0 +1,342 @@
...
+    // Error: `deposit` function has a different signature compared to the `Receiver.deposit`.
+    // So these two cannot co-exist.
+    pub fun deposit()
+}
+```
+
+#### Functions with conditions
+
+If the two conflicting functions with identical signatures have pre/post conditions, then it will still be valid.
+However, the pre/post conditions would be linearized (TBD - refer [questions and discussion topics section](#questions-and-discussion-topics))
```
dsainati1, line 165: Will conditions be handled differently than default implementations? I am curious how these two features will interact in cases of multiple inheritance; is it possible in a diamond situation for one interface to provide a default implementation and for another to provide the conditions? Or would this be considered an ambiguity and cause a static error?

cadence/2022-10-24-interface-inheritance.md
```diff
@@ -0,0 +1,342 @@
...
+Otherwise, interface conformance is not valid.
+
+```cadence
+pub resource interface Receiver {
+    pub fun id: Int
+}
+
+pub resource interface Vault: Receiver {
+    // `id` field has a different type than the `Receiver.id`. Hence this is invalid.
+    pub fun id: UInt64
```
dsainati1, line 126: This is consistent with how concrete types conform to interfaces; subtyping is not permitted even though it is technically safe statically. Perhaps this is something we can add in the future? It wouldn't be a breaking change.

cadence/2022-10-24-interface-inheritance.md
```diff
@@ -0,0 +1,342 @@
...
+}
+```
+
+#### Functions
+
+If the two conflicting functions have identical signatures then it will be valid.
```
dsainati1, line 132:
```suggestion
If the two identically-named functions have identical signatures then it will be valid.
```

Confusing to describe valid functions as "conflicting"

cadence/2022-10-24-interface-inheritance.md
```diff
@@ -0,0 +1,342 @@
...
+```
+
+Many other languages that support multiple-inheritance also follow similar approaches, where inheritances that
+cause ambiguities are rejected by the compiler (see: https://en.wikipedia.org/wiki/Multiple_inheritance).
+
+### Drawbacks
+
+One drawback is, as mentioned in the above section, when there are complex inheritance chains with default functions
+being available from multiple paths, overriding a default function by an interface in the middle of the chain
+can lead to breaking the downstream codes that depend on this interface.
```
dsainati1, line 295: I would argue worse even than breaking downstream code is the potential to change what a piece of code is doing at runtime by changing an interface inheritance chain. Given some chain `A -> B -> C` where `A` provides a default implementation of `foo`, `B` is an interface conforming to `A`, and `C` is an interface conforming to `B`, if `B` adds an overriding implementation to `foo`, this would cause `A`'s implementation to no longer be used by `C`. This means that the author of `B`'s contract can change the behavior of `C`, even though it may not be the author of `C`'s contract. This feels like a big safety concern to me, to the point where I feel we should simply disallow overriding default implementations entirely in order to avoid it.

dsainati1 (commented):
Nice! Thanks for writing this up. This is a really thorny design space so we should take care to iron out all the edge cases here. I had some specific questions below, but also some general ones:

1) How will this new feature interact with the proposed changes in https://github.com/onflow/cadence/issues/2069 or https://github.com/onflow/cadence/issues/2081 ? If events can be "inherited" from interfaces, how do they resolve in cases of multiple inheritance?

2) I think this change would also benefit from a redundancy check in any places where a list of interfaces might appear. In a conformance list, for example, it would be redundant to say `MyVault: Receiver, Vault`, so we should have a static error warning that the `Receiver` annotation is unnecessary in the presence of the `Vault` annotation. The same would go for restricted types' restriction lists.

cadence/2022-10-24-interface-inheritance.md
SupunS: Yes, conditions and default functions behave differently - this is explained in the "Default functions" section.

In cases where a default function exists in one path and a condition in the other path, both rules should be applied; the one and only default implementation would be used (it's an error if there are more), with all conditions linearized. I'll add a section for this too.

cadence/2022-10-24-interface-inheritance.md
SupunS: It might be a problem. e.g: Imagine a case where one inherited interface (say, the parent interface is `T`) narrows down the type of a field, and another inherited interface from the same parent type `T` has a condition/default function which uses the same field in its original type (broader type). Then that could lead to some type-safety issues.

SupunS:
> How will this new feature interact with the proposed changes in https://github.com/onflow/cadence/issues/2069 or https://github.com/onflow/cadence/issues/2081 ? If events can be "inherited" from interfaces, how do they resolve in cases of multiple inheritance?

Thanks for bringing this up! I was thinking the type definitions would also behaves similarly to the default functions. Added a section for this: https://github.com/onflow/flips/pull/40/commits/d98d658f60cfa5b20436426e604e49e331eb7c7e

SupunS:
> I think this change would also benefit from a redundancy check in any places where a list of interfaces might appear. In a conformance list, for example, it would be redundant to say MyVault: Receiver, Vault, so we should have a static error warning that the Receiver annotation is unnecessary in the presence of the Vault annotation. The same would go for restricted types' restriction lists.

Yes definitely, we could add a redundancy check for conformances. Having redundant conformances can sometimes cause static errors by the current rules (i.e: when there are overridden default functions, etc.). Others can be checked at a linter phase.

cadence/2022-10-24-interface-inheritance.md
SupunS: Added it here: 6ff896c83bb7a58c77e62d5060ed5ee146b32bb4

Although, this is not how the concrete type conformance works at the moment. It is a static error there in similar situations. But I think this needed to be relaxed for the interfaces to make it less-annoying / less-breaking. We can also revisit the struct conformance later, if we are ok with the suggested behaviour.

cadence/2022-10-24-interface-inheritance.md
SupunS: I would say, that's a kind of 'trust' a developer would opt in when conforming to an interface - without overriding. That could happen even now, with no inheritance, where the author of the interface can change the default implementation later, which would change the concrete type's behaviour.

cadence/2022-10-24-interface-inheritance.md
```diff
@@ -0,0 +1,449 @@
...
+    // `id` field has same type as the `Receiver.id`. Hence this is valid.
+    pub fun id: UInt64
+}
+```
+
+Otherwise, interface conformance is not valid.
+
+```cadence
+pub resource interface Receiver {
+    pub fun id: Int
```
bluesign, line 121: s/fun/var

cadence/2022-10-24-interface-inheritance.md
bluesign: I agree with @dsainati1 here a bit, default interface implementation involves one on one trust with interface and developer, but now we introduce a web of trust. ( also I think this will bring a lot of fragile stuff )

SupunS:
We had a breakout session regarding the diamond problem today, and rejecting the default function conflicts (as proposed in this FLIP) was favored by the majority.

A detailed summary of the meeting is as below.

- Reject ambiguous cases (proposed in the FLIP)
    - Follows the same approach as the current default functions implementation.
    - Concern:
        - Someone higher up the chain can break downstream code by introducing a default implementation.
    - Conclusion: Accepted solution  ✅ 
- Cadence to pick one, maybe the one closest to the current implementation.
    - Concerns:
        - This may result in unintended/surprising behaviors.
        - Already disregarded this option for default functions ambiguity resolution in concrete type implementations. (They also behave very similarly to what is proposed in the above FLIP)
    - Conclusion: Too "magical"  👎 
- Users to “lock in” to an implementation, optionally.
    - e.g:
        
        ```swift
        pub resource MyVault: Vault {
            pub inherited<Receiver> fun log(_ message: String)
        }
        ```
        
    - Concern:
        - This is optional. Doesn’t guarantee developers will use it. Their code could break if they don’t, same as the original FLIP proposal.
        - Cannot mandate this ‘locking-in’. Because it would contradict the original intention of default functions - which is to update interfaces (add methods) without breaking the downstream code.
   - Conclusion: Doesn't really solve the problem  👎

SupunS:
Updated the FLIP with the latest discussion outcomes (7127f6b198dd85a097de4867d3806faed745da74).

Also added a new section on how the interfaces/conditions would be linearized (9005d7f7bd699fdcb53645252cbeb88f3d9d3e2e). TL;DR is, conditions are linearized in a depth-first pre-order manner, ignoring any duplicates (duplicates are cases where there are two or more paths to the same interface/condition).

cadence/2022-10-24-interface-inheritance.md
SupunS: A section for linearization is added here: 9005d7f7bd699fdcb53645252cbeb88f3d9d3e2e

SupunS:
Reference implementation: https://github.com/onflow/cadence/pull/2112

cadence/2022-10-24-interface-inheritance.md
```diff
@@ -0,0 +1,553 @@
...
+    pub fun test() {
+        let foo: Foo  // This will refer to the `NonFungibleToken.Foo`
+    }
+}
+```
+
+If a user needed to access the `Foo` struct coming from the super interface `Token`, then they can
+access it using the fully qualified name. e.g: `let foo: Token.Foo`.
+
+However, it is not allowed to have two or more inherited type/events definitions with identical names for an interface.
```
turbolent, line 333: Why is this forbidden?

cadence/2022-10-24-interface-inheritance.md
```diff
@@ -0,0 +1,553 @@
...
+// interface is visible to the `NonFungibleToken` via the `Collectible` interface.
+// Another implementation of `Foo` struct is visible to the `NonFungibleToken` via the `Token` interface.
+// This creates ambiguity.
+pub resource interface NonFungibleToken: Token, Provider {}
+```
+
+Many other languages that support multiple-inheritance also follow similar approaches, where inheritances that
+cause ambiguities are rejected by the compiler (see: https://en.wikipedia.org/wiki/Multiple_inheritance).
+
+### Linearizing Conditions
```
turbolent, line 392: Nice!

turbolent (approved):
Great proposal!

cadence/2022-10-24-interface-inheritance.md
SupunS: This is because of the type requirements. If there are types with identical names, the concrete type that implements these two interfaces might not be able to satisfy them both. So here it is taking the most restricted approach by rejecting such conflicts.

However, this can be relaxed after the type requirements are removed.

cadence/2022-10-24-interface-inheritance.md
```diff
@@ -0,0 +1,553 @@
...
+
+When an interface implements another interface, it is possible for the two interfaces to have members
+(functions, fields, and type requirements) with the same name.
+It is important to resolve these ambiguities at the time of interface declaration (i.e: as early as possible),
+rather than waiting for a concrete type declaration to face them when implementing that particular interface.
+The following sections explain how to resolve these ambiguities for different scenarios.
+
+#### Fields
+
+If two fields with identical-names have identical types, then it will be valid.
```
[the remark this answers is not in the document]
SupunS: Good idea! added with 38d5c9a2cadb0f9b5e554e7a925bdb42f5e5fb9b

cadence/2022-10-24-interface-inheritance.md
```diff
@@ -0,0 +1,567 @@
...
+pub resource MyVault: Vault {
+
+    pub fun withdraw(_ amount: Int): @Vault {}
+
+    pub fun deposit(_ something: @AnyResource) {}
+}
+```
+
+The concrete type `MyVault` would not need to explicitly specify the `Receiver` interface conformance.
+By conforming to the `Vault` interface, it is implicitly required to also conform to the `Receiver` interface.
```
joshuahannan, line 83: I still kind of like the idea of requiring the concrete type to declare the implementations of all the interfaces just to make it explicit. Any specific reason we shouldn't do that?

joshuahannan (approved):
I'm pretty happy with this! Just had one comment

cadence/2022-10-24-interface-inheritance.md
SupunS: I initially felt that would be sort of redundant. e.g.: many existing languages with inheritance also follow a similar approach. But that's just my personal opinion.
I understand the reason behind the suggestion and the further I think about it, made me think that's actually a really good idea.

We have the LDM scheduled for next Wednesday (3rd May), maybe we can bring this up again there, and hopefully finalize it. Sorry for not bringing that up in the last meeting, That had gone completely out of my radar!

SupunS: As we discussed in the LDM, I am in favor of this suggestion. Would love to hear what everyone else thinks.

If we hear no objections, I will add this to the FLIP.

bluesign: what is the benefit in this @SupunS ? From that I can see, if `Vault` adds another interface to conform, we break `MyVault`. 

Let's consider Vault decided to conform to Provider interface ( which let's say has a pre/post condition only ) with default implementation. ( which we are allowing/ will allow now ) It will break MyVault.

SupunS: Oh, that's a really good point! 

The only benefit of this suggestion is that it'll improve the readability. But now with the scenario you mentioned, seems it might be too restrictive.

joshuahannan: thats a reasonable concern, but I don't think that'll be very common, so I'm not too worried about it. I still like requiring the implementing type to be more specific and mention all the interfaces, but I understand if y'all think that is too restrictive

SupunS:
Thanks everyone for the suggestions and feedback for this proposal so far!

This FLIP has been open for a while, and it seems we are all happy with the current proposal, and seems there are no any remaining concerns.
Thus, I'm going to consider this FLIP as accepted 🎉  🎉  🎉 

You can follow how the implementation is progressing here ➡️ https://github.com/onflow/cadence/pull/2112

PR #43, opened 2022-10-27
FLIP: Change the syntax for function types

dreamsmasher:
Work towards https://github.com/onflow/cadence/issues/2094

### DIFF
--- a/cadence/2022-10-18-change-fun-type-syntax.md
+++ b/cadence/2022-10-18-change-fun-type-syntax.md
@@ -58,27 +58,56 @@ Given that `fun` is already a reserved keyword for declaring functions, requirin
 
     ```cadence
     // (AnyStruct -> Void) -> AnyStruct -> Void
-    fun call(f: ((AnyStruct): Void), x: AnyStruct) {...}
+    fun apply(f: ((AnyStruct): Void), x: AnyStruct) {...}
 
     // after
-    fun call(f: (fun(AnyStruct): Void), x: AnyStruct) {...}
+    fun apply(f: (fun(AnyStruct): Void), x: AnyStruct) {...}
     ```
 
 This could be implemented as a non-breaking change by still allowing the old syntax of omitting `fun`, but preferring the keyword in new contracts and transactions. The stringification of function types should be updated to use the new syntax.
 
-### Drawbacks
-
-As @turbolent mentioned in an internal discussion, there is some ambiguity in the language grammar regarding the distinction between functions and restricted types. This is what motivated the choice to use the existing syntax without the `fun` keyword.
-
-For example,
+For type signatures ascribed to variables, we currently rely on the colon `:` token in order to parse the type as a function. This requires the type to be wrapped up in parentheses to avoid parsing ambiguity. Reusing a previous example:
 
 ```cadence
-fun(): T{}
+let bar = fun(x: Int): String {...}
+
+// a reference to a function
+let baz: ((Int): String) = foo
 ```
 
-Is this a function that returns a `T` but has an empty body? Or does it return a restricted type `T{}` and is missing its body?
+If we continued to elide the `fun` keyword in function types, parentheses could still be eliminated by making `:` a right-associative operator with a higher precedence than `=`. 
 
-One potential route is to give functions a higher priority in the parser than restricted types, so the above expression is parsed as a function returning  `T` with an empty body. Let's take a more complex example:
+```ebnf
+func-type = '(' [type {, type}] ')' ':' type
+type = func-type | identifier | ...
+assignment = identifier [':' type] '=' expression
+```
+
+At the same time however, how do we express a function that returns `Void` (`()`) with this syntax? We can omit a return type in function declarations, which causes the type to default to `Void`, but referencing the function requires one to explicitly name `Void`:
+
+```cadence
+fun noop() {}
+
+let noop_: ((): Void) = noop
+```
+
+Using the `fun` keyword would simplify parsing rules and also allow for return type elision on procedures:
+
+```cadence
+let noop_: fun() = noop // unambiguous, since `fun` is our marker now instead of `:`
+
+let baz: fun(Int): String = bar
+```
+
+The adjusted grammar would only affect `func-type`:
+
+```ebnf
+func-type = 'fun' '(' [type {, type}] ')' [':' type]
+```
+
+### Drawbacks
+
+To the best of my knowledge, this change should not introduce any ambiguities into the language grammar. There is currently an ambiguous case for function expressions that return restricted types:
 
 ```cadence
 interface Foo {...}
@@ -86,18 +115,18 @@ interface Foo {...}
 let f = fun(): AnyStruct{Foo}
 ```
 
-In this case, `f` can be parsed either as a function returning `AnyStruct` whose body only contains the expression `Foo`, or as a function returning a restricted type `AnyStruct{Foo}` but is missing a body. Assume that we prefer to follow the first parsing rule. Since the "body" of the function is missing an explicit `return`, its return type is inferred to be `Void`. However, the stray `Foo` exists in the type namespace instead of the value namespace, so it will result in an unknown name error unless a variable with the same name exists in scope. A lint could be added that ensures statements are either function calls, assignments, declarations, or moves. For instance,
+In this case, it is unclear to the parser whether `f` is a function that returns a restricted `AnyStruct{Foo}` without a body, or if it returns an unrestricted `AnyStruct` and whose body is the single expression `Foo`. We currently step around this issue by requiring restricted types' curly brackets to follow immediately after the parent type without any whitespace:
 
-```
-var x = 0 // declaration, ok
-fun f(): Int {log("hi"); return 42} // declaration, ok
-x = f() // assignment, ok
-f() // function call with unused return, ok
-x = x + 1 // ok
-x // noop, should throw a lint error
+```cadence
+fun(): AnyStruct{Foo} // return type is a restricted type, no body
+fun(): AnyStruct {Foo} // return type is AnyStruct, body is Foo
 ```
 
-Additionally, if the user intends to write a restricted type with multiple interfaces like `AnyStruct{Foo, Bar}`, the parser will throw an error at the use of bare commas outside of a function call or composite literal. It's akin to writing an expression `Foo, Bar`. It seems that the only cases where this ambiguity doesn't throw an error is when the unrestricted return type is parsed as a supertype of `Void`, and the restricting interface is shadowed by a value. 
+This ambiguity does not extend to types however, and is outside the scope of this FLIP. Using the proposed syntax, we could still write
+
+```cadence 
+let f: fun(): AnyStruct{Foo} = fun(): AnyStruct{Foo} {...}
+```
 
 ### Alternatives Considered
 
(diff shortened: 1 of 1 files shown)


cadence/2022-10-18-change-fun-type-syntax.md
```diff
@@ -0,0 +1,204 @@
...
+// function declaration
+fun foo(_ x: Int): String {...}
+// function expression bound to a name
+let bar = fun(x: Int): String {...}
+
+// a reference to a function
+let baz: ((Int): String) = foo
+```
+
+This convention for representing functions gets even more confusing when currying or higher-order functions are brought into the mix:
```
dsainati1, line 27: Currying is not a common design pattern in Cadence, not worth referring to it here. 

```suggestion
This convention for representing functions gets even more confusing when higher-order functions are brought into the mix:
```

cadence/2022-10-18-change-fun-type-syntax.md
```diff
@@ -0,0 +1,204 @@
...
+f() // function call with unused return, ok
+x = x + 1 // ok
+x // noop, should throw a lint error
+```
+
+Additionally, if the user intends to write a restricted type with multiple interfaces like `AnyStruct{Foo, Bar}`, the parser will throw an error at the use of bare commas outside of a function call or composite literal. It's akin to writing an expression `Foo, Bar`. It seems that the only cases where this ambiguity doesn't throw an error is when the unrestricted return type is parsed as a supertype of `Void`, and the restricting interface is shadowed by a value. 
+
+### Alternatives Considered
+
+An alternative to using `fun` to denote function types is to introduce another keyword, such as `=>` or `->`. This would be more familiar to users coming from languages such as Javascript or Haskell, while remaining as a non-breaking change due to these operators being unreserved currently. For example,
```
dsainati1, line 104: To be honest, when I was looking at the syntax for function types in TypeScript, Swift and Kotlin (all of which Cadence explicitly cites as inspiration for its syntax and behavior: https://developers.flow.com/cadence/language/index#syntax-and-behavior), these languages all use `->` for their function types. Given that prepending `fun` to the type as it exists in Cadence today causes a syntax ambiguity, I actually think it might be a better change to use the `->` syntax instead to separate the parameters from the return type. 

I don't think the mismatch between how functions are declared and how their types are expressed is really that big of an issue. Kotlin, for example, uses `->` in its function types despite using declaration syntax that does not use this symbol. 

```suggestion
An alternative to using `fun` to denote function types is to introduce another keyword, such as `=>` or `->`. This would be more familiar to users coming from languages such as JavaScript, Swift, and Kotlin, while remaining as a non-breaking change due to these operators being unreserved currently. For example, 
```

cadence/2022-10-18-change-fun-type-syntax.md
dreamsmasher: I should open a separate FLIP to allow curried functions like in scala:

```cadence
fun add3(_ x: Int)(_ y: Int)(_ z: Int): Int {
    return x + y + z
}

let plus5 = add3(1)(4)
```

cadence/2022-10-18-change-fun-type-syntax.md
```diff
@@ -0,0 +1,204 @@
...
+One potential route is to give functions a higher priority in the parser than restricted types, so the above expression is parsed as a function returning  `T` with an empty body. Let's take a more complex example:
+
+```cadence
+interface Foo {...}
+
+let f = fun(): AnyStruct{Foo}
+```
+
+In this case, `f` can be parsed either as a function returning `AnyStruct` whose body only contains the expression `Foo`, or as a function returning a restricted type `AnyStruct{Foo}` but is missing a body. Assume that we prefer to follow the first parsing rule. Since the "body" of the function is missing an explicit `return`, its return type is inferred to be `Void`. However, the stray `Foo` exists in the type namespace instead of the value namespace, so it will result in an unknown name error unless a variable with the same name exists in scope. A lint could be added that ensures statements are either function calls, assignments, declarations, or moves. For instance,
+
```
SupunS, line 90: This section describes the ambiguity in function expressions. Is it also the case for function types?
If so, maybe adjust this section to describe the ambiguity in function type.

cadence/2022-10-18-change-fun-type-syntax.md
```diff
@@ -0,0 +1,204 @@
...
+What related issues do you consider out of scope for this proposal,
+but could be addressed independently in the future?
+
+If we use a syntax other than `fun` for denoting function types, such as an arrow `->`, the new keyword could be used to implement a shorthand for declaring closures.
+
+## Prior Art
+
+Most languages with first-class functions use similar language for referring to those functions. For instance,
+
```
SupunS, line 171: Go has a similar syntax to what is proposed  here:
```go
var f func(int) int = ...
```
Would be good to add that too as an example.

cadence/2022-10-18-change-fun-type-syntax.md
```diff
@@ -0,0 +1,204 @@
...
+This could be implemented as a non-breaking change by still allowing the old syntax of omitting `fun`, but preferring the keyword in new contracts and transactions. The stringification of function types should be updated to use the new syntax.
+
+### Drawbacks
+
+As @turbolent mentioned in an internal discussion, there is some ambiguity in the language grammar regarding the distinction between functions and restricted types. This is what motivated the choice to use the existing syntax without the `fun` keyword.
+
+For example,
+
+```cadence
+fun(): T{}
```
SupunS, line 76: related to my other comment regarding ambiguity: This is not a valid place for a 'type'. i.e: a declaration/expression may not start with a type.
Maybe update it with an example where `fun(): T{}` is placed in a context where a type is expected. I feel there won't be any ambiguity there.

cadence/2022-10-18-change-fun-type-syntax.md
dreamsmasher: actually wait, this ambiguity only comes into play for expressions. function types themselves don't have a body, so I can probably just cut this section out entirely

SupunS: nice! maybe add it as a remark, saying the ambiguity is not present for function type 👌

cadence/2022-10-18-change-fun-type-syntax.md
```diff
@@ -0,0 +1,251 @@
...
+
+## Motivation
+
+Currently, functions are declared and referred to with differing syntax. For instance,
+
+```cadence
+// function declaration
+fun foo(_ x: Int): String {...}
+// function expression bound to a name
+let bar = fun(x: Int): String {...}
```
turbolent, line 21: A space between the keyword and the parameter list is unidiomatic for function expressions. Maybe remove here and for examples of the new function type syntax

```suggestion
let bar = fun (x: Int): String {...}
```

cadence/2022-10-18-change-fun-type-syntax.md
```diff
@@ -0,0 +1,251 @@
...
+
+// a reference to a function
+let baz: ((Int): String) = foo
```
turbolent, line 24: Maybe explicitly mention that function types have a different syntax from function declarations and function expressions

cadence/2022-10-18-change-fun-type-syntax.md
```diff
@@ -0,0 +1,251 @@
...
+Rename function types to require the `fun` keyword.
+
+Given that `fun` is already a reserved keyword for declaring functions, requiring the keyword in the type signature of a function is more consistent with the existing syntax. For instance, 
+
+    ```cadence
+    // (AnyStruct -> Void) -> AnyStruct -> Void
+    fun apply(f: ((AnyStruct): Void), x: AnyStruct) {...}
+
+    // after
+    fun apply(f: (fun(AnyStruct): Void), x: AnyStruct) {...}
```
turbolent, line 64: For example, here. Also, the parentheses should not be necessary anymore, right?

```suggestion
    fun apply(f: fun (AnyStruct): Void, x: AnyStruct) {...}
```

cadence/2022-10-18-change-fun-type-syntax.md
```diff
@@ -0,0 +1,251 @@
...
+```cadence
+let noop_: fun() = noop // unambiguous, since `fun` is our marker now instead of `:`
+
+let baz: fun(Int): String = bar
+```
+
+The adjusted grammar would only affect `func-type`:
+
+```ebnf
+func-type = 'fun' '(' [type {, type}] ')' [':' type]
```
turbolent, line 105: This should use the same syntax for rules as the grammar uses in https://github.com/onflow/cadence/blob/master/docs/cadence.ebnf, as well as the same rule names. 

Also, the parameter and result types are not just plain types, but type annotations (`type` -> `typeAnnotation`).

Finally, it looks like this makes the return type optional. Do we want that?

cadence/2022-10-18-change-fun-type-syntax.md
```diff
@@ -0,0 +1,251 @@
...
+let f: fun() -> AnyStruct{Foo} = fun() -> AnyStruct{Foo} 
+```
+
+### Performance Implications
+
+Function types written with either syntax will create identical parse trees, so the performance impact is expected to be limited to another conditional branch in the parser.
+
+### Engineering Impact
+
+* Do you expect changes to binary size / build time / test times?
```
turbolent, line 165: Remove the points from the template here and in the other sections

turbolent (commented):
Nice! Some details need to be addressed, but overall this will be an improvement. 
+1 from me

cadence/2022-10-18-change-fun-type-syntax.md
dreamsmasher: sorry, do you mean idiomatic? as in we suggest authors to insert a space?

cadence/2022-10-18-change-fun-type-syntax.md
dreamsmasher: I added a bit of discussion on omitting the return types. I'd argue in favour of it, especially if we go with a syntax like `fun(T) -> U` or `fun(T): U` that's prefixed by a keyword. Languages like Go and Rust keep this omission rule consistent between function declarations and types.

cadence/2022-10-18-change-fun-type-syntax.md
```diff
@@ -0,0 +1,295 @@
...
+
+Currently, functions are declared and referred to with differing syntax. For instance,
+
+```cadence
+// function declaration
+fun foo(_ x: Int): String {...}
+// function expression bound to a name
+let bar = fun (x: Int): String {...}
+
+// a reference to a function
```
turbolent, line 23: "Reference" is potentially confusing here. Lets only use it when actual references are involved. 
Maybe "a variable declaration with a function value"?

bjartek:
I would really love to see the arrow keyword. if we go with => or -> is the same for me.

cadence/2022-10-18-change-fun-type-syntax.md
dreamsmasher: ah good point

dreamsmasher:
> I would really love to see the arrow keyword. if we go with => or -> is the same for me.
> 
> If we have that though how would the following code look like?
> 
> ```
> let dictionary = {"one": 1, "two": 2, "three": 3}
> 
> dictionary.forEachKey(fun (key: String): Bool {
> 
>     let value = dictionary[key]
> 
>     log(key)
> 
>     log(value)
> 
>     
> 
>     return key != "two" // stop iteration if this returns false
> 
> })
> ```
> 
> Would it be
> 
> ```
> let dictionary = {"one": 1, "two": 2, "three": 3}
> 
> dictionary.forEachKey( key: String -> Bool{
> 
>     let value = dictionary[key]
> 
>     log(key)
> 
>     log(value)
> 
>     
> 
>     return key != "two" // stop iteration if this returns false
> 
> })
> ```

If we were to introduce an infix notation for function expressions, then the parentheses around the argument would likely need to remain, unless we commit to allowing type inference in those cases.

```cadence
dictionary.forEachKey((key: String) -> Bool {...}
```

Assuming we kept the `fun` keyword, then the parentheses should definitely remain consistent with function declarations.

```cadence
dictionary.forEachKey(fun (key: String): Bool {...})
```

Swift has an interesting feature where the keyword `in` can be used to declare closures with type inference for arguments and return values:

```swift
let reversedNames = names.sorted(by: { s1, s2 in s1 > s2 } )
```

but I'm not a fan of introducing multiple ways to declare a lambda expression. I think that extending type inference to work on closure arguments/returns is worth a FLIP on its own, but it's a bit outside the scope of this proposal.

turbolent:
We had intentionally decided to not use the arrow syntax `->` for return types, as this is syntactically too close to the arrows being used for a different concept, resources, where we use `<-`. I still think that was a good decision.

SupunS (approved):
I'm very much in favour of this proposal, as it makes the syntax cleaner/verbose, and unifies it with function declarations and function expressions.

cadence/2022-10-18-change-fun-type-syntax.md
turbolent: Agreed!

turbolent (approved):
👍

dsainati1 (approved):

bjartek:
Some good points here. I approve!

cadence/2022-10-18-change-fun-type-syntax.md
```diff
@@ -0,0 +1,295 @@
+# Change the syntax for function types.
+
+| Status        | (Proposed / Rejected / Accepted / Implemented)       |
```
turbolent, line 3:
```suggestion
| Status        |  Accepted                                            |
```

turbolent (commented):
Accepted 🎉

PR #47, opened 2022-11-02
Update attachments FLIP

dsainati1:
* Explicitly mention that attachment access can occur on references, and returns a non-auth reference.

* Mention that attempting to modify a composite's attachments while iterating over it is an error. 

* Explain that `base` is not populated with a new attachment `A` until `A`'s constructor finishes executing during `attach`

* Rename `super` to `base`.

### DIFF
--- a/cadence/2022-09-21-attachments.md
+++ b/cadence/2022-09-21-attachments.md
@@ -260,6 +260,8 @@ pub resource R {
 }
 ```
 
+Attempting to attach new attachments to `R` or remove attachments from `R` while iterating over it is a runtime error. 
+
 ### Drawbacks
 
 Adding a new language feature has the downside of complexity: users have to learn yet another 
--- a/cadence/2022-09-21-attachments.md
+++ b/cadence/2022-09-21-attachments.md
@@ -191,7 +191,10 @@ first to guarantee that the value does not have that attachment before they atte
 The `attach` expression creates a new value, rather than modifying the old one; thus if the base is a struct, the result of the expression will have the attachment,
 while the original value will not. In the case of a resource, the old resource will be moved into the new one, which will have the attachment present. It is also worth
 noting that while `super` does point to the base during the attachment's constructor, that base does not yet have the attachment present on it during the execution
-of that constructor; that only occurs after the attachment expresion successfully executes. 
+of that constructor; that only occurs after the attachment expresion successfully executes. To see why, consider what would happen if a user accessed that attachment 
+through super; if in `A`'s initializer they wrote `super[A]!.foo?`. This could potentially allow `A`'s `foo` field to be accessed before it has been initialized, and
+would require analysis of statement ordering within the initializer itself to determine when it was safe to access which fields of `A` through super. Rather than add
+this additional complexity, we simply do not attach `A` to the base until after the initializer has finished executing. 
 
 ### Removing Attachments from a Type
 
--- a/cadence/2022-09-21-attachments.md
+++ b/cadence/2022-09-21-attachments.md
@@ -188,6 +188,11 @@ If a users attempts to `attach` an attachment to a value that already possesses 
 reason, if a user has reason to believe that an attachment of the type they wish to add already exists on a value, they should use the `remove` statement 
 first to guarantee that the value does not have that attachment before they attempt to attach it again.
 
+The `attach` expression creates a new value, rather than modifying the old one; thus if the base is a struct, the result of the expression will have the attachment,
+while the original value will not. In the case of a resource, the old resource will be moved into the new one, which will have the attachment present. It is also worth
+noting that while `super` does point to the base during the attachment's constructor, that base does not yet have the attachment present on it during the execution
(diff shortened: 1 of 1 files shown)


dsainati1:
@turbolent is this what you had in mind for making the reference behavior for access more explicit in the FLIP?

cadence/2022-09-21-attachments.md
```diff
@@ -188,6 +188,11 @@ If a users attempts to `attach` an attachment to a value that already possesses
 reason, if a user has reason to believe that an attachment of the type they wish to add already exists on a value, they should use the `remove` statement 
 first to guarantee that the value does not have that attachment before they attempt to attach it again.
 
+The `attach` expression creates a new value, rather than modifying the old one; thus if the base is a struct, the result of the expression will have the attachment,
```
turbolent, line 191: It might be surprising that `attach` on a struct copies, while other common "mutating"-named operations (e.g. appending to a struct-kinded dictionary) actually mutate. 

Maybe attaching to a struct-kinded value should be mutating and not return anything? 

Just thinking out loud, not necessarily saying this is better.

turbolent:
@dsainati1 Looks good! 

The question that came up was: If I own a value, attaching and accessing attachments is safe – I own the value. If I want to share limited access to the value, I can use a restricted reference and share it – this is the common access control pattern (for example, I can share a `FT.Vault` with you and only allow you to read the balance by restricting the reference to the `FT.Balance` interface).

How can I restrict access to the attachments though?

dsainati1:
Mmm good point. I see two potential options: 

1) Attachments can only be accessed from `auth` references. This way you can give someone an `auth` reference if you want them to be able to use attachments on your resource.

2) Attachments cannot be accessed on interfaces; you would need to give someone a reference directly to your resource's concrete type if you want to allow them to access its attachments.

Which do you prefer?

cadence/2022-09-21-attachments.md
```diff
@@ -188,6 +188,11 @@ If a users attempts to `attach` an attachment to a value that already possesses
...
+while the original value will not. In the case of a resource, the old resource will be moved into the new one, which will have the attachment present. It is also worth
+noting that while `super` does point to the base during the attachment's constructor, that base does not yet have the attachment present on it during the execution
+of that constructor; that only occurs after the attachment expresion successfully executes.
```
dreamsmasher, line 194: Oh interesting, is this a consequence of the lazy `super`-binding we discussed during the implementation sync?

dreamsmasher (approved):

cadence/2022-09-21-attachments.md
dsainati1: No; this was a deliberate decision. It would make no sense for the attachment to be present on the `super` object before the constructor has finished executing. Otherwise we'd have to consider what would happen if you accessed that attachment through super; i.e. in `A`'s initializer what if you did `super[A]!.foo`? What if `foo` has not yet been initialized? We do not want to deal with these questions, so we just prevent them by populating `A` on `super` after `A`'s constructor has finished executing.

turbolent: Please add this reasoning to the proposal itself, it's great

cadence/2022-09-21-attachments.md
dsainati1: Yea this is an interesting point. The syntax and semantics for the attachment operations were designed with resources in mind, since that is likely to be the common case, but yea it has some unexpected behavior for structs. The reason I went with a non-mutating expression for `attach`, rather than a statement that modifies in place was to parallel `create`, with the idea that `attach` serves as a constructor-like operation for attachments the way that a `create` expression is for a resource (the same reasoning explains why `remove` is a statement rather than an expression: to parallel `destroy`). 

This also has the nice benefit that one can create an attachment and a resource in the same expression; consider the following code example in which `attach` is a mutating statement:

```
let r <- create R()
attach A() to R()
foo(<-r)
```
vs. how it works in the proposal right now:
```
foo(<-attach A() to <-create R())
```

That said, I think there's also a valid argument to be made that `attach` is actually a parallel operation to a field assignment, given that we've discussed elsewhere how attachments function like implicit fields. In this mental model, `attach A() to r` is less like `create` and more like an assignment of the form `r.A = A()`; in which case it would make more sense for `attach` to be a statement. 

I think there are benefits to both approaches; curious to get other thoughts on this though.

cadence/2022-09-21-attachments.md
```diff
@@ -203,14 +211,24 @@ If a resource containing attachments is `destroy`ed, all its attachments will be
...
+of a resource can restrict which attachments can be accessed on references to their resource using restricted types, much like they would do with any other field or function. E.g.
+
+```
+struct R: I {}
+struct interface I {}
+attachment A for R {}
+fun foo(r: &{I}) {
+    r[A] // fails to type check, because `{I}` is not a subtype of `R`
+}
+```
```
dsainati1, line 230: @turbolent I added some language here to talk explicitly about how restricted types interact with attachments.

PR #55, opened 2022-12-19
FLIP 55: Fungible Token Standard Version 2

joshuahannan:
Proposes a fundamental upgrade to the Flow Fungible Token standard.

Original Proposal here: https://forum.onflow.org/t/streamlined-token-standards-proposal/3075/1
Code PR: https://github.com/onflow/flow-ft/pull/131

### DIFF
--- a/application/20221219-ft-v2.md
+++ b/application/20221219-ft-v2.md
@@ -3,7 +3,7 @@ status: Implemented
 flip: 55
 authors: Joshua Hannan (joshua.hannan@flowfoundation.org)
 sponsor: Joshua Hannan (joshua.hannan@flowfoundation.org) 
-updated: 2024-02-20 
+updated: 2024-03-11 
 ---
 
 # Fungible Token Standard V2
--- a/application/20221219-ft-v2.md
+++ b/application/20221219-ft-v2.md
@@ -204,7 +204,11 @@ so there is no performance requirements to enforce with the methods.
 
 ### Dependencies
 
-* Adds no new dependencies
+* Adds the `Burner` contract.
+  * Contracts won't have to import the `Burner` contract
+    directly because the conformance is handled by
+    `FungibleToken.Vault`, so it isn't a direct dependency,
+    but important to be aware of
 * Dependent projects
     * All fungible tokens on Flow and some projects that utilize them,
     including but not limited to:
@@ -227,7 +231,8 @@ within resources, avoiding public fields, and giving developers flexibility to w
 
 ### Tutorials and Examples
 
-* Coming soon once the proposal reaches a more final state
+* Check out the Cadence 1.0 migration guide for instructions on how to update contracts to the new standards:
+  * https://cadence-lang.org/docs/cadence_migration_guide/
 
 ### Compatibility
 
@@ -250,7 +255,7 @@ to create scoped providers.
 
 This feature is out of the scope of this proposal, but should definitely be a standard
 that lives alongside the main fungible token standard.
-We hope to shepard a proposal for these soon.
+We hope to shepherd a proposal for these soon.
 
 ## Prior Art
 
--- a/application/20221219-ft-v2.md
+++ b/application/20221219-ft-v2.md
@@ -1,0 +1,257 @@
+---
+status: draft 
+flip: NNN (do not set)
+authors: Joshua Hannan (joshua.hannan@dapperlabs.com)
+sponsor: Joshua Hannan (joshua.hannan@dapperlabs.com) 
+updated: 2022-12-19 
+---
+
+# Fungible Token Standard V2
+
+## Objective
+
+This FLIP proposes multiple updates to the Flow Fungible Token Standard contracts,
+primarily about encapsulating functionality within token resources instead of
+with the contracts. There are other smaller quality of life changes to the token standard,
+such as integration of metadata, improving error handling,
+and including a transfer method and interface.
+Some of these changes are dependent on other Cadence FLIPs being approved,
+primarily [interface inheritance](https://github.com/onflow/flips/pull/40),
+removal of nested type requirements,
+and [allowing interfaces to emit events](https://github.com/onflow/cadence/issues/2069).
+
+Most of the changes proposed here would be breaking for all fungible token implementations
+on the Flow blockchain, but should not be for third-party integrations such as
+event listeners and apps that interface with the contracts.
+
+## Motivation
+
+The current fungible token standard for Flow
+was designed in mid 2019, at a time when Cadence itself was still being designed.
+The current standard, though functional, leaves much to be desired.
+
+The current token standard uses contract interfaces.
+They are designed in a way which requires each concrete contract
+to provide exactly one Vault type.
+This means that any project that needs multiple tokens must deploy multiple contracts.
+In the case of very simple tokens, this is a lot of complexity for very little value.
+
+Related to this problem, functionality and metadata associated with some tokens,
+such as paths, events, and empty vault creation methods,
+is only accessible directly through the contract itself, when it should also
+be accessible directly through an instance of a token resource and/or interface.
+
+In the case of events, currently there is no way for the standard to ensure that
+implementations are emitting standardized and correct events, which this upgrade will address.
+
+Additionally, the usage of nested type requirements creates some confusing
+interactions, which are not useful and make learning about
+and interacting with token smart contracts more difficult.
+
+## User Benefit
+
+With these upgrades, users will now be able to:
+* Define multiple tokens in a single contract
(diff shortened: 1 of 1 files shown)


joshuahannan:
We're having another working group meeting on Nov 7th to discuss the token standard improvements! We're hoping to finalize most of the changes so we can more confidently say what will be part of the release. You can see the event on the [Flow public event calendar.](https://calendar.google.com/calendar/embed?src=c_47978f5cd9da636cadc6b8473102b5092c1a865dd010558393ecb7f9fd0c9ad0%40group.calendar.google.com&ctz=America%2FChicago)

joshuahannan:
We had another discussion today about the token standards. We made a few decisions:

* We will not be including a default destruction event for `Vault`. Instead we are adding a global function on `FungibleToken` that can be called to burn a `Vault`. This will only emit a destruction event if the balance of the vault is greater than zero
* We decided to include methods in the default events to say the UUID of all the vaults involved. For example, withdraw will look like this now:
```
emit Withdraw(amount: amount, type: self.getType().identifier, from: self.owner?.address, fromUUID: self.uuid, withdrawnUUID: result.uuid)
```
* We will add the `Balance` interface back to make migration simpler.
* We had a discussion about how we should manage the `balance` field for `Vault` and the various interfaces. We mostly thought that including `balance` in the `Vault` interface was a good idea, but we disagreed on what additional balance-related methods should be included in the standard, particularly on the `Provider` and `Balance` interfaces. Options:
        * `getWithdrawableBalance(): UFix64`: it would return the amount of tokens that are actually available
            to be withdrawn from the container. This would keep the standard relatively simple, 
            but might not help all use-cases because the balance might be too complicated to manage with this method
        * `isAmountAbleToBeWithdrawn(amount: UFix64): Bool`: Would return a boolean to indicate if the requested amount
           can be withdrawn from the container. This could support more complicated use cases

I am of the opinion that just `getWIthdrawableBalance` is enough, but it has not been decided yet. We'll discuss again at the next meeting at 9am Pacific on Thurs Jan 11.

Proposed Agenda: 

* Should we include balance and id as fields in {NFT} and {Vault}, respectively?
* Should we have methods to getWithdrawableBalance() or isAvailableToWithdraw(): Bool in the vault, provider, and balance interfaces?
* Should the token standards be contract interfaces to enforce ViewResolver and createEmptyVault()?
* How to name entitlements
* should ViewResolver.resolveView be renamed to resolveContractView?

joshuahannan:
We had another discussion about the token standards today. here is a summary of what we talked about:

* removing getter functions for standard paths and only including them in metadata views
    * They’re accessible as metadata views so it might make sense to have that be the only way to get them in a standardized way, simplifies the standard
    *** DECISION: 
        * Remove getter functions for standard paths. 
        * remove provider linked path and provider type from VaultData view**

* Should the token standards be contract interfaces 
    * If they are interfaces, we can enforce conformance to ViewResolver and createEmptyVault
    * BUT, we would lose the ability to have standard events and utility functions like burn
    * **DECISION: Keep as a contract and find other ways to enforce that people are conforming to ViewResolver**

* Incompatible upgrades that are part of the new standards
    * Let people know that we are going to have a conversation with the cadence team soon to figure out how to migrate all these weird changes
    * Might be able to include new interfaces that are for legacy tokens that implement the new ones to make migration easier
    * **DECISION: Have a conversation with the Cadence team next week during the next one of these meetings about what the migration difficulties are that need to be reverted or addressed**
    
    
Topics for next week:

* Should we have balance as a field in Vault?
    * Should we have methods to getWithdrawableBalance() or isAvailableToWithdraw(): Bool in the vault, provider, and balance interfaces?
* How to name entitlements
    * Adjectives, Verbs?
    * Introduce three entitlement with Owner and have a mapping that maps to both withdraw and update?

joshuahannan:
We had another discussion about the token standards today. We focused on migration difficulties with the Cadence team. You can find notes [here](https://github.com/onflow/flips/pull/56#issuecomment-1894530961)

Next week, we'll be discussing entitlement naming and default implementations

joshuahannan:
We had another discussion today that, pending final approval of the code, finalizes the design of the standards!
Here is a summary of the decisions we made:

* Contract vs interface
    * We will make both standards interfaces now
    * Events that are emitted from implementations do not have the same type, so they are safe
    * Implementations will have to implement a getTypes method and a createEmpty method that takes a type as an argument
    * We will also propose new views for contracts that return information about token types in contracts
* How to name entitlements
    * nouns are used for "coarse-grained" entitlements (e.g. Storage to access all of the account storage functionality), and verbs are used for "fine-grained" entitlements, like specific functions (e.g. SaveValue to be able to store a value into account storage)
    * Rename entitlements to Withdraw, rename events to Withdrawn and Deposited
    * Add Owner entitlement to NFT standard to map to withdraw and update
* Default implementations
    * Remove default implementations for all view resolver methods
    * Remove them from every other method that makes sense
    
    
  We'll have another chat next Thursday to finish everything up

joshuahannan:
We had our last design meeting for the new token standards yesterday and settled on a design! Here are notes from the meeting that are relevant to the FT standard:

* Decided to not try to have a conditional that emits events only if `to` or `from` is `nil`. It wasn't possible to include this cleanly in a contract interface and the benefit isn't enough to warrant putting them in another contract
* Added the `Burner` contract, which allows resources to specify a method that is called when it is destroyed, but ONLY if you destroy it through the `Burner.burn` method. `FungibleToken.Vault` now inherits from `Burner.Burnable` because it needs the callback to emit the standard Burned event. Token implementations can also implement the `bunCallback()` to update their tokens total supply when tokens are burned, but MUST set the balance of the vault to zero before the end of the function execution. This is to prevent spam.
* Removed `getBalance()` and added `isAvailableToWithdraw()`.  It can be more performant because it doesn’t necessarily need to iterate through all the balances to know the answer. `getBalance()` was also redundant because we had re-added the balance field back to the standard.

The code in [the PR](https://github.com/onflow/flow-ft/pull/131) is now the finalized code for the new standard and will be released in the upcoming CLI version and likely will be the version that is deployed to all the networks unless a critical issue is surfaced in the meantime.

Thank you to everyone who contributed! The standards are looking quite good. 😎

joshuahannan:
Looking for some approvals on this FLIP so we can get it merged since it all has been resolved and implemented for a while now!

bjartek (approved):

application/20221219-ft-v2.md
```diff
@@ -0,0 +1,262 @@
...
+      and have the `createEmptyVault(vaultType: Type)` function.
+
+### Performance Implications
+
+All of the methods in the fungible token interface are expected to be O(1),
+so there is no performance requirements to enforce with the methods.
+
+### Dependencies
+
+* Adds no new dependencies
```
sisyphusSmiling, line 207: Would we consider the `Burner` contract a new dependency for FT or does that not matter as far as downstream projects are concerned?

application/20221219-ft-v2.md
```diff
@@ -0,0 +1,262 @@
...
+* The code can be tested on its own once a compatible version of cadence is released. 
+
+### Best Practices
+
+* Some of the changes illustrate Cadence best practices, such as encapsulating functionality
+within resources, avoiding public fields, and giving developers flexibility to write composable code.
+
+### Tutorials and Examples
+
+* Coming soon once the proposal reaches a more final state
```
sisyphusSmiling, line 230: Could also link the existing migration docs

application/20221219-ft-v2.md
```diff
@@ -0,0 +1,262 @@
...
+
+A critical piece of tooling for fungible tokens would be a struct that contains
+a provider capability but restricts the capability to only be able to withdraw
+a specified amount of tokens from the underlying vault.
+Currently, providers have no limit, but all tokens should be able
+to create scoped providers.
+
+This feature is out of the scope of this proposal, but should definitely be a standard
+that lives alongside the main fungible token standard.
+We hope to shepard a proposal for these soon.
```
sisyphusSmiling, line 253:
```suggestion
We hope to shepherd a proposal for these soon.
```

sisyphusSmiling (approved):
Left some minor comments otherwise LGTM! 🚀

application/20221219-ft-v2.md
joshuahannan: Contracts won't have to import the `Burner` contract directly because the conformance is handled by `FungibleToken.Vault`, but I guess it is still kind of a dependency, so I'll add a description of it here.

austinkline (approved):

franklywatson (approved):

joshuahannan:
Thanks for the reviews and approvals everybody! I've addressed all your comments and I think we have enough approvals now that I'm okay to merge. 🥳 ❤️

PR #56, opened 2022-12-19
FLIP 56: Non-Fungible Token Standard Version 2

joshuahannan:
Proposes a fundamental upgrade to the Flow Fungible Token standard.

Original Proposal here: https://forum.onflow.org/t/streamlined-token-standards-proposal/3075/1
Code PR: https://github.com/onflow/flow-nft/pull/126

### DIFF
--- a/application/20221219-nft-v2.md
+++ b/application/20221219-nft-v2.md
@@ -1,9 +1,9 @@
 ---
 status: Implemented 
 flip: 56
-authors: Joshua Hannan (joshua.hannan@dapperlabs.com)
-sponsor: Joshua Hannan (joshua.hannan@dapperlabs.com) 
-updated: 2024-02-20 
+authors: Joshua Hannan (joshua.hannan@flowfoundation.org)
+sponsor: Joshua Hannan (joshua.hannan@flowfoundation.org) 
+updated: 2024-03-11 
 ---
 
 # Non-Fungible Token Standard V2
@@ -100,8 +100,8 @@ access(all) contract interface ViewResolver {
 
     /// Function that resolves a metadata view for this token.
     /// Some contracts may have multiple resource types that support metadata views
-    /// so there there is an optional parameter for specify which resource type the caller
-    /// is looking for views for.
+    /// so there there is an optional parameter to specify which resource type the caller
+    /// is requesting views for.
     /// Some contract-level views may be type-agnostic. In that case, the contract
     /// should return the same views regardless of what type is passed in.
     ///
@@ -171,7 +171,7 @@ it is a best practice to have getter functions return `nil` if something goes wr
 instead of panicking and reverting the transaction.
 
 ```cadence
-pub fun borrowNFT(id: UInt64): &AnyResource{NFT}?
+access(all) view fun borrowNFT(_ id: UInt64): &{NFT}?
 ```
 
 ### Remove the requirement for the `ownedNFTs` field in `Collection`
@@ -182,14 +182,14 @@ to store tokens in a unique way that is different what what could be specified.
 This proposal removes `ownedNFTs` and only requires a `getIDs()` method
 for the collection to indicate which IDs it contains.
 
-### Move `createEmptyCollection()` to inside the Collection definition in addition to the contract
+### Move `createEmptyCollection()` to inside the NFT and Collection definitions in addition to the contract
 
 It is useful to be able to create a new empty collection
-directly from a `Collection` object instead of having to import
+directly from an `NFT` or a `Collection` object instead of having to import
 the contract and call the method from the contract.
 
 ```cadence
-pub resource interface Collection {
+access(all) resource interface Collection {
 
     /// createEmptyCollection creates an empty Collection
     /// and returns it to the caller so that they can own NFTs
@@ -206,8 +206,9 @@ pub resource interface Collection {
 
 Metadata Views for non-fungible tokens should be easily accessible
 from interfaces defined by the standard. 
-This propsal enforces that implementations implement the `ViewResolver` contract interface
-and `NFT` implementations implement the `ViewResolver.Resolver` interface.
+This proposal enforces that implementations implement the `ViewResolver` contract interface,
+`NFT` implementations implement the `ViewResolver.Resolver` interface,
+and `Collection` implementations implement the `ViewResolver.ResolverCollection` interface.
 
 ### Add support for getting information about Sub-NFTs
 
@@ -325,7 +326,8 @@ within resources, avoiding public fields, and giving developers flexibility to w
 
 ### Tutorials and Examples
 
-* Coming soon once the proposal reaches a more final state.
+* Check out the Cadence 1.0 migration guide for instructions on how to update contracts to the new standards:
+  * https://cadence-lang.org/docs/cadence_migration_guide/
 
 ### Compatibility
 
@@ -348,7 +350,7 @@ to create scoped providers.
 
 This feature is out of the scope of this proposal, but should definitely be a standard
 that lives alongside the main fungible token standard.
-We hope to shepard a proposal for these soon.
+We hope to shepherd a proposal for these soon.
 
 ### Universal Collection
 
--- a/application/20221219-nft-v2.md
+++ b/application/20221219-nft-v2.md
@@ -56,7 +56,7 @@ implementations are emitting standardized and correct events, which this upgrade
 With these upgrades, users will now be able to:
 * Define multiple tokens in a single contract.
 * Query all the data about a token directly through the contract as well as the `Vault` resource and core interfaces.
-* Have standard events emitted correctly for important operations (`Withdrawn`, `Deposited`, `Updated`, and `NonFungibleToken.NFT.ResourceDestroyed`)
+* Have standard events emitted correctly for important operations (`Withdrawn`, `Deposited`, `Updated`, and `NonFungibleToken.NFT.ResourceDestroyed`).
 * Support querying information about NFTs that an NFT directly owns.
 * Check if a `Receiver` accepts a given token type.
 
@@ -87,7 +87,7 @@ access(all) contract interface ViewResolver {
 
     /// Function that returns all the Metadata Views implemented by the resolving contract.
     /// Some contracts may have multiple resource types that support metadata views
-    /// so there there is an optional parameter for specify which resource type the caller
+    /// so there is an optional parameter for specify which resource type the caller
     /// is looking for views for.
     /// Some contract-level views may be type-agnostic. In that case, the contract
     /// should return the same views regardless of what type is passed in.
@@ -168,7 +168,7 @@ access(all) resource interface Receiver {
 In the current standard, `borrowNFT()` is meant to panic if the NFT to borrow
 is not found in the collection. This is clearly the wrong choice, because 
 it is a best practice to have getter functions return `nil` if something goes wrong
-instead of panicing and reverting the transaction.
+instead of panicking and reverting the transaction.
 
 ```cadence
 pub fun borrowNFT(id: UInt64): &AnyResource{NFT}?
--- a/application/20221219-nft-v2.md
+++ b/application/20221219-nft-v2.md
@@ -3,26 +3,30 @@ status: draft
 flip: NNN (do not set)
 authors: Joshua Hannan (joshua.hannan@dapperlabs.com)
 sponsor: Joshua Hannan (joshua.hannan@dapperlabs.com) 
-updated: 2022-12-19 
+updated: 2024-01-26 
 ---
 
 # Non-Fungible Token Standard V2
 
 ## Objective
 
-This FLIP proposes multiple updates to the Flow Non-Fungible Token Standard contracts,
-primarily about encapsulating functionality within token resources instead of
-with the contracts. There are other smaller quality of life changes to the token standard,
-such as integration of metadata, improving error handling,
-and including a transfer method and interface.
-Some of these changes are dependent on other Cadence FLIPs being approved,
+This FLIP proposes multiple updates to the Flow Non-Fungible Token
+Standard contracts as part of the Cadence 1.0 upgrade,
+primarily about adding support for defining multiple token types
+in one contract, adding standard events, integrating `ViewResolver`,
+adding support for updating NFTs, adding support for sub-NFTs, and adding entitlements
+
+Some of these changes are dependent on other Cadence FLIPs that
+made fundamental changes to the language
+that have been approved and implemented,
 primarily [interface inheritance](https://github.com/onflow/flips/pull/40),
 removal of nested type requirements,
+[removal of custom destructors](https://github.com/onflow/flips/blob/main/cadence/20230811-destructor-removal.md),
 and [allowing interfaces to emit events](https://github.com/onflow/cadence/issues/2069).
 
-Most of the changes proposed here would be breaking for all non-fungible token implementations
-on the Flow blockchain, but should not be for third-party integrations such as
-event listeners and apps that interface with the contracts.
+The changes proposed here will be breaking for all non-fungible token implementations
+on the Flow blockchain, as well as contracts that utilize tokens based on the standard,
+third-party integrations such as event listeners, and apps that interface with the contracts.
 
 ## Motivation
 
@@ -30,30 +34,31 @@ The current non-fungible token standard for Flow
 was designed in mid 2019, at a time when Cadence itself was still being designed.
 The current standard, though functional, leaves much to be desired.
 
-The current token standard uses contract interfaces.
-They are designed in a way which requires each concrete contract
-to provide exactly one `NFT` type and one `Collection` type.
+The current token standard uses contract interfaces and nested type requirements.
+They are designed in a way that requires each concrete contract
+to provide exactly one `NFT` and one `Collection` type.
 This means that any project that needs multiple tokens must deploy multiple contracts.
 In the case of very simple tokens, this is a lot of complexity for very little value.
 
 Related to this problem, functionality and metadata associated with some tokens,
-such as paths, events, and empty collection creation methods,
+such as paths and empty vault creation methods
 is only accessible directly through the contract itself, when it should also
 be accessible directly through an instance of a token resource and/or interface.
+
+Many contracts also do not implement MetadataViews properly because
+there is no requirement for it in the standard.
 
 In the case of events, currently there is no way for the standard to ensure that
 implementations are emitting standardized and correct events, which this upgrade will address.
 
-Additionally, the usage of nested type requirements creates some confusing
-interactions, which are not useful and make learning about
-and interacting with token smart contracts more difficult.
-
 ## User Benefit
 
 With these upgrades, users will now be able to:
-* Define multiple tokens in a single contract
-* Query all the data about a token directly through the token resource and core interfaces
-* Have standard events emitted correctly for important operations
+* Define multiple tokens in a single contract.
+* Query all the data about a token directly through the contract as well as the `Vault` resource and core interfaces.
+* Have standard events emitted correctly for important operations (`Withdrawn`, `Deposited`, `Updated`, and `NonFungibleToken.NFT.ResourceDestroyed`)
+* Support querying information about NFTs that an NFT directly owns.
+* Check if a `Receiver` accepts a given token type.
 
 ## Design Proposal
 
@@ -63,77 +68,79 @@ A [pull request with the suggested code changes](https://github.com/onflow/flow-
 is in the flow fungible token github repository.
 
 The main code changes and their implications are described here.
-The linked proposals provide more context.
+The linked proposals provide more context, though the proposal in the forum post
+is quite out of date.
+
+### Add support for multiple token types in a single contract
+
+Now that nested type requirements have been removed, contracts can define
+multiple token types. The type information can be accessed by using the built-in
+`contract.publicTypes(): [Type]` method. A return value of this is now required
+in the `NonFungibleToken.createEmptyCollection(nftType: Type)` method because
+the caller needs to specify which type they are looking for.
+
+`ViewResolver` has also been updated to require type information
+when querying views from a contract:
+
+```cadence
+access(all) contract interface ViewResolver {
+
+    /// Function that returns all the Metadata Views implemented by the resolving contract.
+    /// Some contracts may have multiple resource types that support metadata views
+    /// so there there is an optional parameter for specify which resource type the caller
+    /// is looking for views for.
+    /// Some contract-level views may be type-agnostic. In that case, the contract
+    /// should return the same views regardless of what type is passed in.
+    ///
+    /// @param resourceType: An optional resource type to return views for
+    /// @return An array of Types defining the implemented views. This value will be used by
+    ///         developers to know which parameter to pass to the resolveView() method.
+    ///
+    access(all) view fun getContractViews(resourceType: Type?): [Type]
+
+    /// Function that resolves a metadata view for this token.
+    /// Some contracts may have multiple resource types that support metadata views
+    /// so there there is an optional parameter for specify which resource type the caller
+    /// is looking for views for.
+    /// Some contract-level views may be type-agnostic. In that case, the contract
+    /// should return the same views regardless of what type is passed in.
+    ///
+    /// @param resourceType: An optional resource type to return views for
+    /// @param view: The Type of the desired view.
+    /// @return A structure representing the requested view.
+    ///
+    access(all) fun resolveContractView(resourceType: Type?, viewType: Type): AnyStruct?
+}
+```
+
+While all contracts that implement the `NonFungibleToken` standard
+will need to implement these functions, many of them will still
+only have a single token type in their contract.
+Therefore, they can ignore the `resourceType` parameter.
+
+In addition to these changes to `ViewResolver`, the `MetadataViews.Resolver`
+resource interface has been moved to the `ViewResolver` contract because
+this is the logical place for it to live and it also avoids a circular dependency.
 
 ### Move Event Definitions and emissions to resource interfaces
 
 Instead of requiring events to be defined in the token implementations contracts,
-they will only be defined in the non-fungible token standard smart contract and will
-be emitted in post-conditions from the correct methods
-in the resource interfaces defined in the standard.
-
-This feature is still being designed, so there is not a code sample yet.
-See https://github.com/onflow/cadence/issues/2069 for potential examples
+they will only be [defined in the non-fungible token standard smart contract](https://github.com/onflow/flow-nft/blob/standard-v2/contracts/NonFungibleToken.cdc#L86) and will
+be [emitted in post-conditions](https://github.com/onflow/flow-nft/blob/standard-v2/contracts/NonFungibleToken.cdc#L185)
+from the correct methods in the resource interfaces defined in the standard.
 
 ### Add Type and Metadata parameters to events
 
 Standard events contain more information about the NFT that is being transferred,
-such as the type of the NFT and important metadata about the NFT.
+such as the type of the NFT and important metadata about the transfer.
+This includes UUIDs of the collections involved in the token movements
+in case this information is useful.
 
-Here is a simple example of what this could look like:
-
-```cadence
-pub event Deposit(id: UInt64, to: Address?, type: Type, displayView: MetadataViews.Display?, serialView: MetadataViews.Serial?)
-```
-Or a more involved example: (from [find](https://f.dnz.dev/4fd2a0ccdc3a6cdb2774872b4ac5b32c167e60ae5850ba2b34ef4a749433fe4e)):
-
-```
-from: 0x46625f59708ec2f8
-fromName: null
-to: 0x1eef2d5df9d6dd7a
-toName: null
-uuid: "122520380"
-nftInfo:
-  id: "122520380"
-  name: "Alphonso Davies - CH01P01 #308"
-  thumbnail: "ipfs://bafybeidfyk5boobk3n767gz3dc3sma2do6b43k7qsu3s6dlob4h7mmnb2y"
-  type: "A.46625f59708ec2f8.AeraPanel3.NFT"
-  rarity: "Common"
-  editionNumber: null
-  totalInEdition: null
-  scalars:
-    panel_id: "1.00000000"
-    player_number: "19.00000000"
-    edition_set_max: "455.00000000"
-    chapter_id: "1.00000000"
-    slot: "1.00000000"
-    chapter_index: "1.00000000"
-    edition_set_number: "307.00000000"
-    player_id: "23540953.00000000"
-  tags:
-    player_jersey_name: "Davies"
-    panel_description: "Alphonso Davies’ parents, Victoria and Debeah Davies, decide to escape the horrors of civil war by gathering all their possessions and leaving Liberia with their children in search of a better life."
-    player_position: "LB"
-    player_nationality: "Canada"
-  collectionName: "Aera"
-  collectionDescription: "Aera by OneFootball"
-context:
-  tenant: "onefootball"
-  challenge: "test1"
-remark: null
-```
-
-### Include Transferable Interface with transfer method
-
-For managing simple transfers, tokens can now implement the transferable interface
-and the transfer method.
+Here is an example of the proposal:
 
 ```cadence
-    pub resource interface Transferable {
-        /// Function for a direct transfer instead of having to do a deposit and withdrawal
-        /// Returns Bool to indicate if the transfer succeeded or not
-        pub fun transfer(id: UInt64, receiver: Capability<&AnyResource{Receiver}>): Bool
-    }
+/// The event that is emitted when tokens are withdrawn from a Vault
+access(all) event Withdrawn(type: String, amount: UFix64, from: Address?, fromUUID: UInt64, withdrawnUUID: UInt64)
 ```
 
 ### Add `getAcceptedTypes()` method to `Receiver` interface
@@ -142,15 +149,18 @@ It is useful to be able to query a non-fungible token receiver to see what types
 it can accept.
 
 ```cadence
-    pub resource interface Receiver {
+access(all) resource interface Receiver {
+    /// getSupportedNFTTypes returns a list of NFT types that this receiver accepts
+    /// @return A dictionary of types mapped to booleans indicating if this
+    ///         reciever supports it
+    access(all) view fun getSupportedNFTTypes(): {Type: Bool}
 
-        /// deposit takes an NFT as an argument and adds it to the Collection
-        ///
-        pub fun deposit(token: @AnyResource{NFT})
-
-        /// getAcceptedTypes returns a list of NFT types that this receiver accepts
-        pub fun getAcceptedTypes(): {Type: Bool}
-    }
+    /// Returns whether or not the given type is accepted by the collection
+    /// A collection that can accept any type should just return true by default
+    /// @param type: An NFT type
+    /// @return A boolean indicating if this receiver can recieve the desired NFT type
+    access(all) view fun isSupportedNFTType(type: Type): Bool
+}
 ```
 
 ### Change the return type of `borrowNFT()` to an optional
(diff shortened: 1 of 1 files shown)


bjartek:
I have been thinking about this quite a lot recently and I have some comments around this. I do not think this NFTv2 suggestion takes the changes far enough. I belive the community could have more benefit if we do even more.

1. Having multiple NFT types in the same slot is not a good idea IMHO. (A slot here is defined as storage+links for a given Colleciton). Large dictionaries eat up gas. Iterating over things and loading lots of registers does not make more sence. 


2. Use the same Collection implementation to store multiple NFT types, but _not_ in the same slot. You do not have to rewrite the code to create a collection you only have to know what limitations this NFT type has on the collection it can be put into. 

3. Have an NFT of a certain type be storable under different slots of a collection with different classifiers. Standarize the paths. A path to a slot should contain 

- the identifier of the collection
- the identifier of the type
- a subcollection or a classifier or maybe even both?


Lets take Flovatar as an example. They have their own Collection implementation for FlovatarComponents since it borrows a specific type to run some checks on it. so the start of a slot path would be

A.abcd.FlovatarComponent.Collection/A.abcd.FlovatarComponent.NFT

Then flovatar have multiple types of components and they would fit perfectly into the classifier/subcomponent. We have

- sparks
- boosters
- flobot
- hats
- equipment
- background
- eyeglasses

This would fit very nicely into Dieters vision about your wallet as your home folder. I can have a folder for my flovatar components seperated into subcollections on a subpath and then using classifiers to maybe seperate my hats into "trendy" or "to-be-burned" paths.

4. Remove as much boilerplate as possible from an NFT contract. 

What I think we can do to fix this:

- make it possible to create a NFT Collection that can store multiple NFT types. The slot that stores a single type has information about what that type is. We have a method that returns `what type is stored here` in the collection
- have a contract that holds a NFT (yes only one) that does not also need to have a collection. It needs to expose some information in a standard on way on what it needs to be put into a collection and how to link it.
- This also removes the need for NFTCollectionDisplay/NFTCollectionData on every NFT in a collection. This data can be moved to the collection itself
- a method on AuthAccount to setup a slot with a given classifier would be a bonus. Paths are standradized and hidden as discussed above.

An example NFT contract could then be as simple as 
```
//imports removed for brevity
pub contract ExampleNFT: NonFungibleToken {
  
    pub resource NFT: NonFungibleToken.INFT, MetadataViews.Resolver {
    
        pub let id:UInt64
        /// Metadata fields
        pub let name: String
        pub let description: String
        pub let thumbnail: String
        access(self) let royalties: [MetadataViews.Royalty]
        access(self) let metadata: {String: AnyStruct}
    
        init(
            id: UInt64,
            name: String,
            description: String,
            thumbnail: String,
            royalties: [MetadataViews.Royalty],
            metadata: {String: AnyStruct},
        ) {
            self.id = id
            self.name = name
            self.description = description
            self.thumbnail = thumbnail
            self.royalties = royalties
            self.metadata = metadata
        }


        pub fun getViews(): [Type] {
            return [
                Type<MetadataViews.Display>(),
                Type<MetadataViews.Royalties>(),
                Type<MetadataViews.Editions>(),
                Type<MetadataViews.Serial>(),
                Type<MetadataViews.Traits>()
            ]
        }

        /// Function that resolves a metadata view for this token.
        ///
        /// @param view: The Type of the desired view.
        /// @return A structure representing the requested view.
        ///
        pub fun resolveView(_ view: Type): AnyStruct? {
            switch view {
                case Type<MetadataViews.Display>():
                    return MetadataViews.Display(
                        name: self.name,
                        description: self.description,
                        thumbnail: MetadataViews.HTTPFile(
                            url: self.thumbnail
                        )
                    )
                case Type<MetadataViews.Editions>():
                    // There is no max number of NFTs that can be minted from this contract
                    // so the max edition field value is set to nil
                    let editionInfo = MetadataViews.Edition(name: "Example NFT Edition", number: self.id, max: nil)
                    let editionList: [MetadataViews.Edition] = [editionInfo]
                    return MetadataViews.Editions(
                        editionList
                    )
                case Type<MetadataViews.Serial>():
                    return MetadataViews.Serial(
                        self.id
                    )
                case Type<MetadataViews.Royalties>():
                    return MetadataViews.Royalties(
                        self.royalties
                    )
                case Type<MetadataViews.ExternalURL>():
                    return MetadataViews.ExternalURL("https://example-nft.onflow.org/".concat(self.id.toString()))
               
                case Type<MetadataViews.Traits>():
                    // exclude mintedTime and foo to show other uses of Traits
                    let excludedTraits = ["mintedTime", "foo"]
                    let traitsView = MetadataViews.dictToTraits(dict: self.metadata, excludedNames: excludedTraits)

                    // mintedTime is a unix timestamp, we should mark it with a displayType so platforms know how to show it.
                    let mintedTimeTrait = MetadataViews.Trait(name: "mintedTime", value: self.metadata["mintedTime"]!, displayType: "Date", rarity: nil)
                    traitsView.addTrait(mintedTimeTrait)

                    // foo is a trait with its own rarity
                    let fooTraitRarity = MetadataViews.Rarity(score: 10.0, max: 100.0, description: "Common")
                    let fooTrait = MetadataViews.Trait(name: "foo", value: self.metadata["foo"], displayType: nil, rarity: fooTraitRarity)
                    traitsView.addTrait(fooTrait)
                    
                    return traitsView

            }
            return nil
        }
    }


   //expose out data on what is needed to store this. Like if you need a certain interface to be present in the collection.
 

    //code that allows you to mint

}
```

austinkline:
Nice to see this make its way into a flip! A few initial questions/comments to get my bearings:

## Events
1. Are we settled on what the "rich" events might be for Deposit/Withdraws? Display makes total sense, I'm curious what Serial gives us if we aren't moving towards storing based on UUID (whole other tangled mess of complexity there). Ideally You're using display anyway to show what the event was, and the id of the nft tells you how to get to it.

2. Do we plan to emit rich details in both events (Deposit and Withdraw)? For the most part both deposit and withdraw go together, but not always. Flowty, for example, wouldn't emit a deposit when a loan is funded, it gets put into another resource until that loan is expired or it is repaid which would be outside the scope of a collection resource.

## Transfers
3. I'm not sure how I feel about `transfer` returning a bool, would it not just panic if the receiver can't take the item? Otherwise we might have cases where a thinks something worked (the transaction executed successfully) but the outcome wasn't what they expected. Has there been discussion about that?

## Helper methods 
1. Full agreement on `getIDs` and its challenges, I'll have to think of a way that we could maybe make that work for pagination and come back later with something
2. We'd had some discussion about a `safeBorrow` but it sounds like this is being handled by making `borrowNFT` return an optional, I think that's great
3. Would it make sense to add a method on the receiver for whether a type is able to be deposited? It might go nicely with `getAcceptedTypes` to have an `isAcceptedType(type: Type)`

m-Peter:
After reading the comments in the PR and the original proposal, I would like to share a few thoughts also. Mostly they elaborate on comments from @bjartek.

> There is however one part I do not like. And that is identification. I believe all NFTs should use uuid as the ID that is used to store it in the collection. If not there will be conflicts and problems.

I totally agree that this is a must-have prerequisite. Since it is something provided by the language, it should be used in all possible places, such as event payloads for example. It will provide a more solid tracing ability, than having to deal with custom NFT IDs from each possible implementation.

> Having multiple NFT types in the same storage path is not a good idea IMHO. Large dictionaries eat up gas.

Having paths store a single NFT type, will allow for a more fine-grained control to cover various use cases. For example, a "heavy" NFT type, will eat up gas for transactions dealing with "light" NFT types, because they all share the same storage path.

> Use the same Collection implementation to store multiple NFT types, but not in the same path. You do not have to rewrite the code to create a collection you only have to know what limitations this NFT type has on the collection it can be put into.

The Collection in the current NonFungibleToken contract is a first-class citizen, with more responsibility/authority than it should have. The changes should be towards making Collections just arrange NFTs and add capabilities to the NFT that it needs to function, like a primitive data structure but little more supercharged. I think limitations should be a more defining factor, for what is possible in joining a given collection. Basically the end-user should be able to mix-and-match NFTs, to desired collections, or  product lists as they would be called in traditional marketplaces.

> When I add a new NFT type that can be stored with a given Collection I do not have to touch that collection code, only the NFT code. This composes better.

This sounds like the Open-Closed Principle put into practice!

> Make it possible to create a NFT Collection code that can store multiple NFT types. The path that stores a single type has information about what that type is. We have a method that returns what type is stored here in the collection.

That's a pretty good use of leveraging the type system of the language.

> Have a contract that holds a NFT (yes only one) that does not also need to have a collection. It needs to expose some information in a standard on way on what it needs to be put into a collection and how to link it.

With the addition of static functions, this contract can act as a factory blueprint, on how to add a NFT type in a given Collection. Using interfaces to restrict what a NFT needs to be stored correctly in a Collection, is another SOLID thing to to.

I hope these remarks are not very confusing, and they do make some sense! :pray:

chriswayoub:
A few thoughts:

**NFT IDs**
I agree wholeheartedly that the UUID of the resource should be used as the identifier for the NFT everywhere. I do not think this should be something that is left up to the developer to define, and should be able to be trusted to refer to the specific NFT in question at all times. I've seen contracts that inadvertently minted two NFTs with the same ID, and it causes havoc with emitted events since there is no way to tell them apart. I'm sure there are also scenarios where this could be used maliciously by the owner of a contract.

**Collections**
I know this is a hot topic (shared vs separate collections), and personally I'm on the side of a single collection. But I will go a step further and say that collections implemented in a contract should not exist in the first place for a normal use-case. There should be a language-level way to store a resource in an account (without a storage path) and there should be a way to query that storage and retrieve a resource by UUID.

I think that this interface can be very simple:
* Store any resource
* Query storage for number of stored resources (perhaps filtering by Type)
* Retrieve resource from storage by UUID

If I have stored a resource in my account, then I believe that I should not have to rely on a contract implementation to access that resource. As it stands right now, I could prevent someone from withdrawing their own NFT that they are supposed to have ownership of. This can also happen if there is a problem with the contract itself (as we've seen with the secure cadence update), although that might become a non-issue with Stable Cadence.

I do think there are some things that need to be thought through with this approach, like how to give capabilities to withdraw/deposit, etc. I think if the core resource storage interface can be as lightweight as possible, and mainly just facilitate saving/loading resources to a built-in simple collection, then hopefully everything else can be implemented at the contract-level to allow for exposing custom functionality/capabilities to that storage.

joshuahannan:
A lot to respond to here, so I'll try to respond to groups of comments

@bjarte, I think your suggestion is good, but I think that a proper way of storing multiple NFT types in the same collection is out of the scope of this proposal. Is there anything in this proposal that makes your idea not possible?

Additionally, what you suggest with implementing an NFT without implementing a collection is already possible. We’re getting rid of static type requirements, so if you want to implement an NFT, all you have to do is implement the NFT resource interface on your NFT resource. The contract interface is just an optional thing you can implement if you want more of the convenience methods on your contract.

joshuahannan:
@austinkline
 
We are definitely not settled on what rich events can include. I’m not an app developer, so I don’t know what apps are expecting. If you have suggestions, please post them and we can discuss. I would like to include metadata in all events if possible though

Transfer:

That is just something I was playing around with, but I will probably remove it. Basically, since transactions can have any arbitrary code, I’d like for cadence to one day move to a paradigm where functions don’t usually panic, but return error codes and messages and the transaction code can handle the error and decide to panic if it wants or continue execution if the error isn’t a critical issue. This is already how many modern languages work, so it would be great for Cadence to be able to do that to. I don’t really think it works here though because we don’t have all the language support for it yet.

`isAcceptedType(type: Type)` is a good idea, I’ll add that

joshuahannan:
### Using uuid

I 100% agree that using uuid is the ideal way to mange this, but I can’t think of a way to enforce that contracts rely on uuid from now on. Many projects and smart contract are already using and will continue to use their contract specific IDs and migration to a completely uuid-focused paradigm seems extremely difficult operationally, but if you have ideas, please share them.

@m-peter, I’m a little confused about most of your comments about bjartek’s comments. Can you share a proposal about what the standard would need to do to address your concerns?

joshuahannan:
@chriswayoub Your proposal about managing collections and resources in storage sounds like something that is a core language feature, not a token standard feature. Have you tried making a FLIP for it? That would be the best approach to get discussion started on it

m-Peter:
Hi @joshuahannan, my apologies for the confusion :sweat_smile:. I will try to clarify things below :pray: 

My comments were basically affirmations on bjartek's comments, with some extra details on how they resonate with me.

Ideally, in the final/stable form, I see Collections as a "primitive data type" (maybe not included in the language standard library, but rather as a contract/contract interface possibly). They should be decoupled from a standard storage path, as they can very likely reside inside Dictionary/Array fields, in the contract-level (just like we have done [here](https://github.com/Build-Squad/nft-pawnshop/blob/main/cadence/contracts/NFTPawnshop.cdc#L6)). However, most of the times they will reside in an account's storage, but this should be up to the developer to decide where.

Just like dictionaries have some limitations on what can be used as the key (any object/data type that can generate a numeric `hash` value, for fast lookup, for example), Collections should leverage the type-system of the Cadence language, and take this a step further.
The "values" that a specific Collection will hold (in analogy to a Dictionary), doesn't have to be a concrete NFT type, but rather an interface, or list of interfaces. It could be all NFTs which have a `Display` Metadata View, or that have `Editions`, or `Royalties` or anything that the developer finds reasonable.

In closing, NFTs most likely evolve quite fast that's why Collections should be closed for modification, but open for extension, to seamlessly accommodate the need for integrating new NFT contracts/types.

That being said, this is just what I see as the end goal. The changes proposed for the V2 are clearly an improvement step, considering the current version of the `NonFungibleToken` contract interface.

I hope this does remove some/all of the confusion :pray:

bjartek:
We at .find can take a stab at creating a view in MetadataViews that is appropriate to put into events. Currently you cannot have functions and AnyStructs in there so some of the views will need to be transformed a little for this to work.

It will be based on FindMarket.NFTInfo and will support both &{NonFungibleToken.INFT} and &{MetadataViews.Resolver}

@joshuahannan I will take a closer read in a bit to see if any of the current suggestions are prohibiting my suggestions but I do not think so.

joshuahannan:
Here are the notes from the call last week with the open questions that we still have! I tried to organize them a little more so that are easier to follow. We might be talking about it again next tuesday at the SC eng open house.


- How to provide a standard for collections that store many NFTs
    - A single dictionary is not scalable when loading elements because you have to load the whole dictionary. When accounts have millions of NFTs, this can not scale
    - maybe a collection that stores sub collections in different storage paths when they get too large. Could this be a standard?
        - Would use AuthAccount capabilities or the proposed restricted auth account feature
    - Can we paginate `getIDs()`
- Interface Inheritance
    - [https://forum.onflow.org/t/flip-interface-inheritance-in-cadence/3750](https://forum.onflow.org/t/flip-interface-inheritance-in-cadence/3750)
    - This FLIP needs to be approved in order to enable enforcing that `Vaults` and `Collections` implement all the correct interfaces
    - Needs to have the default implementations conflicts resolved
    - Not required for what we’re trying to do, but if we don’t have it, then we won’t be able to enforce that types that implement `Vault` and `Collection` also implement the other standard iterfaces
    - If nested type requirements are not removed and we still have only one token per contract, then this is not required
- Nested Type Requirements Removal
    - We want to remove nested type requirements from the language, which would mean we need to change how we do events and interfaces.
    - This makes multiple tokens per contract possible. Is multiple NFTs per contract useful?
        - On the surface it seems like a nice flexible option to have, but it isn’t clear how many projects will use it.
        - It also allows us to encapsulate more of the functionality into the resource objects instead of at the contract level
            - this is nice since it is all in one place
            - also allows projects that have multiple NFTs to store them in the same collection instead of in separate collections
        - It makes some of the typical functions different and a little more awkward
            - Getting an empty vault or collection from the contract level now requires a type argument
            - Getting the standard paths now requires querying the metadata view from the contract, then looking at the paths there
            - getting total supply directly from the contract requires a type argument
        - We need more input from the community on if we want this feature or not
            - some might not see the value of having multiple token types and removing nested type requirements. might need some examples to illustrate
        - Universal collections might not be used so it might not be as important to remove type requirements
- Using UUID
    - Everyone agrees that if would have been perfect could have started using UUID for all NFTs from the beginning, but it wasn’t available then so we haven’t
    - Using uuid universally would be useful, but with existing collections who don’t use them this is hard to be compatible because most current collections, marketplaces, etc are keyed with the regular IDs, not uuid, so migrating would be extremely difficult
    - projects with old collections might have to define a new collection type using uuid that lives alongside the old collection to allow them to migrate over time
    - some apps like Flowty can’t change because they have extended loans that use the old system. migration is really hard and error prone and might permanently break existing state
    - Everyone agrees that sticking with the old ID system probably isn’t sustainable though. currently, NFTs can have the same ID, which is potentially a big problem, and it would be hard to tell which projects are using uuid and which aren’t
    - Events would also be changed with uuid changes
    - We could add more withdraw options to the standard collection such as `withdrawWithUuid()` or `withdrawWithType()` so that both systems can still be used
    - Could also have two different collection interfaces, one for the old and one for the new implementations, to allow projects to migrate and use the workarounds in the meantime
    - Need more input and suggestions on this issue
- Events emitted from Interfaces
    - [https://github.com/onflow/cadence/issues/2069](https://github.com/onflow/cadence/issues/2069)
    - with these events, uuid events would be emitted from day one
    - not required, but would be very nice to have
- Universal Collection
    - not very popular
    - if storage fees were different, it might not be an issue because it can be an attack vector
- Standard events?
    - Use Type or Type identifier as an event argument
    - convenient to be able to see views directly from events instead of having to query afterwards
    - might be too high of a cost to get these views for every withdrawal and deposit and it might make the events too large
    - some views might not be safe to display because some projects don’t allow display legally
    - may need better events that can give us more information for cases like in the staking contract where they are meaningless
- Upgrade issues
    - Circular dependency with metadata views
        - We would prefer to have the NFT interface directly implement the `MetadataViews.Resolver` interface, but can’t because that would create a circular dependency because `MetadataViews` imports `NonFungibleToken`
        - Create empty collection method is the only one that uses the circular dependency
        - could move the definition to a different contract to avoid the circular dependency
        - maybe move all the NFT specific views to a different contract
        - might break implementations because the types would be incompatible
        - Need to find a resolution for this
    - Upgrade issues that should hopefully be able to be resolved with a special exception in Cadence
        - Can’t give the interfaces the same name as the types in the original standards
        - Can’t remove the original type definitions
        - Can’t include structs in the event emissions
- Need to decide if we want to define new views for contract view resolver for the NFT standard, especially if we have multiple NFTs per contract

bjartek:
A good list. I would really love richer deposit/withdraw events. Just adding things like thumbnail and name would make it so much more useful. However I can see the issue as well since there are so many of these events so I am willing to compromise here unless somebody else has a really strong opinion about it. It was me that wanted it in the first place i think.

joshuahannan:
Here are the notes from the working group on Apr 4. The action items are bolded. I'll be adding these updates to the FT and NFT PRs and I'll share some of my findings in the next working group. Anyone who is interested in contributing and/or helping with trying out some of our ideas, please let me know!

- Nested Type Requirements Removal
    - **Decided to move forward with removing nested type requirements and encapsulating token functionality within resources**
- Using UUID
    - **Josh is going to play around with the below suggestions to see what might work for this and make a writeup for the pros and cons of each one**
    - We could add more withdraw options to the standard collection such as `withdrawWithUuid()` or `withdrawWithType()` so that both systems can still be used
    - Could also have two different collection interfaces, one for the old and one for the new implementations, to allow projects to migrate and use the workarounds in the meantime
    - maybe we can provide `fun getUDID(type:Type, id:UInt64): UInt64 ?` so all contracts can use uuid to migrate before we do the spork for stable cadence
        - borrow resource by uuid - talk to Dete
    - What if we emit an event in the getUUID() method that includes both the id and uuid so indexer can map both to each other? We could probably add that as a default implementation in NFT interface
        - already in the deposit and withdraw events
        - need to include get ID methods for old and and uuid
    - Could provide a method on NFTs for projects to specify if they are using UUID or not to avoid misunderstandings after the upgrade
- Events emitted from Interfaces
    - **Deniz is going to have a FLIP ready for this by the end of the week**
    - [https://github.com/onflow/cadence/issues/2069](https://github.com/onflow/cadence/issues/2069)
- Standard events?
    - **Will only include name and thumbnail as metadata event arguments for now**
    - Use Type or Type identifier as an event argument?
        - Type is serialized as the identifier in the event. need the identifier off-chain to be human readable. need to know what the type is actually emitted as
        - **Josh will do some research on this**
    - Even if events don’t have all the metadata, a third party could offer a service
        - access node could monitor events, then could query the metadata views with a script so that the full “event” can be script enhanced
- How to provide a standard for collections that store many NFTs
    - **Josh is going to research some more for this issue**
    - Very important that an NFT can tell where it is stored. You know the type of the NFT from an event, but you don’t know the path. An NFT should own where it can be stored. This is already in the collection view, but could be more general
    - NFTs can be stored anywhere, so how would it indicate a specific location
    - Could use the `Collection` interface to be the composable NFT standard interface
        - This could also be a bare minimum interface that doesn’t necessarily include all the methods in the collection interface
    - maybe a collection that stores sub collections in different storage paths when they get too large. Could this be a standard? Would use AuthAccount capabilities or the proposed restricted auth account feature
    - Same with attachments, it is hard to know where they are. What if an NFT is part of an attachment. How do you know where it is?
    - Rich events would help to indicate this, like for loading/saving resources, linking types, etc

joshuahannan:
I made some changes to the PR: https://github.com/onflow/flow-nft/pull/126
I described the changes in a [comment ](https://github.com/onflow/flow-nft/pull/126#issuecomment-1505864702). They  are basically the action items that I bolded in the comment above this one

bjartek:
What if Deposit/Withdraw events contains the uuid of the collection. If they do it is possible to distinguish these from each other if a user happends to have two of them linked.

joshuahannan:
We're having another working group meeting on Nov 7th to discuss the token standard improvements! We're hoping to finalize most of the changes so we can more confidently say what will be part of the release. You can see the event on the [Flow public event calendar.](https://calendar.google.com/calendar/embed?src=c_47978f5cd9da636cadc6b8473102b5092c1a865dd010558393ecb7f9fd0c9ad0%40group.calendar.google.com&ctz=America%2FChicago)

joshuahannan:
In the most recent smart contract design calls, we made some modifications to the standard proposal:

-   Removed transfer interface, function, and events
-   Removed uses UUID method
-   Removed uuid and type withdraw methods
-   Removed default implementations for getID, get supported types
        
We also have some things we still need to discuss:

- standard NFT update event: https://github.com/onflow/flow-nft/pull/196
- a way to get information about owned resources: https://github.com/onflow/flow-nft/pull/126#issuecomment-1836216586

If you have thoughts about wanting to keep some of the things that were removed or want to discuss the proposed changes or anything else about the standard, please join the next token standards discussion on Tuesday Dec 5th to share your feedback! The event is on the flow public events calendar

joshuahannan:
We had another discussion today about the token standards. We made a few decisions:

* We will include a default destruction event for `NFT`:
```
ResourceDestroyed(id: UInt64 = self.id, uuid: UInt64 = self.uuid)
```
* We will add the `id` field back to `NFT` to make it accessible in the default destruction event and for better enforcement.
* We will likely also remove `getID` because this is now superfluous.

* We decided to include methods in the default events to say the UUID of all the NFTs and collections involved. For example, withdraw will look like this now:
```
Withdraw(type: String, id: UInt64, uuid: UInt64, from: Address?, collectionUUID: UInt64)
```

* We had a discussion about how we should manage the `balance` field for `Vault`, but this doesn't affect the NFT standard.

We'll discuss this topics and more at the next meeting at 9am Pacific on Thurs Jan 11.

Proposed Agenda: 

* addition of `id` to NFT and removal of `getID()`
* Should we have methods to getWithdrawableBalance() or isAvailableToWithdraw(): Bool in the vault, provider, and balance interfaces?
* Should the token standards be contract interfaces to enforce ViewResolver and createEmptyVault()?
* How to name entitlements
* should ViewResolver.resolveView be renamed to resolveContractView?

joshuahannan:
We had another discussion about the token standards today. here is a summary of what we talked about:

* Should we include id as a field in {NFT}?
    * People can lie about the ID in the method, which is a risk
    * **DECISION: Keep the id field, remove the getID method**

* removing getter functions for standard paths and only including them in metadata views
    * They’re accessible as metadata views so it might make sense to have that be the only way to get them in a standardized way, simplifies the standard
    * Counterpoint: If it isn’t in the standard, people might forget to implement it
    * Probably should remove it since it likely won’t be used much
    * **DECISION: 
        * Remove getter functions for standard paths. 
        * Add function to get the NFTCollectionData metadata view. 
        * remove provider linked path and provider type from NFTCollectionData**

* Should the token standards be contract interfaces 
    * If they are interfaces, we can enforce conformance to ViewResolver and createEmptyVault
    * BUT, we would lose the ability to have standard events and utility functions like burn
    * **DECISION: Keep as a contract and find other ways to enforce that people are conforming to ViewResolver**

* should ViewResolver.resolveView be renamed to resolveContractView?
    * Makes it clear that it is for contract metadata instead of resource metadata
    * **DECISION: Rename them to getContractViews and resolveContractViews**
* Incompatible upgrades that are part of the new standards
    * Let people know that we are going to have a conversation with the cadence team soon to figure out how to migrate all these weird changes
    * Might be able to include new interfaces that are for legacy NFTs that implement the new ones to make migration easier
    * **DECISION: Have a conversation with the Cadence team next week during the next one of these meetings about what the migration difficulties are that need to be reverted or addressed**
    
    
Topics for next week:

* Should we keep the default implementations for ViewResolvers as nil?
    * These are really important, so we should make it as difficult as possible to not implement them properly
* Should we have balance as a field in Vault?
    * Should we have methods to getWithdrawableBalance() or isAvailableToWithdraw(): Bool in the vault, provider, and balance interfaces?
* How to name entitlements
    * Adjectives, Verbs?
    * Introduce three entitlement with Owner and have a mapping that maps to both withdraw and update?

joshuahannan:
We continued our token standards discussion today and focused on migration difficulties with the Cadence team.
Most of the discussion involved flagging the parts of the standard that probably needed a special migration.
Here are some notes with decisions bolded:

- Changing fungible token from an interface to a contract
    - Old interface has to be removed and new contract has to be added
    - Contract has no state, so that won’t be an issue
    - Types defined in the old interface will still work with the new contract
    - Need to check if a contract that implements a contract interface can emit an event defined in the interface
    - **Decision: Pending the investigation about events**
- Removing type definitions from standards and contracts
    - Example, removing event definitions, removing an interface definition
    - Should be possible with relaxed upgrade rules in the migration
- Changing vault, NFT, and Collection, from a resource type requirement to an interface definition
    - Lots of resources are stored as `FungibleToken.Vault` , but after this upgrade. that resource type will not exist any more and will be `{FungibleToken.Vault}` instead.
    - Will require a custom migration and there is no way around this because nested type requirements are being removed from the language.
    - Also involves removing the INFT resource interface in favor of the just NFT resource interface
        - **Decision: Still include the INFT interface and have it inherit from the NFT interface**
- Changing interface conformance for types in implementing contracts
    - For example, any token contract will have to update their `Collection` resource from `Collection: NonFungibleToken.Receiver, NonFungibleToken.Provider` to `Collection: NonFungibleToken.Collection`
    - Normal upgrade rules don’t allow this
    - Should be possible with relaxed upgrade rules in the migration
- Updating capabilities like to their correct types
    - Specific examples like `Provider` capabilities need to have the correct entitlement
    - More generally, capabilities that use restricted types will need to have their types updated to just be the concrete type
    - Cadence team is already aware of this. will require a custom migration
- Moving MetadataViews.Resolver to ViewResolver.Resolver
    - Updating any capabilities that are linked as MetadataViews to ViewResolver
    - Any stored types that have this type would also need to be updated
    - Also includes `ResolverCollection`
    - Should be fairly easy to fix in the migration

Next time, we will discuss entitlement naming and default implementations

joshuahannan:
We had another discussion today that, pending final approval of the code, finalizes the design of the standards!
Here is a summary of the decisions we made:

* Contract vs interface
    * We will make both standards interfaces now
    * Events that are emitted from implementations do not have the same type, so they are safe
    * Implementations will have to implement a getTypes method and a createEmpty method that takes a type as an argument
    * We will also propose new views for contracts that return information about token types in contracts
* How to name entitlements
    * nouns are used for "coarse-grained" entitlements (e.g. Storage to access all of the account storage functionality), and verbs are used for "fine-grained" entitlements, like specific functions (e.g. SaveValue to be able to store a value into account storage)
    * Rename entitlements to Withdraw, rename events to Withdrawn and Deposited
    * Add Owner entitlement to NFT standard to map to withdraw and update
* Default implementations
    * Remove default implementations for all view resolver methods
    * Remove them from every other method that makes sense
    
    
  We'll have another chat next Thursday to finish everything up

joshuahannan:
We had our last design meeting for the new token standards yesterday and settled on a design! Here are notes from the meeting that are relevant to the NFT standard:

* Decided to not try to have a conditional that emits events only if `to` or `from` is `nil`. It wasn't possible to include this cleanly in a contract interface and the benefit isn't enough to warrant putting them in another contract
* We updated the function that emits the `NonFungibleToken.Updated` event to be `access(contract)` so that it could not be spammed. Contracts who want to emit that event when an NFT is updated should call `{NameofNFTContract}.emitNFTUpdated()`

The code in [the PR](https://github.com/onflow/flow-nft/pull/126) is now the finalized code for the new standards and will be released in the upcoming CLI version and likely be the version that is deployed to all the networks unless a critical issue is surfaced in the meantime.

Thank you to everyone who contributed! The standards are looking quite good. 😎

joshuahannan:
Looking for some approvals on this FLIP so we can get it merged since it all has been resolved and implemented for a while now!

application/20221219-nft-v2.md
```diff
@@ -0,0 +1,369 @@
...
+
+`ViewResolver` has also been updated to require type information
+when querying views from a contract:
+
+```cadence
+access(all) contract interface ViewResolver {
+
+    /// Function that returns all the Metadata Views implemented by the resolving contract.
+    /// Some contracts may have multiple resource types that support metadata views
+    /// so there there is an optional parameter for specify which resource type the caller
```
bthaile, line 90: there there, double word

application/20221219-nft-v2.md
```diff
@@ -0,0 +1,369 @@
...
+    access(all) view fun isSupportedNFTType(type: Type): Bool
+}
+```
+
+### Change the return type of `borrowNFT()` to an optional
+
+In the current standard, `borrowNFT()` is meant to panic if the NFT to borrow
+is not found in the collection. This is clearly the wrong choice, because 
+it is a best practice to have getter functions return `nil` if something goes wrong
+instead of panicing and reverting the transaction.
```
bthaile, line 171: panicing, should be "panicking" just spelling.

application/20221219-nft-v2.md
```diff
@@ -0,0 +1,369 @@
...
+
+In the case of events, currently there is no way for the standard to ensure that
+implementations are emitting standardized and correct events, which this upgrade will address.
+
+## User Benefit
+
+With these upgrades, users will now be able to:
+* Define multiple tokens in a single contract.
+* Query all the data about a token directly through the contract as well as the `Vault` resource and core interfaces.
+* Have standard events emitted correctly for important operations (`Withdrawn`, `Deposited`, `Updated`, and `NonFungibleToken.NFT.ResourceDestroyed`)
```
bthaile, line 59: nit, add a ending period to be consisten.

bthaile (approved):
a few nits

bjartek (approved):

application/20221219-nft-v2.md
```diff
@@ -0,0 +1,369 @@
...
+    /// @param resourceType: An optional resource type to return views for
+    /// @return An array of Types defining the implemented views. This value will be used by
+    ///         developers to know which parameter to pass to the resolveView() method.
+    ///
+    access(all) view fun getContractViews(resourceType: Type?): [Type]
+
+    /// Function that resolves a metadata view for this token.
+    /// Some contracts may have multiple resource types that support metadata views
+    /// so there there is an optional parameter for specify which resource type the caller
+    /// is looking for views for.
```
sisyphusSmiling, line 104:
```suggestion
    /// so there there is an optional parameter to specify which resource type the caller
    /// is requesting views for.
```

application/20221219-nft-v2.md
```diff
@@ -0,0 +1,369 @@
...
+
+### Change the return type of `borrowNFT()` to an optional
+
+In the current standard, `borrowNFT()` is meant to panic if the NFT to borrow
+is not found in the collection. This is clearly the wrong choice, because 
+it is a best practice to have getter functions return `nil` if something goes wrong
+instead of panicing and reverting the transaction.
+
+```cadence
+pub fun borrowNFT(id: UInt64): &AnyResource{NFT}?
```
sisyphusSmiling, line 174:
```suggestion
access(all) view fun borrowNFT(_ id: UInt64): &{NFT}?
```

application/20221219-nft-v2.md
```diff
@@ -0,0 +1,369 @@
...
+for the collection to indicate which IDs it contains.
+
+### Move `createEmptyCollection()` to inside the Collection definition in addition to the contract
+
+It is useful to be able to create a new empty collection
+directly from a `Collection` object instead of having to import
+the contract and call the method from the contract.
+
+```cadence
+pub resource interface Collection {
```
sisyphusSmiling, line 192:
```suggestion
access(all) resource interface Collection {
```

application/20221219-nft-v2.md
```diff
@@ -0,0 +1,369 @@
...
+
+### Remove the requirement for the `ownedNFTs` field in `Collection`
+
+The requirement to include an `ownedNFTs: {UInt64: @NFT}` field
+in Collection implementations is restrictive because developers may want
+to store tokens in a unique way that is different what what could be specified.
+This proposal removes `ownedNFTs` and only requires a `getIDs()` method
+for the collection to indicate which IDs it contains.
+
+### Move `createEmptyCollection()` to inside the Collection definition in addition to the contract
```
sisyphusSmiling, line 185: Might be worth mentioning this is also added to the `NFT` interface as well for similar reasoning.

application/20221219-nft-v2.md
```diff
@@ -0,0 +1,369 @@
...
+    }
+}
+```
+
+### Add Metadata Views methods to the standard
+
+Metadata Views for non-fungible tokens should be easily accessible
+from interfaces defined by the standard. 
+This propsal enforces that implementations implement the `ViewResolver` contract interface
+and `NFT` implementations implement the `ViewResolver.Resolver` interface.
```
sisyphusSmiling, line 210:
```suggestion
This proposal enforces that implementations implement the `ViewResolver` contract interface, `NFT` implementations implement the `ViewResolver.Resolver` interface, and `Collection` implementations implement the `ViewResolver.ResolverCollection` interface.
```

application/20221219-nft-v2.md
```diff
@@ -0,0 +1,369 @@
...
+
+A critical piece of tooling for non-fungible tokens would be a struct that contains
+a provider capability but restricts the capability to only be able to withdraw
+specified IDs or number of tokens from the underlying collection.
+Currently, providers have no limit, but all tokens should be able
+to create scoped providers.
+
+This feature is out of the scope of this proposal, but should definitely be a standard
+that lives alongside the main fungible token standard.
+We hope to shepard a proposal for these soon.
```
sisyphusSmiling, line 351:
```suggestion
We hope to shepherd a proposal for these soon.
```

sisyphusSmiling (approved):
Looks great!

austinkline (approved):

application/20221219-nft-v2.md
```diff
@@ -0,0 +1,369 @@
+---
+status: Implemented 
+flip: 56
+authors: Joshua Hannan (joshua.hannan@dapperlabs.com)
+sponsor: Joshua Hannan (joshua.hannan@dapperlabs.com)
```
franklywatson, line 5:
```suggestion
authors: Joshua Hannan (joshua.hannan@flowfoundation.org)
sponsor: Joshua Hannan (joshua.hannan@flowfoundation.org) 
```

franklywatson (approved):

joshuahannan:
Thanks for the reviews and approvals everybody! I've addressed all your comments and I think we have enough approvals now that I'm okay to merge. 🥳 ❤️

PR #57, opened 2023-01-10
Access Node Minimum Stake Requirement

joshuahannan:
Adds a FLIP proposing a staking minimum for access nodes

### DIFF
--- a/protocol/20230110-accessnode-stake.md
+++ b/protocol/20230110-accessnode-stake.md
@@ -1,0 +1,121 @@
+---
+status: draft 
+flip: NNN (do not set)
+authors: Joshua Hannan (joshua.hannan@dapperlabs.com)
+sponsor: Joshua Hannan (joshua.hannan@dapperlabs.com) 
+updated: 2023-01-10 
+---
+
+# Access Node Staking Minimum
+
+## Objective
+
+Introduce a minimum required stake for access nodes
+in the Flow protocol and staking smart contract.
+
+## Motivation
+
+Access nodes were proposed to be an easy way to run a Flow node
+without having to commit much capital for extended periods of time.
+The access node is a staked node that receives and indexes
+all protocol state data and serves [the Access API.](https://developers.flow.com/nodes/access-api)
+
+While access nodes aren't directly critical to the core functioning
+of the network, they serve an important role of being the point
+of access for transaction submission, state querying, and more.
+They are also capable of malicious behavior that can negatively affect
+users and the network. Until now, access nodes have been staked
+in the identity table, but haven't been required to actually stake any FLOW
+to participate.
+
+This FLIP proposes increasing the access node minimum to a value
+that reflects its importance to the network and prevents certain vulnerabilities
+in the smart contract that are a result of having a zero minimum stake requirement.
+
+The Flow community is getting ready to allow anyone to run access nodes
+without permission from the consensus committee. As part of these changes,
+the staking smart contract will be updated to remove the approval restriction
+for access nodes and introduce the concept of a candidate node list,
+which refers to nodes who have registered and committed sufficient stake
+to be a node operator for the next epoch,
+but aren't currently in the identity table.
+
+The protocol uses this list to randomly select nodes to be node operators,
+and there is a limit to the number of nodes who can be candidate nodes for each role.
+Because access nodes' minimum stake requirement is currently zero,
+an attacker can register any number of nodes until the limit is reached,
+effectively blocking other users from registering access nodes
+until the next epoch when the candidate list is reset,
+at which point they could just do it again.
+
+To attempt to address this vulnerability while also keeping access nodes relatively cheap,
+this FLIP proposes that the access node minimum stake requirement is set to .
+This requirement will be treated the same as the minimum stake requirements
+for other node types, meaning that access nodes will be removed if their committed
+stake is less than the minimum.
+
+**This proposal does not enable access nodes to receive rewards**.
+While access nodes will have to stake FLOW, they will still not receive any rewards.
+
+## User Benefit
+
+This change would not have any direct benefit to users,
+but would prevent attackers from blocking access to node registrations
+by exploiting the vulnerability.
+
+## Design Proposal
+
+Remove any restrictions in the contract that do not count access nodes
+as staked or prevent them from staking, such as in
+`getTotalStaked`](https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowIDTableStaking.cdc#L1406)
+
+The service account committee runs a transaction, 
+[`set_minimums.cdc`](https://github.com/onflow/flow-core-contracts/blob/master/transactions/idTableStaking/admin/change_minimums.cdc)
+to set the access node minimum to the desired value.
+
+### Drawbacks
+
+This makes running an access node slightly more expensive.
+
+### Alternatives Considered
+
+* Keeping the stake requirement at zero, but requiring a deposit at registration time
+that is returned to the node operator at the end of the epoch.
+This was discarded because it is only a temporary solution that adds
+more work but doesn't have much benefit in the short term.
+
+### Performance Implications
+
+There should be no performance changes with this update.
+
+### Dependencies
+
+This change will not affect any standard transactions or scripts.
+Users who are registering access nodes will need to be made aware that 
+they must commit tokens for their node now.
+
+### Engineering Impact
+
+This is a small change that can be easily tested and maintained. 
+
+### User Impact
+
+* This will be rolled out with an upgrade to the staking contract
+and a transaction to set the staking minimum for access nodes.
+
+## Related Issues
+
+This proposal is related to the automated slot selection proposal
+and its related PRs:
+* https://github.com/onflow/flow-core-contracts/pull/321
+* https://github.com/onflow/flow-core-contracts/pull/324
+* https://github.com/onflow/flow-core-contracts/pull/328
+
+That proposal contains much bigger changes to how nodes are selected
+for each epoch as well as some code refactors to improve performance
+of the epoch smart contracts.
+
+## Questions and Discussion Topics
+
+Need to define the values for slot limits, candidate node limits,
+and the access node staking minimum amount.
--- a/protocol/20230110-accessnode-stake.md
+++ b/protocol/20230110-accessnode-stake.md
@@ -33,7 +33,7 @@ that reflects its importance to the network and prevents certain vulnerabilities
 in the smart contract that are a result of having a zero minimum stake requirement.
 
 The Flow community is getting ready [to allow anyone to run access nodes
-without permission from the consensus committee](https://github.com/onflow/flips/blob/main/protocol/20220719-automated-slot-assignment.md). As part of these changes,
+without permission from the service account committee](https://github.com/onflow/flips/blob/main/protocol/20220719-automated-slot-assignment.md). As part of these changes,
 the staking smart contract will be updated to remove the approval restriction
 for access nodes and introduce the concept of a candidate node list,
 which refers to nodes who have registered and committed sufficient stake
@@ -70,8 +70,7 @@ This change would benefit users operating Access Node by preventing attackers fr
 
 ## Design Proposal
 
-Remove any restrictions in the contract that do not count access nodes
-as staked or prevent them from staking, such as in
+Remove any restrictions in the contract that do not account for access nodes committing tokens as stake, such as in
 `getTotalStaked`](https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowIDTableStaking.cdc#L1406)
 
 The service account committee runs a transaction, 
--- a/protocol/20230110-accessnode-stake.md
+++ b/protocol/20230110-accessnode-stake.md
@@ -1,5 +1,5 @@
 ---
-status: Proposed 
+status: Accepted 
 flip: FLIP-57
 authors: Joshua Hannan (joshua.hannan@dapperlabs.com)
 sponsor: Joshua Hannan (joshua.hannan@dapperlabs.com) 
--- a/protocol/20230110-accessnode-stake.md
+++ b/protocol/20230110-accessnode-stake.md
@@ -40,7 +40,7 @@ which refers to nodes who have registered and committed sufficient stake
 to be a node operator for the next epoch,
 but aren't currently in the identity table.
 
-The protocol uses this list to randomly select nodes to participate in the network,
+The protocol uses this list to randomly select nodes from the candidate list to participate in the network,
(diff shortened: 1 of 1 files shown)


protocol/20230110-accessnode-stake.md
```diff
@@ -0,0 +1,121 @@
...
+stake is less than the minimum.
+
+**This proposal does not enable access nodes to receive rewards**.
+While access nodes will have to stake FLOW, they will still not receive any rewards.
+
+## User Benefit
+
+This change would not have any direct benefit to users,
+but would prevent attackers from blocking access to node registrations
+by exploiting the vulnerability.
```
jordanschalm, line 64:
```suggestion
This change would benefit users operating Access Node by preventing attackers from blocking access to node registrations by exploiting the vulnerability.
```

Sounds like a benefit to me! 😀

protocol/20230110-accessnode-stake.md
```diff
@@ -0,0 +1,121 @@
...
+### Drawbacks
+
+This makes running an access node slightly more expensive.
+
+### Alternatives Considered
+
+* Keeping the stake requirement at zero, but requiring a deposit at registration time
+that is returned to the node operator at the end of the epoch.
+This was discarded because it is only a temporary solution that adds
+more work but doesn't have much benefit in the short term.
```
jordanschalm, line 85:
```suggestion
more engineering effort but doesn't have much benefit in the long term, because it would eventually be replaced by stake requirements similar to this proposal.
```

jordanschalm (commented):
Looks good to me once we have a proposal for stake requirement.

protocol/20230110-accessnode-stake.md
joshuahannan: good point. 😆

protocol/20230110-accessnode-stake.md
```diff
@@ -0,0 +1,121 @@
...
+## User Benefit
+
+This change would not have any direct benefit to users,
+but would prevent attackers from blocking access to node registrations
+by exploiting the vulnerability.
+
+## Design Proposal
+
+Remove any restrictions in the contract that do not count access nodes
+as staked or prevent them from staking, such as in
```
peterargue, line 74: nit: access nodes currently don't commit tokens as stake, but are still "staked"
```suggestion
Remove any restrictions in the contract that do not account for access nodes
committing tokens as stake, such as in
```

protocol/20230110-accessnode-stake.md
```diff
@@ -0,0 +1,121 @@
...
+users and the network. Until now, access nodes have been staked
+in the identity table, but haven't been required to actually stake any FLOW
+to participate.
+
+This FLIP proposes increasing the access node minimum to a value
+that reflects its importance to the network and prevents certain vulnerabilities
+in the smart contract that are a result of having a zero minimum stake requirement.
+
+The Flow community is getting ready to allow anyone to run access nodes
+without permission from the consensus committee. As part of these changes,
```
peterargue: should this be?
```suggestion
without permission from the service account committee. As part of these changes,
```

protocol/20230110-accessnode-stake.md
```diff
@@ -0,0 +1,121 @@
...
+The protocol uses this list to randomly select nodes to be node operators,
+and there is a limit to the number of nodes who can be candidate nodes for each role.
+Because access nodes' minimum stake requirement is currently zero,
+an attacker can register any number of nodes until the limit is reached,
+effectively blocking other users from registering access nodes
+until the next epoch when the candidate list is reset,
+at which point they could just do it again.
+
+To attempt to address this vulnerability while also keeping access nodes relatively cheap,
+this FLIP proposes that the access node minimum stake requirement is set to .
```
peterargue:
```suggestion
this FLIP proposes that the access node minimum stake requirement is set to [TBD].
```

protocol/20230110-accessnode-stake.md
```diff
@@ -0,0 +1,140 @@
...
+users and the network. Until now, access nodes have been staked
+in the identity table, but haven't been required
+to actually stake any FLOW to participate.
+
+This FLIP proposes increasing the access node staking minimum to a value
+that reflects its importance to the network and prevents certain vulnerabilities
+in the smart contract that are a result of having a zero minimum stake requirement.
+
+The Flow community is getting ready [to allow anyone to run access nodes
+without permission from the consensus committee](https://github.com/onflow/flips/blob/main/protocol/20220719-automated-slot-assignment.md). As part of these changes,
```
peterargue, line 36: should this be?
```suggestion
without permission from the service account committee](https://github.com/onflow/flips/blob/main/protocol/20220719-automated-slot-assignment.md). As part of these changes,
```

protocol/20230110-accessnode-stake.md
```diff
@@ -0,0 +1,139 @@
...
+This requirement will be treated the same as the minimum stake requirements
+for other node types, meaning that access nodes will be removed if their committed
+stake is less than the minimum.
+
+The exact number of FLOW required for minimum stake is not that important.
+It basically needs to be small enough that it is affordable
+to run an access node for most people,
+but not so small that it is easy to fill the candidate node list.
+The candidate node limit for access nodes will be over 1,000 nodes,
+so in order to fill up the list, it would take over 100,000 FLOW every week to perform this attack, which we believe is enough to dissuade potential attackers from exploiting it.
```
sisyphusSmiling, line 62: How can we be so convinced that the mentioned 100K FLOW is sufficient to mitigate the spam vector? If we're talking about an attack from a competing protocol, state-level actor, or otherwise highly capitalized & motivated actor, 100K seems pretty menial. How critical is the vulnerability posed by the mentioned attack vector?

A follow up question that may be out of the scope of this flip: are there fallback mechanisms the community can leverage to recover from this attack should it occur? Such a mechanism could minimize the risk posed by setting an accessible minimum which I realize is important for community access & participation.

bluesign:
We have limit to nodes afaik, this will be (AN) included in that limit? And what is the limit actually? 

Also now we have limit of nodes in the network, and we have limit to candidates to be includes in the limited number of spots in the network? 

I know 100 has no mathematics involved but can we involve some economics work here? i think we have someone for this (though don’t know github username)

I think it is not depending on the queue size but also how fast queue is moving. If 100 people waiting as candidate this week, and lets say only 2 will be AN, then 98 will be candidate next week, that means only 2 spots for new candidates will be available. 

if we have slow moving queue I feel we should have bigger amount staked, like 5000-10000. And I feel we should target this amount to expected time of wait in the queue. (Which should be 2-3 weeks max in my opinion )

joshuahannan:
@bluesign There will be a limit for each node role, so access nodes will only count towards the limit for access nodes. The limit to start out will be around 120, but will get increased periodically.

Also, I think you're misunderstanding how the candidate list works. It is completely cleared every epoch, so if you want to be eligible again, you need to commit tokens again.

There actually was some math done to decide on this number, but afaik, the actual number doesn't matter that much as long as it is something that isn't insignificant. I pinged Dete and Kshitij to get them to respond with some more thorough context, but basically in order to exploit this with 1000 access node slots, you'd need 100k FLOW that you'd have to have continuously locked up for as long as you want to be exploiting it, which is a pretty big amount of FLOW to use. Additionally, there wouldn't be any other negative affect on the network if this was exploited and there is no other benefit to the attacker besides preventing others from registering access nodes for a week.

I think we're open to increasing it to up to 1000 FLOW, so if you think that it should be more, that might be possible if everyone agrees, but we don't want it to be too high because we want access nodes to be relatively affordable

bluesign:
@joshuahannan  I think 100 FLOW is great but I am scared that it will fill up fast ( 120 capacity ). Imagine 120 capacity is full, and we increase 5 slot per week. 

Few points: 

- We will have a lot of repeat customers for candidate list. Even candidate count is let's say 100, each will have 10x entry ( as logical )  
- 100k FLOW staking income for a week is ~ 180 FLOW, which is not a big cost to not fill candidate list. ( 0.18 FLOW for 100 FLOW ) 
- I believe if we cannot guarantee access node, it should be expensive. As it is limited resource, setting this high is much better. If we make lottery it should be more expensive to enter ( not 0.18 FLOW ) 
- I also think we can make it higher to be candidate, but running can be cheaper.

joshuahannan:
We mention in the FLIP that access nodes still will not receive any rewards, so the only benefit to users staking access nodes is that they get to run their own, and it will actually cost them money to run one on top of the staking minimum, so I don't think the demand is going to be as high as you think it will be. Does that affect your calculation at all?

joshuahannan:
Also, I think it is important to mention that this is only a temporary solution. We're going to start working on actual staking auctions this year, so nodes will have to actually bid for slots instead of just being randomly chosen with a flat minimum in the current design. So after that point, the market will set what the "minimum" should be.

bluesign:
> Also, I think it is important to mention that this is only a temporary solution. We're going to start working on actual staking auctions this year, so nodes will have to actually bid for slots instead of just being randomly chosen with a flat minimum in the current design. So after that point, the market will set what the "minimum" should be.

ah ok, then all is good. I thought this will stay long time.

protocol/20230110-accessnode-stake.md
KshitijChaudharyHBS: Hey @sisyphusSmiling, like @vishalchangrani explained, the vulnerability posed is not huge and only exists at the networking layer. However, as the number of slots in the registration list (FlowIDTable) is limited, to ensure that attackers don’t acquire all slots and bar anyone else from being considered to run a node, we decided to introduce a positive yet nominal minimum stake (100 Flow) per slot for an operator to legitimately acquire a slot in the FlowIDTable. After all interested operators have registered on the FlowIDTable and have submitted at least the minimum required stakes to the addNodeInfo function, an algorithm randomly selects some operators from the filled slots, allowing them to run an access node in the subsequent epoch.

sisyphusSmiling: Thank you both @vishalchangrani and @KshitijChaudharyHBS for the explanation. The mechanisms you noted as well as the [stated plans](https://github.com/onflow/flips/pull/57#issuecomment-1381995597) for upcoming staking auctions addressed any concerns I had.

bluesign: > While it cannot impact safety, an errant Access node can impact liveness 

@vishalchangrani what does this mean exactly ?

franklywatson (approved):

protocol/20230110-accessnode-stake.md
vishalchangrani: hi @bluesign - I mean that errant access nodes can't cause a fork or affect the validity of the chain however they could cause the chain to slow down or even stop block production.


(Specifically, “safety” can be thought of as nothing incorrect ever happening; and “liveness” can be thought of as the correct thing always eventually happening [source](https://learn.radixdlt.com/article/what-do-security-safety-and-liveness-mean#:~:text=For%20blockchains%20and%20DLTs%2C%20these,correct%20thing%20always%20eventually%20happening.)

vishalchangrani (approved):

protocol/20230110-accessnode-stake.md
```diff
@@ -0,0 +1,139 @@
...
+without having to commit much capital for extended periods of time.
+The access node is a staked node that receives and indexes
+all protocol state data and serves [the Access API.](https://developers.flow.com/nodes/access-api)
+
+While access nodes aren't directly critical to the core functioning
+of the network, they serve an important role of being the point
+of access for transaction submission, state querying, and more.
+They are also capable of malicious behavior that can negatively affect
+users and the network. Until now, access nodes have been staked
+in the identity table, but haven't been required
```
AlexHentschel, line 28: > staked in the identity table

it might be worth-while just briefly explaining what the means

```suggestion
in the identity table (i.e. they register their node ID, address and keys with the network), but haven't been required
```

protocol/20230110-accessnode-stake.md
```diff
@@ -0,0 +1,139 @@
...
+
+The Flow community is getting ready [to allow anyone to run access nodes
+without permission from the service account committee](https://github.com/onflow/flips/blob/main/protocol/20220719-automated-slot-assignment.md). As part of these changes,
+the staking smart contract will be updated to remove the approval restriction
+for access nodes and introduce the concept of a candidate node list,
+which refers to nodes who have registered and committed sufficient stake
+to be a node operator for the next epoch,
+but aren't currently in the identity table.
+
+The protocol uses this list to randomly select nodes to be node operators,
```
AlexHentschel, line 43: > select nodes to be node operators

to me this sounds odd. Are we conflating nodes (machines) and node operators (humans)?

```suggestion
The protocol uses this list to randomly select nodes to participate in the network,
```

protocol/20230110-accessnode-stake.md
```diff
@@ -0,0 +1,139 @@
...
+While access nodes will have to stake FLOW, they will still not receive any rewards.
+
+## User Benefit
+
+This change would benefit users operating Access Node by preventing attackers from blocking access to node registrations by exploiting the vulnerability.
+
+## Design Proposal
+
+Remove any restrictions in the contract that do not account for access nodes committing tokens as stake, such as in
+`getTotalStaked`](https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowIDTableStaking.cdc#L1406)
```
AlexHentschel, line 74:
```suggestion
[`getTotalStaked`](https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowIDTableStaking.cdc#L1406)
```

AlexHentschel (approved):
Very nice ✨. 

Just a few minor wording suggestions:

protocol/20230110-accessnode-stake.md
```diff
@@ -0,0 +1,139 @@
...
+this FLIP proposes that the access node minimum stake requirement is set to 100 FLOW.
+This requirement will be treated the same as the minimum stake requirements
+for other node types, meaning that access nodes will be removed if their committed
+stake is less than the minimum.
+
+The exact number of FLOW required for minimum stake is not that important.
+It basically needs to be small enough that it is affordable
+to run an access node for most people,
+but not so small that it is easy to fill the candidate node list.
+The candidate node limit for access nodes will be over 1,000 nodes,
```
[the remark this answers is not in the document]
AlexHentschel: I think we are planning to change this and have a candidate list for each node role separately.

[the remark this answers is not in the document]
joshuahannan: yes, we have a candidate list for each node role

protocol/20230110-accessnode-stake.md
```diff
@@ -0,0 +1,139 @@
+---
+status: Proposed
```
vishalchangrani, line 2:
```suggestion
status: Accepted 
```

PR #58, opened 2023-01-17
Improve Cadence external mutability restrictions

SupunS:
Work toward https://github.com/onflow/cadence/issues/2273


SupunS:
Yes, I agree. While this technically solves the existing problems, the added complexity (hence reduced usability/ composability) does not seem to be acceptable or justifiable.

I prefer #59 (with modifications if needed) over this.

bluesign:
> Make let modifier behave similarly to constants.

does this mean you can set only on init once?

SupunS:
>> Make let modifier behave similarly to constants.
>
> does this mean you can set only on init once?

Yes. Only impacts container types, because simple/primitive types already behave in that way. With this `let` containers cannot be modified after creation. Would have to add runtime checks to prevent modification of container-typed constants.

I don't feel it's an elegant solution though, because of the need for runtime checks. Hence why added as a potential alternative solution.

bluesign:
> I don't feel it's an elegant solution though, because of the need for runtime checks. Hence why added as a potential alternative solution.

why do we need runtime checks ? Can't we put a restriction on contract upgrades and say you cannot change field from let to var ?

turbolent:
We had breakout sessions, so far the sentiment is negative, and there has been no further feedback. 

FLIPs #89 and #86 have been opened with proposals for alternative solutions that might better solve the same problems that this FLIP tried to address.

PR #69, opened 2023-02-06
Discoverability of FT Vault types

satyamakgec:
This adds a FLIP for introducing an additional method in `FungibleToken.Receiver` interface to retrieve the expected vault type.

### DIFF
--- a/flips/20230206-fungible-token-vault-type-discovery.md
+++ b/flips/20230206-fungible-token-vault-type-discovery.md
@@ -56,12 +56,22 @@ So new `Receiver` interface would look like this -
         pub fun deposit(from: @Vault)
 
         /// Returns the type of implementing resource i.e If `FlowToken.Vault` implements
-        /// this then it would return `[Type<@FlowToken.Vault>()]`
+        /// this then it would return `[Type<@FlowToken.Vault>()]` and if any custom receiver
+        /// uses the default implementation then it would return empty array as its parent
+        /// resource doesn't conform with the `FungibleToken.Vault` resource.
         ///
-        /// @return Optional list of vault types.
+        /// @return list of supported vault types by the implemented resource.
         /// 
         pub fun getSupportedVaultTypes() :[Type] {
-            return [self.getType()]
+            // Below check is implemented to make sure that run-time type would
+            // only get returned when the parent resource conforms with `FungibleToken.Vault`. 
+            if self.getType().isSubtype(of: Type<@FungibleToken.Vault>()) {
+                return [self.getType()]
+            } else {
+                // Return empty array as the default value for the resource who don't
+                // implements `FungibleToken.Vault`, such as `FungibleTokenSwitchboard`, `TokenForwarder` etc.
+                return []
+            }
         }
     }
 ```
--- a/flips/20230206-fungible-token-vault-type-discovery.md
+++ b/flips/20230206-fungible-token-vault-type-discovery.md
@@ -1,6 +1,6 @@
 ---
 status: Draft
-flip: <TBD>
+flip: #69
 author: Satyam Agrawal (satyam.agrawal@dapperlabs.com)
 updated: 2023-02-06
 --- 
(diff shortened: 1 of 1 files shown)


alilloig:
Looks good so far! Probably it would be a good thing to include also how `TokenForwarder` contract will implement this, maybe not in the FLIP but in the future PR for updating the FT repo

flips/20230206-fungible-token-vault-type-discovery.md
```diff
@@ -0,0 +1,132 @@
+---
+status: Draft
+flip: <TBD>
```
joshuahannan, line 3:
```suggestion
flip: #69 
```

flips/20230206-fungible-token-vault-type-discovery.md
```diff
@@ -0,0 +1,132 @@
...
+author: Satyam Agrawal (satyam.agrawal@dapperlabs.com)
+updated: 2023-02-06
+--- 
+# Fungible Tokens Vault Type Discovery
+
+## Objective
+
+The purpose of this proposal is to enable Fungible Tokens (FTs) to be returned as vault types using the `FungibleToken.Receiver` interface. This would enhance the discoverability of FT vault types and reduce the likelihood of deposit vault failures. This is because the smart contract would now have prior knowledge of the acceptable vault types for deposits.
```
joshuahannan, line 11:
```suggestion
The purpose of this proposal is to enable the Fungible Token (FT) `Receiver`s to returne a list of vault types that they can accept.
This would enhance the discoverability of FT vault types and reduce the likelihood of unintended failures in the deposit method because the deposited vault is not supported.
This is because Cadence code trying to perform a deposit would now be able to query the acceptable vault types before making a deposit.
```

flips/20230206-fungible-token-vault-type-discovery.md
```diff
@@ -0,0 +1,132 @@
...
...only receive Flow tokens and a FUSD Vault can only receive FUSD tokens. Previously, there was no programmatic way to determine the type of token a `Vault` would receive, leading to the possibility of a failed deposit due to an incorrect assumption about the Vault's capabilities. To address this issue, a new proxy receiver, the [`FungibleTokenSwitchboard`](https://github.com/onflow/flow-ft/blob/master/contracts/FungibleTokenSwitchboard.cdc), was created to allow for the receipt of multiple Vault types through a single capability, the `Capability<&{FungibleToken.Receiver}>`. The [`SwitchboardPublic`](https://github.com/onflow/flow-ft/blob/4416bbe585629671d00d3acfa6fd8052104dd861/contracts/FungibleTokenSwitchboard.cdc#L37) interface also includes the `getVaultTypes` function, which partially solves the problem of `Vault` discoverability. However, this solution is not complete as the provided capability, `Capability<&{FungibleToken.Receiver}>`, can represent any type of receiver, including the `FungibleTokenSwitchboard`, a standard receiver, or another proxy receiver. As the Flow ecosystem expands, the number of proxy receivers developed to solve various problems is likely to increase.
```
joshuahannan, line 15:
```suggestion
The [Fungible Token Standard](https://github.com/onflow/flow-ft) mandates that a single Vault can only receive one type of token, 
meaning that a Flow Vault can only receive Flow tokens and a FUSD Vault can only receive FUSD tokens.
The `FungibleToken.Receiver` interface, on the other hand, is designed to be able to accept any Vault type in its `deposit` method.
Currently, there is no programmatic way to determine the type of token that a `Receiver` is able receive, leading to the possibility of a failed deposit due to an incorrect assumption about the Vault that the `Receiver` is linked to.
To address this issue, a new proxy receiver, the [`FungibleTokenSwitchboard`](https://github.com/onflow/flow-ft/blob/master/contracts/FungibleTokenSwitchboard.cdc), was created to allow for the receipt of multiple Vault types through a single `FungibleToken.Receiver` capability.
The [`SwitchboardPublic`](https://github.com/onflow/flow-ft/blob/4416bbe585629671d00d3acfa6fd8052104dd861/contracts/FungibleTokenSwitchboard.cdc#L37) interface also includes the `getVaultTypes` function, which solves the problem of knowing what vaults the receiver accepts, but only for the switchboard and only for code that is aware of the switchboard. 
This solution is not complete. A `{FungibleToken.Receiver}` capability can represent any type of receiver such the `FungibleTokenSwitchboard`, a standard receiver, or any other custom `Receiver`.
As the Flow ecosystem expands, the number and complexity of proxy receivers developed to solve various problems is likely to increase.
```

flips/20230206-fungible-token-vault-type-discovery.md
```diff
@@ -0,0 +1,132 @@
...
+
+The current state of the `FungibleToken.Receiver` interface presents a challenge, as it does not offer a clear-cut method for determining the expected type of Vault that a particular receiver is able to receive. This issue is compounded by the fact that proxy receivers, which implement the `FungibleToken.Receiver` interface, can also return `Capability<&{FungibleToken.Receiver}>`. As a result, the true nature of the receiving Vault, whether it be derived from the standard Vault or a proxy receiver, remains ambiguous.
```
joshuahannan, line 17:
```suggestion
The current `FungibleToken.Receiver` interface only contains the `deposit` method. It does not offer a clear-cut method for knowing the `Vault` types that a particular receiver is able to receive.
This issue is compounded by the fact that proxy receivers, which implement the `FungibleToken.Receiver` interface, can also return `Capability<&{FungibleToken.Receiver}>`.
As a result, the true nature of the receiver, whether it be derived from the standard Vault or a proxy receiver, remains ambiguous.
```

flips/20230206-fungible-token-vault-type-discovery.md
```diff
@@ -0,0 +1,132 @@
...
+
+## User Benefit
+
+The revelation of the Vault type from the receiver capability has the potential to streamline the FT receiving process and minimize transaction failures. This is because developers now have the ability to programmatically verify the compatibility between the given receiver capability and the intended Vault type, allowing them to take proactive measures accordingly.
```
joshuahannan, line 21:
```suggestion
Adding type information to `Receiver` capabilities has the potential to streamline the FT receiving process and minimize transaction failures.
This is because developers now have the ability to programmatically verify the compatibility between the given receiver capability and the intended Vault type,
allowing them to take proactive measures accordingly such as depositing the tokens elsewhere instead of the transaction just failing.
```

flips/20230206-fungible-token-vault-type-discovery.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Design Proposal
+
+The essence of this proposal lies in modifying the `FungibleToken.Receiver` interface to furnish the expected type of receiving `Vault` and necessitate that all `FungibleToken.Vault` and custom receivers abide by this implementation.
```
joshuahannan, line 25:
```suggestion
The essence of this proposal lies in modifying the `FungibleToken.Receiver` interface to include a method to return the expected `Vault` types that it can receive and include a default implementation so that the upgrade will not be breaking for existing fungible tokens.
```
It won't require that `Vaults` or `Receivers` abide because we are including a default implementation

flips/20230206-fungible-token-vault-type-discovery.md
```diff
@@ -0,0 +1,132 @@
...
+        ///
+        pub fun deposit(from: @Vault)
+
+        /// Returns the type of implementing resource i.e If `FlowToken.Vault` implements
+        /// this then it would return `[Type<@FlowToken.Vault>()]`
+        ///
+        /// @return Optional list of vault types.
+        /// 
+        pub fun getExpectedVaultTypes() :[Type] {
+            return [self.getType()]
```
joshuahannan, line 51: Have you tested this? I think this would just return `FungibleToken.Receiver`, right? or does it return the actual underlying type?

flips/20230206-fungible-token-vault-type-discovery.md
```diff
@@ -0,0 +1,132 @@
...
+        /// @param from: The Vault resource containing the funds that will be deposited
+        ///
+        pub fun deposit(from: @Vault)
+
+        /// Returns the type of implementing resource i.e If `FlowToken.Vault` implements
+        /// this then it would return `[Type<@FlowToken.Vault>()]`
+        ///
+        /// @return Optional list of vault types.
+        /// 
+        pub fun getExpectedVaultTypes() :[Type] {
```
joshuahannan, line 50:
```suggestion
        pub fun getSupportedVaultTypes() :[Type] {
```

joshuahannan (commented):
Great proposal. I just made some wording changes to be a little more clear. From now on, can you please break up your paragraphs into multiple lines so we can more easily give comments? Thank you!

flips/20230206-fungible-token-vault-type-discovery.md
satyamakgec: Yeah I have done test for this and it is working as expected. It returns the Vault type like `Type<@FlowToken.Vault>()`, instead of Receiver type.

joshuahannan: thats good news. it would be really nice to have this by default so that token implementations don't have to upgrade to support it, but it doesn't really make sense for custom receivers. Is there a way to ensure that types are only included in this array if they are `FungibleToken.Vault` types?

satyamakgec: Yeah, for custom receivers, Sadly there is no way to provide the correct default implementation, Let me do some testing or digging whether it is possible or not.

alilloig:
@joshuahannan should we change on the switchboard 'getExpectedVaultTypes` for `getSupportedVaultTypes`? It does sound better indeed

joshuahannan:
Once this FLIP is completed and upgraded, we'll need to update the swtichboard anyway to include `getSupportedVaultTypes` but we shouldn't remove `getExpectedVaultTypes` because we don't want to break any other code unnecessarily.

flips/20230206-fungible-token-vault-type-discovery.md
satyamakgec: so I think it is possible to do so.

```
pub fun getSupportedVaultTypes(): [Type] {
      // Below check is implemented to make sure that run-time type would
      // only get returned when the parent resource conforms with `FungibleToken.Vault` 
      if self.getType().isSubtype(of: Type<@FungibleToken.Vault>()) {
          return [self.getType()]
      } else {
          return []
      }
 }
```

joshuahannan: That is great! Lets get that tested to make sure it works and include it in the proposal

flips/20230206-fungible-token-vault-type-discovery.md
```diff
@@ -0,0 +1,155 @@
...
+meaning that a Flow Vault can only receive Flow tokens and a FUSD Vault can only receive FUSD tokens.
+The `FungibleToken.Receiver` interface, on the other hand, is designed to be able to accept any Vault type in its `deposit` method.
+Currently, there is no programmatic way to determine the type of token that a `Receiver` is able receive, leading to the possibility of a failed deposit due to an incorrect assumption about the Vault that the `Receiver` is linked to.
+To address this issue, a new proxy receiver, the [`FungibleTokenSwitchboard`](https://github.com/onflow/flow-ft/blob/master/contracts/FungibleTokenSwitchboard.cdc), was created to allow for the receipt of multiple Vault types through a single `FungibleToken.Receiver` capability.
+The [`SwitchboardPublic`](https://github.com/onflow/flow-ft/blob/4416bbe585629671d00d3acfa6fd8052104dd861/contracts/FungibleTokenSwitchboard.cdc#L37) interface also includes the `getVaultTypes` function, which solves the problem of knowing what vaults the receiver accepts, but only for the switchboard and only for code that is aware of the switchboard.
```
franklywatson, line 22:
```suggestion
The design of the Fungible Token standard offers no built-in way to determine the token type that a `Receiver` handles. This leaves an unresolved edge case where deposits may fail due to incompatible Vault types being linked to the `Receiver`.
To address this, a new proxy receiver, the [`FungibleTokenSwitchboard`](https://github.com/onflow/flow-ft/blob/master/contracts/FungibleTokenSwitchboard.cdc), was created to handle multiple Vault types through a single `FungibleToken.Receiver` capability.
The [`SwitchboardPublic`](https://github.com/onflow/flow-ft/blob/4416bbe585629671d00d3acfa6fd8052104dd861/contracts/FungibleTokenSwitchboard.cdc#L37) interface provides the `getVaultTypes` function to answer which vaults the receiver accepts, however, this only applies to code in the switchboard.
```

joshuahannan:
@satyamakgec What is the status of this? I'd love to get this finished and upgraded ASAP. Have you been able to work on the implementation in the flow-ft repo?

franklywatson:
@joshuahannan I checked with Satyam earlier and this is his next priority

flips/20230206-fungible-token-vault-type-discovery.md
```diff
@@ -0,0 +1,155 @@
...
+        /// this then it would return `{Type<@FlowToken.Vault>(): true}` and if any custom receiver
+        /// uses the default implementation then it would return empty dictionary as its parent
+        /// resource doesn't conform with the `FungibleToken.Vault` resource.
+        ///
+        /// @return Dictionary of supported vault types by the implemented resource.
+        /// 
+        pub fun getSupportedVaultTypes() :{Type: Bool} {
+            // Below check is implemented to make sure that run-time type would
+            // only get returned when the parent resource conforms with `FungibleToken.Vault`. 
+            if self.getType().isSubtype(of: Type<@FungibleToken.Vault>()) {
```
dryruner, line 68: Receiver actually checks the FT type in the pre-condition in L219: https://f.dnz.dev/0x9a0766d93b6608b7/FungibleToken.

Also with the pre-condition check, FT.receiver cannot take multiple types of vaults iirc, which also makes total sense ...

joshuahannan: That check is happening in the `Vault.deposit` method, which makes it work for only Vault implementations. any resource can implement `Receiver`, and the deposit method in `Receiver` doesn't care about what kind of Vault it is

satyamakgec:
@joshuahannan can you please approve and merge this as it is already implemented and deployed. Do you think of any place where we need to update the documentation to reflect this change ?

joshuahannan (approved):

PR #54, opened 2022-12-16
FLIP for entitlements and safe reference downcasting

dsainati1:
This adds a FLIP proposing a Cadence change adding a new entitlement construct, a new access control modifier using entitlements, and changing the behavior of references to permit safe downcasting.

PR #72, opened 2023-02-23
FLIP - AuthAccount Capabilities Management Standard

sisyphusSmiling:
Forum Post: https://forum.onflow.org/t/account-linking-authaccount-capabilities-management/4314

This FLIP introduces a proposed standard for how AuthAccount Capabilities are managed and brings into discussion a number of open questions around the topic of account linking.


application/20230223-auth-account-capability-management-standard.md
```diff
@@ -0,0 +1,825 @@
...
+- [Objective](#objective)
+    - [Non-goals](#non-goals)
+    - [Essential](#essential)
+    - [For Consideration](#for-consideration)
+- [Existing Work](#existing-work)
+- [Motivation](#motivation)
+- [Design Proposal](#design-proposal)
+- [Example Implementation](#example-implementation)
+- [Considered For Inclusion](#considered-for-inclusion)
+    - [Standardizing child accounts’ resources access](#standardizing-child-accounts-resources-access)
```
louisguitton, line 26:
```suggestion
    - [Storage Iteration Convenience Methods](#storage-iteration-convenience-methods)
```

louisguitton (commented):
impressive work 👏  the demo during the hackathon was really interesting and I'll need to find the time to properly read through this. For now I've found this typo.

lukaracki:
The events look good, I'm just wondering what the benefit is of having a create account method in the Manager and Creator resource?

application/20230223-auth-account-capability-management-standard.md
```diff
@@ -0,0 +1,857 @@
...
+
+With this in mind, the following events and values have been proposed, though additional feedback is requested as the hope is these events can be as helpful as possible to dApps relying on this shared standard.
+
+- **Linking Accounts & Removing Linked Accounts** - Whenever an account is added as a child of a parent account, an event is emitted denoting both sides of the link. It may be helpful to include values from the child account's `ChildAccountTag.info` metadata struct to better identify the linked account.
+    ```js
+    pub event AccountAddedAsChild(parent: Address, child: Address)
+    pub event ChildAccountRemoved(parent: Address, child: Address)
+    ```
+
+- **Account Creation** - Emitted when an account is created from either the `ChildAccountManager` or `ChildAccountCreator` `createChildAccount()` methods. Since this method takes the arguments `signer: AuthAccount`, `initialFundingAmount: UFix64`, and `childAccountInfo: ChildAccountInfo`, there is room to include more information that may be relevant to callers, namely values from the `childAccountInfo` metadata struct.
```
lukaracki, line 424: I think having the originating publicKey as one of the params of these two events would be very beneficial.

lukaracki: Something like
```
pub event ChildAccountCreatedFromManager(parent: Address, child: Address, originatingPublicKey: String)
pub event AccountCreatedFromCreator(creator: Address?, newAccount: Address, originatingPublicKey: String)
```

sisyphusSmiling: That's a good suggestion. I believe you raised the issue in the Open House yesterday - would this event data allow dApps to determine the relevant creation event?

lukaracki: Yea I did raise it yesterday, I think a combination of the creator address, which you already specified, and the originatingPublicKey would allow dApps to determine the events that they care about.

sisyphusSmiling: Great, thanks for the helpful feedback! I'll add this now

application/20230223-auth-account-capability-management-standard.md
```diff
@@ -0,0 +1,868 @@
...
+
+/// Anyone holding this resource can create accounts, keeping a mapping of their
+/// originating public keys to their addresses. These accounts can then later
+/// be associated with a parent account by wrapping the ChildAccountTag and AuthAccount
+/// Capabilities in a ChildAccountController & saving in the parent account's
+/// ChildAccountManager
+/// 
+pub resource ChildAccountCreator : ChildAccountCreatorPublic {
+    /// mapping of public_key: address
+    access(self) let createdChildren: {String: Address}
```
lukaracki, line 393: Isn't this a cadence anti-pattern?

Is this field really necessary if people can already query the events to find out which account has created which children? 

I can imagine after a dApp creates a big number of accounts this dictionary will not be able to take new inputs. So we would need this to be sharded or it's not really scalable.

lukaracki: @sisyphusSmiling Right, that's very true, imo it would be a much cleaner solution if the ChildAccount contract did not include any account creation methods at all, and only have the Child -> Parent account linking features.

Then the dapps/users could handle account creation in their own custom ways and only use the ChildAccount contract for the purpose of linking and managing child(parent) accounts.

sisyphusSmiling:
Quick update on progress over the week as it's not visible in the contents of this FLIP:

We responded to concerns raised by community developers related to sharing user-controlled access on app-custodied accounts. The proposed solution builds on this FLIP, though works to separate concerns between user and developer side of the house with this FLIP solving the user side of the equation.

Essentially, the proposed solution, outlined in [this forum post](https://forum.onflow.org/t/hybrid-custody-restrictions-on-linked-accounts/4561), enables developers to define restricted access on the linked account which takes effect in the same transaction which links user and app accounts. The user then has unrestricted access while the app retains access to only the Capabilities needed to act on behalf of the user in the context of the dapp.

Moving forward, the focus for this week is increasing test coverage of the implementation contract & supporting transactions to ensure sound security. Additionally, I'll be working to break down concepts related to account linking & hybrid custody that will help devs get up to speed quickly.

Please check out the linked forum post, appreciate it all!

sisyphusSmiling:
Adding another quick update here now that we've aligned on an access model:

After lots of great discussion, the proposal for an unrestricted user/restricted app access model is being flipped in favor of a restricted user/unrestricted app access model. Check out the [flowtyio/restricted-child-account repo](https://github.com/Flowtyio/restricted-child-account) for work implementation level details and work on this front moving forward.

This is due to a number of factors including:
- prioritizing ease of app implementation
- minimizing potential foot guns for app developers
- interoperability with existing app custodial infrastructure and common practices
- need to return Typed Capabilities (very cool work with @austinkline's CapabilityFactory to enable this)

You can find the discussion in the previously linked [forum post](https://forum.onflow.org/t/hybrid-custody-restrictions-on-linked-accounts/4561) and [meeting notes](https://docs.google.com/document/d/1uw6pKUVmP5Jr6rwbSj5XF0h4-4H79TV-WQl1CNuJIfE/edit?usp=sharing).

Remaining open questions:
- Specifics around event data emitted
- Rules around multi-tenancy (IOW managing multiple parents & rule sharing)
- Planning for upgradability with upcoming AuthAccount entitlements - #54 (cc: @turbolent @dsainati1)
- Decide how we want to handle access promotion from restricted -> unrestricted user access

Next steps:
- Update this FLIP with new access model overview
- Feedback and iteration on [HybridCustody contract PR](https://github.com/Flowtyio/restricted-child-account/pull/14)
- Submit PR with suggested events as conversation point on the issue

austinkline:
Thanks for add a summary @sisyphusSmiling! Adding some notes about the current PR as it pertains to outstanding issues

> Rules around multi-tenancy (IOW managing multiple parents & rule sharing)

How it works in the active PR is that a `ChildAccount` resource creates and manages underlying `ProxyAccount` resources which each have their own rules managed by the account's owner. Parents are shared a capability to a proxy only meant for them. Even if the mechanisms in place are complex, the owner of the child account only needs to know about a few methods to add, revoke, or edit existing parents (edits aren't done yet)

> Decide how we want to handle access promotion from restricted -> unrestricted user access

The `ChildAccount` has a few methods to change an account's ownership which would then alter restricted -> unrestricted access. Mainly, there are two forms of modified ownership:
1. Transferred Ownership - The current owner of a child account gives ownership of the account to another wallet. This will revoke all outstanding keys on the wallet, unlink all existing AuthAccount capabilities, and make a new one for existing parent/child accounts to keep operating as usual. The new owner gets a capability to the ChildAccount which they can then access as they please. This will give them the full AuthAccount capability which means they have unrestricted access
2. Relinquished Ownership - Needs a new name, probably. But the idea here is that no one fully owns the account. An unowned account could be jointly shared between two parents but neither can change the access another one has. 


Here is a diagram for how the current PR works:
HybridCustody-diagram

btspoony:
Some thoughts and implementation from my side.

https://github.com/onflow/hybrid-custody/pull/140

Feel free to further discuss.

sisyphusSmiling:
FYI This FLIP has been updated to reflect the current state of [HybridCustody contracts](https://github.com/onflow/hybrid-custody) currently [deployed to Testnet](https://f.dnz.dev/0x294e44e1ec6993c6/HybridCustody) and [proposed for Mainnet](https://forum.onflow.org/t/hybrid-custody-mainnet-deployment-plan/5050/).

Please take a look and leave your thoughts/reviews.

joshuahannan (approved):

application/20230223-auth-account-capability-management-standard.md
```diff
@@ -0,0 +1,1201 @@
...
+- As a user, why would I want a separate account I share access with? Doesn’t that put all of my assets at custodial risk?
+    - Again, only the child account shares access with another party, meaning your main account is safe from custodial risk. In fact, partitioning assets across accounts in this way enhances security over a model that requires all transactions be signed by your main account. A user can keep all of their more valuable assets in their main account, out of reach without a user-signed transaction, while keeping less valuable dApp assets in a shared account for ease of use.
+- As an application developer, won't I expose myself to undo risk by giving a user access on an account I custody?
+    - The newly proposed design introduces the ability to restrict delegated access. This means that you can set the rules on what a user can access via the delegation you grant them, thereby setting their scope as you define it. For example, want users to be able to access an NFT Collection in your app-custodied account? That can be easily configured!
+
```
franklywatson, line 134: - This standard and design introduce a lot of complexity. Could this not have been solved in other ways, such as through the use of keys or other approaches?
    - There were a number of previous iterations and designs that came before the Restricted Child Account proposal from Flowty and can be found in [Alternatives Considered](#alternatives-considered) where the issues and limitations of those approaches are detailed.

PR #77, opened 2023-03-31
FLIP - Interaction Templates v1.1.0

JeffreyDoyle:
# FLIP - Interaction Templates v1.1.0

## Overview

This update to the InteractionTemplate and InteractionTemplateInterface data structures is the result of the need to better define the order of various properties of the InteractionTemplate and InteractionTemplateInterface, for use during the serialization process required for ID generation. 

This change uses JSON Arrays to express the order of certain elements of the InteractionTemplate and InteractionTemplateInterface, instead of Objects as we're previously used which do not have a notion of order.

## Required updates

- [ ] FCL Interaction Template utils updated to support v1.1.0 of InteractionTemplate and InteractionTemplateInterface (https://github.com/onflow/fcl-js/pull/1614)
- [ ] FCL Query/Mutate updated to support v1.1.0 of InteractionTemplate
- [ ] [FLIX Interaction Templates](https://github.com/onflow/flow-interaction-template-service) migrated to version v1.1.0 
- [ ] [Interaction Template Tools](https://github.com/onflow/flow-interaction-template-tools) (CLI, WebApp and JS Utils) updated to support version v1.1.0


bjartek:
Should we consider changing the way pins are generated as well here to be debth first since we are creating a new version?

I think the pin should be sha3(your_code, sha3(dep1), sha3(dep2)) or something like that so if you have done sha3(dep1) you can memoize that for all FLIX. Or even cache it.

sideninja:
@m-Peter I think this could be an improvement to an existing solution that is currently lacking. The way I see is we have to solve this problem for any transaction or script, then after we solve the mapping we can improve the mapping for flix templates by integrating into flix. I don't see the other way around to be better.

bthaile:
Support new import syntax: In order to be consistent across all flow tools, it would be beneficial for flix to support new import syntax. This syntax does not have a contract address placeholder in the cadence code. It uses `import "ContractName"` which is interpreted by flow tools. The change needed is in the "dependencies" section, the contract address placeholder isn't needed. See below:

Pervious version:
```
dependencies: {
      "0xFUNGIBLETOKENADDRESS": { // Network (mainnet || testnet) dependent locations for 0xFUNGIBLETOKENADDRESS contract.
          "FungibleToken" : {
              mainnet: {
                  address: "0xf233dcee88fe0abe", // Address of the account the contract is located.
                  fq_address: "A.0xf233dcee88fe0abe.FungibleToken", // Fully qualified contract identifier.
                  pin: "asdfasdfasdfasdfasdfasdfsadf123123123123", // Unique identifier of the interactions dependency tree.
                  pin_block_height: 10123123123 // Block height the pin was generated against.
              },
              testnet: {
                  address: "0x9a0766d93b6608b7",
                  fq_address: "A.0x9a0766d93b6608b7.FungibleToken",
                  pin: "asdfasdfasdfasdfasdfasdfsadf123123123123",
                  pin_block_height: 10123123123
              }
          }
      }
  },
```

New version:
```
dependencies: {
    "FungibleToken" : {
        mainnet: {
            address: "0xf233dcee88fe0abe", // Address of the account the contract is located.
            fq_address: "A.0xf233dcee88fe0abe.FungibleToken", // Fully qualified contract identifier.
            pin: "asdfasdfasdfasdfasdfasdfsadf123123123123", // Unique identifier of the interactions dependency tree.
            pin_block_height: 10123123123 // Block height the pin was generated against.
        },
        testnet: {
            address: "0x9a0766d93b6608b7",
            fq_address: "A.0x9a0766d93b6608b7.FungibleToken",
            pin: "asdfasdfasdfasdfasdfasdfsadf123123123123",
            pin_block_height: 10123123123
        }
    }
},
```

flips/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1053 @@
...
+
+#### `data.cadence`
+
+The cadence code the interaction executes.
+
+#### `data.dependencies`
+
+For each dependency of the interaction (each contract that is imported in the cadence of the interaction), there must be network keyed (mainnet || testnet) dependency information. The information for each network should contain the address of the account where the contract is deployed, the fully qualified identifier for the contract, dependency tree pin and block height the pin was preformed at. The dependency tree pin is performed by the following pseudocode:
+
+```javascript
```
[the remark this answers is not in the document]
bthaile: Changed the example to be depth first

bthaile:
todo: change "argument(s)" to "parameter(s)"

JeffreyDoyle (commented):
Updates according to feedback from the Interaction Template working group on August 14 2023

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1103 @@
...
+
+#### `data.dependencies`
+
+For each dependency of the interaction (each contract that is imported in the cadence of the interaction), there must be network keyed (mainnet || testnet) dependency information. The information for each network should contain the address of the account where the contract is deployed, the fully qualified identifier for the contract, dependency tree pin and block height the pin was preformed at. The dependency tree pin is performed by the following pseudocode:
+
+```javascript
+let contract_imported_in_interaction_cadence = ...
+let import_hash = ""
+
+function processContract(contract) {
```
bjartek, line 382: I think there is a problem with this psudo code, when recalculating a pin for a single chaged contract this leads to a lot on unneccesary work. 

Lets take the ExampleNFT in the flow-nft repo as an example. It has the following hirarchy
```
ExampleNFT
 - NonFungibleToken
 - MetadataViews
  - FungibleToken
  - NonFungibleToken
 - ViewResolver
```
With the current algorithm if ExampleNFT changes you have to nest all the way down into the dependencies of MetadataViews to calculate the pin. 

What i propose instead is a recursive algorithm like this:
 ```

 function generateDependentPin(service, contract, contracts) {
    // Generate pin for self if not already generated
    if (!contract.selfPin) {
        contract.selfPin = shaHex(contract.body);
    }

    // Initialize an array to store pins
    const pins = [contract.selfPin];

    // Iterate through each import (dependency) in the imports list of the current contract
    for (const imp of contract.imports) {
        // Recursively generate dependent pins for the imported contract
        const depPins = generateDependentPin(service, imp, contracts);
        // Append the dependent pins to the pins array
        pins.push(...depPins);
    }

    // Calculate a pin for the current contract by applying the shaHex function to the concatenated pins
    contract.pin = shaHex(pins.join(''));

    // Add the updated contract to the contracts object using its key
    contracts[contract.contractKey()] = contract;

    // Return the array of pins
    return pins;
}

```

The main difference here is that once you have calculated the pin for a contract and its dependencies that contract can be reused until it changes.

**Review by bjartek — CHANGES_REQUESTED** [review]

File: protocol/20230330-interaction-templates-1.1.0.md (hunk 2 above)
Comment (bjartek) [reply protocol/20230330-interaction-templates-1.1.0.md:382]: the algorithm above needs to first delete the pin of a changed contract before that code is called.

Comment (bjartek) [reply protocol/20230330-interaction-templates-1.1.0.md:382]: I have now implemented this, this is an example row of how this works for a niftory based NFT

```{"pin": "7701e8bd1d17f32498edcced9afa7dfa5cd198be40c2c7c993897569261746dd", "pinParts": ["e788e78352b04a38495cef43bab10e1d7ddf8d6ba0af86cdf36c6381c29ed934", "b2ab6c499be905685f1b34cfd0e6fb5740be90421589a9a230bc6b193d813b74", "d0605dd3cc79f9c0c8e68dbda9e18f7ef57302ab1fac83481b5c9817dcabb1b7", "db3aef2f72d1096cb7f9ddcaad0c6692d770bda92fa4c202675d01d18abeb6fd", "16811e966e5ec36cc9df7aeb15e10ff58b82e18c63f57dabc5a452201d4089f6", "d816a566ae515d0bc5845fef8fb937bbb80315346aff9af6173f877e942da7ea", "38dcd652b06d634678bc0d009e28f1b7c149bcd503f1cad652fad65b0b54c495", "63c334bcdbad3b4b28a3843a11ab22ba1d36f141f978f1d80e5a7c6e4bd32f81", "40575f755c046e2851c8b97e8fcf31b0f3deb185f941063dc59960aec0210f6f", "5ff238dd74d2a80a28314f494ee17e11a4b56c0cd038309c3e8858d6e8c41cb9", "cf8a66a03aff94024434bf67a35143820db62e7faf5fd0529709f2310904764b", "987952d0f5aeaece55159bb059a80a4bae11abffc464e142c635fc8081f67b30", "ce611f41a7261389215901bcfa4871de4c8bfdbbf88e8d033e6e3b763e442433"], "contract": "A.2772aa34edded534.rojanscontract", "pinSource": ["A.2772aa34edded534.rojanscontract", "A.1d7e57aa55817448.NonFungibleToken", "A.1d7e57aa55817448.MetadataViews", "A.1d7e57aa55817448.ViewResolver", "A.7ec1f607f0872a9e.MutableMetadata", "A.7ec1f607f0872a9e.MutableMetadataTemplate", "A.7ec1f607f0872a9e.MutableMetadataSet", "A.7ec1f607f0872a9e.MutableMetadataSetManager", "A.7ec1f607f0872a9e.MetadataViewsManager", "A.7ec1f607f0872a9e.NiftoryNonFungibleToken", "A.7ec1f607f0872a9e.NiftoryNFTRegistry", "A.7ec1f607f0872a9e.NiftoryMetadataViewsResolvers", "A.7ec1f607f0872a9e.NiftoryNonFungibleTokenProxy"], "selfPin": "e788e78352b04a38495cef43bab10e1d7ddf8d6ba0af86cdf36c6381c29ed934"}
```

bjartek: Currently only the selfPin and pin are saved here, we do not save the pinParts or pinSource. note that order of pinParts/pinSource is the order of the imports in the file, that is why i did not make it a map, because order does matter here.

bjartek: psudo javascript code
```javascript
function generateDependentPin(contractKey, contracts) {
    const contract = contracts[contractKey];

    if (contract.Pin !== '') {
        return contract.Pin;
    }

    contract.SelfPin = shaHex(contract.Body);

    const pins = [contract.SelfPin];
    for (const imp of contract.Imports) {
        const dep = generateDependentPin(imp.ContractKey(), contracts);
        pins.push(dep);
    }

    contract.Pin = shaHex(pins.join(''));
    return contract.Pin;
}

// Example contracts map
const contracts = {
    contract1: {
        Body: '...',
        Imports: [...],
        Pin: '',
        SelfPin: ''
    },
    contract2: {
        Body: '...',
        Imports: [...],
        Pin: '',
        SelfPin: ''
    },
    // ... more contracts ...
};

// Compute pins for each contract
for (const contractKey in contracts) {
    generateDependentPin(contractKey, contracts);
}
```

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1385 @@
...
+
+Creating a mechanic that allows trusted entities to act as auditors, to produce a proof that vouches for the correctness and safety of a Interaction Template; and a mechanic for verifiers (applications, wallets, etc) to prove an audit was created by an auditor they trust, will improve the overall safety and security of the use of Interaction Templates on Flow.
+
+## Design Proposal
+
+### Interaction Interfaces
+
+Many interactions aim to achieve the same class of action according to how that action must be performed with a certain project. Classes of interactions may be things like: "Transfer", "Mint", "Bid", "List", "Destroy" etc.
+
+For example, different marketplaces may have different mechanics behind how a user may "List" their asset on them, but across all implementations, the action "List" is what they all aim to do.
```
bjartek, line 68: While i def see the end goal here, I think that ITI (InteractionTemplateInterfaces) are less usefull the the templates themselves. 

Because a lot different vendors will have diferent parameters to the same "type" of Action. Some marketplaces needs a storefrontAddress for instance to list while others do not. 

It does not hurt to have it here, but I dont see it beeing used very much unfortunately

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1385 @@
...
+          {
+            tag: "zh-CN",
+            translation: "签署此消息以转移FLOW。"
+          }
+        ]
+      }
+    ],
+    cadence: // Cadence code this interaction executes.
+    `
+    import "FlowToken"
```
bjartek, line 202: In this example we use the new format of imports, but I am not sure that is what the flixkit uses currently? I think this new format is the way to go personally.

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1385 @@
...
+                translation:  "Quantité de token FLOW à transférer"
+              },
+              {
+                tag: "zh-CN",
+                translation: "要转移的 FLOW 代币数量"
+              }
+            ]
+          }
+        ],
+        balance: "FungibleToken" // The token this parameter acts upon.
```
bjartek, line 312: This should be a fully qualified identifier to the token IMHO, or can it be? If this is not network specific it cant be.

protocol/20230330-interaction-templates-1.1.0.md
bjartek, line 202: This should import FungibleToken to be correct and not FlowToken?

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1385 @@
...
+              {
+                network: "testnet",
+                address: "0x7e60df042a9c0868",
+                dependency_pin_block_height: 10123123123, // Block height the pin was generated against.
+                dependency_pin: {
+                  pin: "c8cb7cc7a1c2a329de65d83455016bc3a9b53f9668c74ef555032804bac0b25b", // Unique identifier of this contract dependency and it's dependency tree tree.
+                  pin_self: "38d0cca4b74c4e88213df636b4cfc2eb6e86fd8b2b84579d3b9bffab3e0b1fcb", // Unique identifier of this dependency
+                  pin_contract_name: "FlowToken",
+                  pin_contract_address: "0x7e60df042a9c0868",
+                  imports: [
```
bjartek, line 255: Do we really need dependencies of dependencies here infinitely added? Is it not enough to have 1 level of direct imports and their pins?

protocol/20230330-interaction-templates-1.1.0.md
bjartek: Can we take a look at my alternate suggestion here? it really scales tons better.

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1385 @@
...
+    post {
+        getAccount(to).balance == self.receiverBalanceBefore + amount
+        getAccount(from).balance = self.senderBalanceBefore - amount
+    }
+}
+```
+
+For more on transaction phases, see [here](https://docs.onflow.org/cadence/language/transactions).
+
+#### Interaction Template Discovery
```
bjartek, line 646: I think this sections needs more fleshing out. In order for this solution to be really used by devs we need to standardize this a little more I think. Atleast provide a reference implementation on how to do the correct thing. So lets say i develop a contract put it on testnet and publish some templates. IMHO that should be as easy as doing

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1385 @@
...
+GET interactions.nft-project.com/purchase-nft -> InteractionTemplate
+```
+
+Consumers of this Interaction Template could query the static identifier, and always get the most up to date Interaction Template stored there.
+
+Repositories of Interaction Templates could be created to store Interaction Templates from various projects. Developers of Interaction Temaplates could choose to submit their Interaction Templates to such repositories.
+
+Systems for Interaction Template discovery could proxy requests for an Interaction Template to the repository it's known to exist in. Since Interaction Templates are just data, they could be cached and distributed amongst such repositories as desired.
+
+### Interaction Template Audits
```
bjartek, line 662: Are we trying to do to much in 1 flip here? A lot of these 2. day operations needs some more experimentation with the core flix before we really know how to make a workflow that works and scales for everybody.

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1385 @@
...
+
+## Dependencies
+
+Interaction Templates depend on contract developers producing and making available Interaction Templates for their contracts. Interaction Template Interfaces depend on the Flow developer community coming to consensus on interfaces for interactions they implement. Interaction Template Audits depend on trusted entities producing and making available audits of Interaction Templates.
+
+### Template, Interface and Audit Tooling
+
+To make the production of Interaction Template, Interaction Template Interface and Interaction Template Audits simpler for developers, support for generating these could be added to a CLI tool or webapp interface. Making these data structures easy to create will be essential for promoting this new pattern.
+
+### FCL Integration
```
bjartek, line 704: Should we mention flixkit and the bindings here?

bjartek (changes requested):

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1385 @@
...
+    dependencies: [
+      {
+        contracts: [
+          {
+            contract: "FlowToken",
+            networks: [
+              {
+                network: "mainnet",
+                address: "0x1654653399040a61", // Address of the account the contract is located.
+                dependency_pin_block_height: 10123123123 // Block height the pin was generated against.
```
bjartek, line 229: body_hash: hex.EncodeToString(sha256.Sum256(<resolved cadence>)

bthaile: "body_hash" is confusing since the fully population transaction code would having fully qualified imports. 
maybe a other name would be better, "fully_qualified_hash"

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1385 @@
...
+                  pin_self: "38d0cca4b74c4e88213df636b4cfc2eb6e86fd8b2b84579d3b9bffab3e0b1fcb", // Unique identifier of this dependency
+                  pin_contract_name: "FlowToken",
+                  pin_contract_address: "0x1654653399040a61",
+                  imports: [
+                    {
+                      pin: "b8a3ed26c222ed67016a28021d8fee5603b948533cbc992b3c90f71a61b2b312", // Unique identifier of this contract dependency and it's dependency tree.
+                      pin_self: "7bc3056ba5d39d130f45411c2c05bb549db8ce727c11a1cb821136a621be27fb",  // Unique identifier of this dependency
+                      pin_contract_name: "FungibleToken",
+                      pin_contract_address: "0xf233dcee88fe0abe",
+                      imports: []
```
bthaile, line 241: Seems having an imports inside imports is over kill since the `pin` should have all the information needed.

protocol/20230330-interaction-templates-1.1.0.md
bthaile: I agree, will remove the second layer of `imports[]`

protocol/20230330-interaction-templates-1.1.0.md
bthaile: Let's call it `cadence_hash` using sha256.

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1385 @@
...
+            translation:  "Signez ce message pour transférer FLOW."
+          },
+          {
+            tag: "zh-CN",
+            translation: "签署此消息以转移FLOW。"
+          }
+        ]
+      }
+    ],
+    cadence: // Cadence code this interaction executes.
```
JeffreyDoyle, line 200: TODO:

Add cadence hashes feature here as decided in FLIX working group:

```javascript
// Top Level
...
"cadence": {
  "body": "...."
  "pins": [
    {
      "network": "mainnet",
      "pin": "abc123"
    },
    {
      "network": "testnet",
      "pin": "def456"
    }
  ]
}
...
```

This also needs to be added to the serialization spec.

JeffreyDoyle (commented):
Adding comments from working group Nov 9

JeffreyDoyle:
Hello all!

As per the FLIX Working Group's previous meeting on November 9th 2023, the group has aligned on the following updates to the FLIX specification to be included in version 1.1:

- Add support for the "new import syntax" in the cadence included in the Interaction Template 
- Add the full dependency tree inside the Interaction Template's dependency section.
- Use the depth first search algorithm proposed by @bjartek for computing dependency pins (see: https://github.com/onflow/flips/pull/77#discussion_r1302790294)
- Include hashes of the cadence included in the Interaction Template resolved per each supported network.

The following updates are reflected in the FLIP as of the date of this comment (Nov 14 2023). Please let us know if you have any thoughts or feedback!

-- 

If you're not in the FLIX working group and would like to be, please feel free to send an email to wallet-wg@flow.com and we will include you in the following meeting invites! 😄

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1451 @@
...
+                translation:  "Quantité de token FLOW à transférer"
+              },
+              {
+                tag: "zh-CN",
+                translation: "要转移的 FLOW 代币数量"
+              }
+            ]
+          }
+        ],
+        balance: "FungibleToken" // The token this parameter acts upon.
```
bjartek, line 323:
```suggestion
        balance: "FlowToken" // The token this parameter acts upon.
```

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1451 @@
...
+                translation: "到"
+              }
+            ]
+          },
+          {
+            key: "description",
+            i18n: [ // Internationalised (BCP-47) set of human readable messages about the parameter
+              {
+                tag: "en-US",
+                translation: "Amount of FLOW token to transfer", // Messages might consume parameters.
```
bjartek, line 523: This is wrong?

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1451 @@
...
+    output: {   // only needed for scripts
+      label: "balance",
+      type: "UFix64",
+      messages: [ // optionally, only needed to give clarification to consumers
+        {
+          key: "description",
+          i18n: [ 
+            {
+              tag: "en-US",
+              translation: "Amount of FLOW token to transfer",
```
bjartek, line 547: This is wrong?

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1451 @@
...
+#### `data.dependencies`
+
+For each dependency of the interaction (each contract that is imported in the cadence of the interaction), there must be network keyed (mainnet || testnet) dependency information. The information for each network should contain the address of the account where the contract is deployed, the fully qualified identifier for the contract, dependency tree pin and block height the pin was preformed at. The dependency tree pin is performed by the following pseudocode:
+
+```javascript
+let contract_imported_in_interaction_cadence = ...
+let import_hash = ""
+
+function processContract(contract) {
+  let contract_code = getContract(contract)
```
bjartek, line 621: this pseudo code is still not debt first is it?

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1451 @@
...
+
+Internationalized, human readable messages explaining each of the cadence parameters. For each message, there can be any number of translations provided. Translations should use [BCP 47](https://en.wikipedia.org/wiki/IETF_language_tag) language tags.
+
+Parameter may correspond to a balance of a fungible or identifier of a non-fungible token. In this case, the balance or identifier the parameter corresponds to should point to its dependency identifier.
+
+#### `data.output`
+
+Like data.parameters, human readable messages explaining the result of  a script. Output property is needed for clarification for consumers and for code generators. This will provide better integration.
+
+#### Arbitrary Execution Phase
```
bjartek, line 648: I honestly think that this FLIP covers so much and that this should be left for later.

bjartek (changes requested):

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1451 @@
...
+                .getCapability(/public/flowTokenReceiver)!
+                .borrow<&{FungibleToken.Receiver}>()!
+                .deposit(from: <-self.vault)
+            }
+        }
+        `,
+        pins: [
+        {
+          network: "mainnet",
+          pin: "186e262ce6fe06b5075ec6569a0e5482a79c471881182612d8e4a665c2977f3e"
```
wfalcon0x, line 223: In my opinion
`hash: "186e262ce6fe06b5075ec6569a0e5482a79c471881182612d8e4a665c2977f3e"`
could better describe what the value is.
The array is called `pins: [` , so users would understand each item in the collection describes a pin.

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1451 @@
...
+          pin: "f93977d7a297f559e97259cb2a95fed0f87cfeec46c5257a26adc26a260d6c4c"
+        }
+      ]
+    },   
+    dependencies: [
+      {
+        contracts: [
+          {
+            contract: "FlowToken",
+            networks: [
```
wfalcon0x, line 236: Using `pins: [` instead of `networks: [` would make it consistent with the naming in the cadence object.

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1451 @@
...
+              {
+                network: "mainnet",
+                address: "0x1654653399040a61", // Address of the account the contract is located.
+                dependency_pin_block_height: 10123123123 // Block height the pin was generated against.
+                dependency_pin: {
+                  pin: "c8cb7cc7a1c2a329de65d83455016bc3a9b53f9668c74ef555032804bac0b25b", // Unique identifier of this contract dependency and it's dependency tree.
```
wfalcon0x, line 242: In my opinion
hash: "c8cb7cc7a1c2a329de65d83455016bc3a9b53f9668c74ef555032804bac0b25b"
could better describe what the value is. See above...

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1451 @@
...
+                  pin_self: "38d0cca4b74c4e88213df636b4cfc2eb6e86fd8b2b84579d3b9bffab3e0b1fcb", // Unique identifier of this dependency
```
wfalcon0x, line 243: In my opinion
hash_self: "38d0cca4b74c4e88213df636b4cfc2eb6e86fd8b2b84579d3b9bffab3e0b1fcb"
could better describe what the value is. See above...

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1451 @@
...
+                  pin_contract_name: "FlowToken",
```
wfalcon0x, line 244: As the object is already called `dependency_pin: {` above, it is sufficient to call this value just `contract_name`

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1451 @@
...
+                  pin_contract_address: "0x1654653399040a61",
```
wfalcon0x, line 245: As the object is already called `dependency_pin: {` above, it is sufficient to call this value just `contract_address`

protocol/20230330-interaction-templates-1.1.0.md
wfalcon0x: I think it should be fully qualified identifier, as it is possible to create any contract with the same name under a different account. It would mean this value needs to have separate value for the different networks.

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1451 @@
...
+    [ ...template-output-content-message-translation ]
+]
+template-output-content         = [
+    sha3_256(template-output-content-type),
+    [ ...template-output-content-message ]
+]
+template-output-label         = Label for an output
+template-output               = [ sha3_256(template-output-label), [ ...template-output-content ]]
+
+template-cadence-networks     = [
```
JeffreyDoyle, line 890:
```suggestion

template-cadence-body-resolved   =  The cadence body for the transaction/script with imports resolved for the given network
cadence-pins-network.                     =  The network name for this pin ("mainnet" | "testnet" | "emulator")
cadence-pins-pin                               =  sha3_256(template-cadence-body-resolved )

template-cadence-networks     = [
  sha3_256(cadence-pins-network), 
  cadence-pins-pin,
]

template-cadence-body          = The body of the cadence transaction or script with string imports

template-cadence-content      =  [
  sha3_256(template-cadence-body ),
  [...template-cadence-networks]
]

```

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1451 @@
...
+- Do viable alternatives to this proposal exist?
+- Could the emergence of use of common Cadence contract & resource interfaces across projects reduce the ecosystems dependency on this solution?
+
+## Data Structure Serialization & Identifier Generation
+
+A deterministic serialization algorithm is required to be applied prior to hashing its result to produce identifiers for the `InteractionTemplate` and `InteractionTemplateInterface` data structures.
+
+By serializing each data structure into a specific format, then [RLP encoding](https://ethereum.org/en/developers/docs/data-structures-and-encoding/rlp/) that format, then hashing that encoding, we can generate identifiers for each data structure.
+
+### `InteractionTemplate` f_version 2.0.0
```
JeffreyDoyle, line 815:
```suggestion
### `InteractionTemplate` f_version 1.1.0
```

protocol/20230330-interaction-templates-1.1.0.md
JeffreyDoyle: Having all of the dependencies directly in the Interaction Template will help wallets identify, if it recomputes an Interaction Template's dependency pin and it doesn't match what's in the Interaction Template, which dependencies in the tree changed to cause the discrepancy.

protocol/20230330-interaction-templates-1.1.0.md
```diff
@@ -0,0 +1,1467 @@
...
+  id: "a2b2d73def...aabc5472d2", // Unique ID for the data structure.
+  data: {
+    type: "script", // "transaction" || "script"
+    messages: [
+      {
+        key: "title",
+        i18n: [ // Internationalised (BCP-47) set of human readable messages about the interaction
+          {
+            tag: "en-US",
+            translation: "Get FLOW Balane"
```
JeffreyDoyle, line 385:
```suggestion
            translation: "Get FLOW Balance"
```

protocol/20230330-interaction-templates-1.1.0.md
JeffreyDoyle: Implemented in the FLIP!

protocol/20230330-interaction-templates-1.1.0.md
JeffreyDoyle: Todo: Address Interaction Template Interface in further working group convos

protocol/20230330-interaction-templates-1.1.0.md
JeffreyDoyle: Implemented in the FLIP

protocol/20230330-interaction-templates-1.1.0.md
JeffreyDoyle: Per FLIX working group: Let's leave updating this section for a future discussion

protocol/20230330-interaction-templates-1.1.0.md
JeffreyDoyle: Per FLIX working group: Let's work on this in a future discussion.

protocol/20230330-interaction-templates-1.1.0.md
JeffreyDoyle: Per FLIX working group: Let's work on this in a future discussion.

bthaile (approved):
congrates, looks good

gregsantos (approved):
lg2G

PR #78, opened 2023-04-03
FLIP78- Delegator Staking Minimum

joshuahannan:
Proposes adding a minumum stake for delegators to receive rewards

### DIFF
--- a/protocol/20230403-delegator-minimum.md
+++ b/protocol/20230403-delegator-minimum.md
@@ -1,0 +1,124 @@
+---
+status: Implemented
+flip: FLIP-XX
+authors: Joshua Hannan (joshua.hannan@dapperlabs.com)
+sponsor: Joshua Hannan (joshua.hannan@dapperlabs.com) 
+updated: 2023-04-03
+---
+
+# Delegator Staking Minimum
+
+## Objective
+
+Introduce a minimum required stake (100 FLOW) for rewards
+for staking delegators in the Flow protocol and staking smart contracts.
+
+## Motivation
+
+Delegators were proposed as an easy way to participate
+in the security of the network and earn staking rewards
+without having to run a node or commit much capital for extended periods of time.
+
+While delegators aren't directly critical to the core functioning
+of the network, as a group, they provide a signaling function of trusted nodes
+and trust in the network generally. Delegators have been staked
+in the identity table and choose a specific node to delegate to.
+They can delegate any amount of FLOW and will receive rewards proportional to the 
+amount of FLOW they delegate.
+Unfortunately, the low barriers to entry take a toll on the network. 
+Since there are so many delegators in the identity table, epoch transition operations
+such as moving tokens between buckets and paying rewards are costly
+and time-consuming. 
+
+Currently, there are over 16k delegators in the network, which means that nearly
+50k events are emitted for each delegator when their rewards are paid because each delegator
+gets a `FlowToken.TokensWithdrawn`, `FlowToken.TokensDeposited`, and `FlowIDTableStaking.DelegatorRewardsPaid`
+event. Over 30k of those are for amounts less than 0.2 FLOW, which is very wasteful for such a small amount.
+
+You can find these amount by using this command on the command line:
+
+```
+curl -s https://rest-mainnet.onflow.org/v1/transaction_results/84eca4ff612ef70047d60510710cca872c8a17c1bd9f63686e74852b6382cc84 \
+| jq '.events | map(select( .type == "A.1654653399040a61.FlowToken.TokensDeposited" )) | map(.payload | @base64d | fromjson) | map(select((.value.fields[0].value.value | tonumber) > 0.2)) | length'
+```
+You'll need to multiply the result by three to get the total number of events for each
+delegator above the specified rewards amount.
+
+This FLIP proposes increasing the delegator staking minimum
+to receive rewards to 100 FLOW,
+to avoid more unnecessary iterations and events for small staking and reward amounts.
+
+This FLIP acknowledges that delegating even small amounts of FLOW is
+important for new users to get an easy win and feel some ownership 
+in the network, so it does not enforce this minimum to stake, only to receive rewards,
+because that is where this problem is the most costly.
+
+The number of staked FLOW proposed for minimum rewards was chosen because it is the same as access nodes
+and it currently will cut down the number of rewards events emitted by almost two-thirds.
+It is also still affordable at current FLOW prices.
+With current rewards calculations, this means that the minimum amount of rewards
+that a delegator can receive in an epoch is 0.2 FLOW.
+
+This value can definitely change in the future if network and economic changes require it.
+
+## User Benefit
+
+This change would benefit everyone using the network because the rewards transaction will
+no longer take multiple seconds to execute.
+
+It will also benefit applications who are monitoring the network for events because it will drastically
+reduce the number of events in this transaction, making it easier to parse and store.
+
+## Design Proposal
+
+In `FlowIDTableStaking.calculateRewards()`, add a conditional that only 
+calculates rewards for a delegator if their `tokensStaked` is greater than 100 FLOW.
+
+The service account committee runs a transaction, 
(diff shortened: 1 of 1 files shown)


protocol/20230403-delegator-minimum.md
```diff
@@ -0,0 +1,124 @@
...
+delegator above the specified rewards amount.
+
+This FLIP proposes increasing the delegator staking minimum
+to receive rewards to 100 FLOW,
+to avoid more unnecessary iterations and events for small staking and reward amounts.
+
+This FLIP acknowledges that delegating even small amounts of FLOW is
+important for new users to get an easy win and feel some ownership 
+in the network, so it does not enforce this minimum to stake, only to receive rewards,
+because that is where this problem is the most costly.
```
jordanschalm, line 54: I agree that being able to delegate a small amount is important to make participation have a low barrier. However I doubt many people will want to participate if their delegation doesn't do anything.

We will also need to provide guidance in UIs like in Flow Port that, yes you _can_ delegate, but it will have no effect for values below X. Also X might change from time to time, so you might delegate an amount and receive rewards for a while, then suddenly stop receiving rewards. Seems like a recipe for confusion for exactly those users who are delegating amounts in this range, and who we are designing this proposal to accomodate.

AlexHentschel (approved):
Thanks, I like this. A few thoughts on possible degrees of freedom we could use:
* I think [Jordan has a really good point](https://github.com/onflow/flips/pull/78#discussion_r1156272668): 
   * How do we deal with the fact that the threshold might change and avoid confusion?  
   * We have a similar situation if you delegated to a node that drops out of the network. In this case, we simply just return the delegated tokens. 
   
       I think we could do the same with the delegated tokens below the threshold: return them and do not accept any new delegations below the threshold.
* I think 100 Flow is comparatively large threshold. What if it was 10Flow? Below that amount, we just don't accept any delegation. And you get rewards immediately.


I think this would have 2 benefits with nice synergies:
* We remove data from the system smart contract. 
   * If I understand correctly, in your approach, we would primarily reduce the number of events for small delegations below the threshold. Nevertheless, the computational cost of storage cost remains, as we have to maintain the data and probably iterate over it. 
* We make additional UX changes optional: delegations below the threshold will fail. Otherwise, we would need a very clear warning (on all the platforms that allow delegation).

protocol/20230403-delegator-minimum.md
joshuahannan: thats a good point. I think that plus Alex's points are arguments for just having a minimum to stake at all

joshuahannan:
These are good points. I think that is enough of an argument to change the proposal to just have a minimum to stake anything, not to just receive rewards.

bluesign:
I think 50 is too low, makes like 0.08-0.09 FLOW reward approx per epoch. Also delegated tokens are the biggest problem of  flow economy. Last time I checked account with over 1 FLOW or more non staked balance was very very low percentage. 

Putting something like, 250-500 FLOW minimum and requiring staking amount to be multiple of 250 or 500, can improve economy a lot. ( if I have 400 FLOW, and limit is 250; I can either go buy 100 FLOW or I can spend 150 FLOW ( or at least have unstaked flow, if I see some nice NFT ) )

joshuahannan:
@bluesign The minimum was chosen more for security and performance reasons than economic reasons. Like we say in the FLIP, we still want the minimum to be accessible for regular users so they can feel like they can easily contribute to the network and be rewarded for it. I also would be a little uncomfortable having the delegator minimum be greater than the access node minimum. Can you elaborate a bit more on why you think it should be that high?

bluesign:
@joshuahannan yeah I understand the motivation, but this is also an opportunity to fix delegation I think. 

My though process is a bit like this: ( all in back of napkin style, approx numbers ) 

- we have 2 parts of the system; staking node + staking via delegation for rewards 
- currently staking node is practically not possible ( in the short term at least ) 
- there is no slashing ( in the short term at least ) 
- delegation has 92% of the rewards ( 
- total staked is like 700M , probably like 50-55% of total supply, inflation is like 5%, RoR should be like 9-10%
- if we target 1 FLOW reward per epoch, 52 FLOW rewards per year => 500 FLOW staked with 10% 

Considering worse case scenario, flow price stays same at around 1$,  still nothing can beat FLOW staking RoR, this causes:

- it is risk free, very good RoR 
- no defi can compete with that ( not sure how increment.fi is doing, but I don't think very good ) 
- no new defi will come 

if we put let's say 500 FLOW limit, there will be contracts on FLOW that allow pooling etc; most wallets already have support to allow staking on their nodes. 

>  we still want the minimum to be accessible for regular users so they can feel like they can easily contribute to the network 

I totally agree that this is good mechanic, ( I mean we use this on mobile games as daily rewards, to make user to come daily ( here is weekly ) and get something for free ) but here it feels similar;  contribute to the network part is like saying: "you go deposit your money to bank and feel like you can easily contribute to economy."

redallica:
@joshuahannan I do agree with @bluesign. The amount of delegation is one of the biggest challenges for Flow protocol economics. Any action to reduce the delegated amount and put back the staked Flow in circulation in the network would be a great initiative. With a 9% - 10% APR for delegation, it is a killer for any initiative (DeFi, NFTs, everything) in the network.

Maybe @KshitijChaudhary666 has a view on this one as he is also working on proposals to reduce the inflation rate of the Flow network, which is imo the biggest issue of the Flow protocol economics.

I would recommend taking this into consideration while fixing the Delegator Staking minimum.

KshitijChaudhary666:
Here’s what I think. Having a minimum stake (M) for delegation could have two opposite economic effects, depending on how delegators perceive Flow RoR against other networks or even securities -

CASE-1. Those with stakes <M no longer want to participate, decide to unstake and liquidate. If this is the case, there are two effects - (1A) total stake in the economy reduces, thus the reward share of individual entity increases (ceteris paribus) and we can match today’s rewards with lesser tokens issued each epoch, great. But -> (1B) liquidation could lead to supply spike potentially affecting Flow value if there are no immediate use-cases to absorb the excess tokens. This problem is similar to what we face when we discuss reducing the minimum staking requirements for nodes, which would help reduce inflation but could create a big token surplus problem provided we don't have enough investing/spending opportunities

CASE-2. Those with stakes <M could increase their stake to reach M. If this happens, the total stake in-fact grows, having an adverse effect on inflation

To have this change impact inflation, we want to be sure that first, case-1 overpowers case-2, and second, the impact of liquidation of unstaked tokens (1B) is minimal. Imo, larger the M we set, the higher the probability of case-1 (and 1A) overpowering Case-2, but also creates a larger 1B problem. A mix of the two cases seems most practical, achieved with a non-trivial but fairly small M. While many minority stakers might stop participating, some believers would increase their stake to the new minimum requirement. Maybe a distribution chart of delegated stakes can help understand how much the total stake could reduce, and the max tokens that could go in circulation at various 'M's (M= 50, M=100, M=500, etc.), but don’t think we can build a conclusive judgment here on how inflation would be affected.

That said, I agree that high-yet-risk-free-RoR for delegators is a key problem, and we should aim to solve it via economic incentives instead of minimum caps - for eg an increase in delegation cut (take-rate) that operators receive, along with effective slashing conditions affecting all parties (incl delegators) would ensure the delegators no longer receive risk-free pie. This can also incentivize operators, and either increase their APY with the same inflation, or maintain their APY with a reduced inflation. I'm working on this as one of the recommendations re inflation. Meanwhile, a small minimum delegation requirement would address other non-economic but crucial parameters (security/ performance) like @joshuahannan mentioned.

bluesign:
@KshitijChaudhary666 I would normally not agree, but I just take a look into the numbers, delegators are not a big problem as I imagined. 

Case 1 is totally ignorable, numbers are really weird. Even if we setup 500 FLOW limit, get rid of 78% of the delegators, it is just 0.38% ( 0.0038 ) of total delegated. ( would not make a dent in the price ) 

Case 2 would work perfectly and flow price would for sure increase, but I don't feel good suggesting it with this distribution. 

I think what @joshuahannan suggested is pretty good for now.

joshuahannan:
I created [an issue in the flow-core-contracts repo](https://github.com/onflow/flow-core-contracts/issues/363) and I'm going to get started on it soon so we can have an implementation to show for the FLIP

PR #79, opened 2023-04-03
FLIP79- Service Account Disable Event Emissions Flag

joshuahannan:
This FLIP proposes a Cadence flag that only the service account can use to turn off event emissions while it is enabled.
It will be only used in the staking contract to avoid emitting meaningless `FlowToken` events.

### DIFF
--- a/protocol/20230403-cadence-event-flag.md
+++ b/protocol/20230403-cadence-event-flag.md
@@ -10,7 +10,7 @@ updated: 2023-04-03
 
 ## Objective
 
-Introduce a flag in Cadence that is only allowed to be used by the service account
+Introduce a global function in Cadence that is only allowed to be used by the service account
 that disables event emission while it is enabled.
 
 ## Motivation
@@ -41,7 +41,7 @@ curl -s https://rest-mainnet.onflow.org/v1/transaction_results/84eca4ff612ef7004
 You'll need to multiply the result by two to get the total number of events for each
 delegator above the specified rewards amount.
 
-This FLIP proposes adding a flag to Cadence that is only usable by the service account
+This FLIP proposes adding a function to Cadence that is only usable by the service account
 that turns off event emissions while it is enabled. This would allow the staking
 contract to avoid emitting the meaningless `FlowToken` events.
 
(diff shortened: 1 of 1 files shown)


joshuahannan:
I just want to leave a comment for those who don't think that the service account should have special privileges. I understand that sentiment and I think the jury is still out on if it is needed in this situation, but generally, non-smart contract protocol software already inherently has privileged permissions, so I don't see why we can't have some privileges for the protocol smart contracts as well. We had to design them in an inefficient way partly in order to make the protocol function properly and safely, so I don't think that having privileges to make up for that is much of an issue.

protocol/20230403-cadence-event-flag.md
```diff
@@ -0,0 +1,110 @@
...
+contract to avoid emitting the meaningless `FlowToken` events.
+
+## User Benefit
+
+This change would benefit the network because the epoch transition operations would complete faster
+and the event payload would be much smaller and less cluttered with meaningless events.
+
+## Design Proposal
+
+Propose a pragma for cadence that can be enable by the staking contract every time
```
turbolent, line 55: Could you please describe in more detail what pragma is proposed and how it behaves?

joshuahannan: I was hoping that y'all would help me define that if we decide this is a good idea. I was basically just thinking something that allows us to call a method every time we want to turn event emissions off or on, so something like [this section in the staking contract](https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowIDTableStaking.cdc#L1217-L1224) would instead look something like this:

```
               for delegator in rewardBreakdown.delegatorRewards.keys {
                    let delRecord = nodeRecord.borrowDelegatorRecord(delegator)
                    let delegatorReward = rewardBreakdown.delegatorRewards[delegator]!
                     
                    turnOffEvents()
                    delRecord.tokensRewarded.deposit(from: <-rewardsVault.withdraw(amount: delegatorReward))
                    turnOnEvents()

                    emit DelegatorRewardsPaid(nodeID: rewardBreakdown.nodeID, delegatorID: delegator, amount: delegatorReward)
                }
```

Does that help?

turbolent: Yeah, I would prefer a function that's injected into transactions signed by the service account, compared to a pragma that'd defined for the language.

joshuahannan: okay I added the example to the FLIP and removed the pragma mentions. For now though, I mostly just want to hear from more people about if they think this proposal is a good idea from a philosophical standpoint

joshuahannan:
I'm going to close this for now. I don't think we need it right now

PR #82, opened 2023-04-17
Events emitted from interfaces

bluesign:
draft: events emitted from interfaces

### DIFF
--- a/flips/20230417-events-emitted-from-interfaces.md
+++ b/flips/20230417-events-emitted-from-interfaces.md
@@ -138,6 +138,7 @@ Existing contracts would still emit their old events in addition to these new ev
 The Flow team should be responsible for maintaining this feature and
 incorporating it into standard contracts.
 
+This change will be included in the [FT/NFT v2 standards](https://github.com/onflow/flips/pull/56) along with all the other updates and breaking changes that are part of the proposals.
 ### Best Practices
 
 This introduces a new best practice, allowing interface owners the option to
--- a/flips/20230417-events-emitted-from-interfaces.md
+++ b/flips/20230417-events-emitted-from-interfaces.md
@@ -1,0 +1,161 @@
+# Events emitted from interfaces
+
+| Status        | (Proposed / Rejected / Accepted / Implemented)       |
+:-------------- |:---------------------------------------------------- | |
+**FLIP #**    | [NNN](https://github.com/onflow/flow/pull/NNN) (update when you
+have PR #)| | **Author(s)** | Deniz Mert Edincik (deniz@edincik.com)
+| | **Sponsor**   |                                                      | |
+**Updated**   | 2023-04-17                                           | |
+
+## Objective
+
+The objective of this proposal is to enable interfaces to emit events.
+
+## Motivation
+
+At present, Cadence allows the declaration of events within interfaces, but it
+does not permit the emission of events directly from these interfaces. Instead,
+it relies on the implementation to emit events as needed.  This approach can
+lead to potential issues with contracts, including the emission of incorrect
+events, events with improper parameters, multiple emissions of the same event,
+or the omission of necessary events. 
+
+By viewing interfaces as a set of governing rules for implementation,
+incorporating event restrictions could serve as an effective solution to these
+challenges.
+
+## User Benefit
+
+This will establish a more secure environment for the user, ensuring that the
+emitted event is a direct outcome of an action specified within the interface.
+
+## Design Proposal
+
+For instance, a sample extract from the FungibleToken contract can be currently
+represented as:
+
+
+```cadence 
+
+pub contract FungibleToken {
+   
+    pub event TokenWithdrawal(type: Type, amount: UFix64, from: Address?)
+   
+    pub resource interface Provider{ 
+        pub fun withdraw(amount: UFix64): @{FungibleToken.Vault} 
+        { 
+            post 
+            { 
+                result.balance == amount: "Withdrawal amount 
+                must be the same as the balance of the withdrawn Vault" 
+            } 
+        } 
+    }
+ 
+```
+
+We can enhance our approach by defining a trampoline function, which not only
+eliminates the need for altering the cadence but also offers a more efficient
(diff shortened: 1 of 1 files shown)


flips/20230417-events-emitted-from-interfaces.md
```diff
@@ -0,0 +1,161 @@
...
+challenges.
+
+## User Benefit
+
+This will establish a more secure environment for the user, ensuring that the
+emitted event is a direct outcome of an action specified within the interface.
+
+## Design Proposal
+
+For instance, a sample extract from the FungibleToken contract can be currently
```
joshuahannan, line 34:
```suggestion
For instance, a sample extract from the `FungibleToken` contract can be currently
```

flips/20230417-events-emitted-from-interfaces.md
```diff
@@ -0,0 +1,161 @@
...
+
+pub contract FungibleToken {
+   
+    pub event TokenWithdrawal(type: Type, amount: UFix64, from: Address?)
+    
+    //Temporary trampoline to emit event 
+    access(contract) fun emitTokenWithdrawal(type: Type, amount: UFix64, from: Address?): Bool{ 
+         emit TokenWithdrawal(type:type, amount:amount, from:from) 
+         return true 
+    }
```
joshuahannan, line 74: Does it have to be part of a separate function?

joshuahannan (approved):
I think this is a great idea and I don't see any issues with it! I'm gonna share around to try to get some more feedback

bjartek:
+1

sisyphusSmiling (approved):
Looks great!

flips/20230417-events-emitted-from-interfaces.md
bluesign: this is a temporary workaround ( as we don't have a way to emit function from post conditions ) but also it has advantage that we can emit different events depending to parameters.

turbolent (commented):
Great idea 👍 

Unless I'm missing something, looking at the "Design Proposal" section, it looks like no actual changes to Cadence are proposed, but rather a pattern. If that is the case, then we do not really need a FLIP, but as mentioned, can describe this pattern in the documentation and recommend it to contract authors.

The "Alternatives Considered" section describes an actual change to Cadence – was the intent to propose that change in the FLIP, or is the idea that such a language change is not necessary, given that the pattern can be used?

If you do want to propose the Cadence change, it might make sense to restructure the proposal and move what is in "Alternatives Considered" to the top of "Design Proposal", and then use the current content of that section (the pattern) as a rational for the proposed feature.

flips/20230417-events-emitted-from-interfaces.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Design Proposal
+
+For instance, a sample extract from the `FungibleToken` contract can be currently
+represented as:
+
+
+```cadence 
+
+pub contract FungibleToken {
```
turbolent, line 40: Is this declaration missing an `interface` modifier? Currently, the `FungibleToken` standard is an interface.

turbolent:
Just to clarify my understanding:

Is the idea that an event type which is declared in an interface could be refereed to as a concrete type, and thus instantiated, even directly in the interface? e.g.

```cadence
contract interface SomeContractInterface {

    event SomeEvent()

    fun someDefaultFunction() {
        emit SomeEvent()
    }
}
```

Currently the semantics of a type definition nested in an interface is that it is a nested type requirement, i.e. it is not a concrete type, but just declares that the concrete type conforming to the interface must define such a type.

Is the idea to change the behaviour just for events and make them already concrete in the interface?
What about e.g. other kinds of type declarations, like structs and resources.
It might be a bit inconsistent that event type declarations would be concrete, but other kinds of type declarations nested in interfaces would stay nested type requirements, as they are right now.

Maybe asked more generally, does the proposal require/imply the removal of nested type requirements at the same time?

joshuahannan:
I think the proposal would imply the removal of nested type requirements since this is the last thing we need to change to make it okay to remove them. That would make it much cleaner.

I'm not sure for this feature if declaring the events in the resource interface type would make the most sense, or if it would make sense to just force the events to still be declared in the contract instead. I lean more towards declaring them in the contract so it doesn't seem like they are requirements for the interface.

bluesign:
Sorry I was on vacation just saw the comments.

@turbolent I think this should be adopted in core contracts ( FT, NFT etc ) so maybe FLIP can make sense even it is not a cadence change. I started as a draft actually to decide between alternatives, I don't have any idea how hard would it be to allow post conditions to emit events to be honest. 

> Maybe asked more generally, does the proposal require/imply the removal of nested type requirements at the same time?
Yeah you are totally right, I built my prototype on `contract` instead of `contract interface` which requires additional `contract` if we remove nested type requirements. ( which I totally missed ) When thinking about it, other than making events special case, there doesn't seem to be solution. Having a second contract to declare events makes no sense.

joshuahannan:
I'm changing the v2 token standard proposals to be contracts instead of contract interfaces, so you're right that this technically could be done now with separate functions, but I like the idea of being able to emit the events directly from the pre/post conditions, so it would still be nice to make the change to cadence to allow that.

turbolent:
So just to clarify my understanding, is it correct that the FLIP is basically proposing these two things:
- Allow events to be emitted in conditions directly, without the need for a helper function (most important)
- Allow nesting of events in contract interfaces (once/while at the same time removing nested type requirements

bluesign:
> Allow events to be emitted in conditions directly, without the need for a helper function (most important)

I think this one is final decision ( I will update the FLIP ) 

> Allow nesting of events in contract interfaces (once/while at the same time removing nested type requirements)

I think if we will break things and FT / NFT will be contract from now on. I don't think we need to allow nesting of events in contract interfaces.

bluesign:
@SupunS I would appreciate that

SupunS:
Opened a new PR https://github.com/onflow/flips/pull/111 with the updated proposal.

turbolent:
replaced by #111

PR #83, opened 2023-05-03
FLIP: Improve interface conformance

SupunS:
The FLIP solves https://github.com/onflow/cadence/issues/2471

More discussions can be found in the ^ issue.


dsainati1 (commented):
I don't expect this to be controversial, and I am in favor of this change. However I do feel it would be best to delay this change until the Stable Cadence release, since until we have the guarantees provided by `view` functions, we can't be sure that a precondition might not change the observable behavior of a function.

turbolent:
I'm in favor of the change, but Daniel has a great point. 

It would be great to get some more feedback from the community if there is any demand to deliver this change earlier, or if it is acceptable to delay it until the Stable Cadence release. AFAICS so far we hadn't heard any demands for the change.

SupunS:
Yeah, definitely. This doesn't have to be shipped immediately. As I also mentioned in the FLIP, the usage of this lessened restriction is likely to be more useful with interface inheritance, which is also going to be in Stable Cadence. So it makes it even more sense to add along with that.

joshuahannan (approved):

cadence/2023-05-03-improve-conformance.md
```diff
@@ -0,0 +1,99 @@
+# Improve interface conformance
+
+| Status        | Proposed                                     |
+:-------------- |:---------------------------------------------|
+| **FLIP #**    |                                              |
+| **Author(s)** | Supun Setunga (supun.setunga@dapperlabs.com) |
+| **Sponsor**   | Supun Setunga (supun.setunga@dapperlabs.com) |
+| **Updated**   | 2023-05-03                                   |
```
turbolent, line 8:
```suggestion
---
status: proposed
flip: 83
authors: Supun Setunga (supun.setunga@dapperlabs.com)
sponsor: Supun Setunga (supun.setunga@dapperlabs.com)
updated: 2023-05-03
---

# FLIP 83: Improve interface conformance
```

cadence/2023-05-03-improve-conformance.md
turbolent, line 1: File needs to be renamed to `20230503-improve-conformance.md`

turbolent (approved):
👍 from me

satyamakgec (approved):

turbolent:
@SupunS let's approve and merge 🎉

PR #85, opened 2023-05-05
FLIP for removing restricted types

dsainati1:
This proposes to remove restricted types and replace them with InterfaceSet types, as they are no longer necessary after the proposed changes in https://github.com/onflow/flips/pull/54.


bluesign:
supporting 100%

cadence/2023-05-05-remove-restricted-types.md
```diff
@@ -0,0 +1,64 @@
+---
+status: proposed 
+flip: NNN (do not set)
+authors: Daniel Sainati (daniel.sainati@dapperlabs.com)
+sponsor: Daniel Sainati (daniel.sainati@dapperlabs.com)
+updated: 2023-05-05
+---
+
+# Remove Restricted Types
```
turbolent, line 9: Maybe mention new feature name
```suggestion
# Replace Restricted Types with Interface Set Types
```

cadence/2023-05-05-remove-restricted-types.md
```diff
@@ -0,0 +1,64 @@
...
+This would remove the "restricted type" portion of the restricted type. Users will no longer be able to write types with a `T` in 
+the `T{I}` position of an old restricted type. 
+
+For the new semantics of the interface set type, we can use the existing semantics of the `AnyStruct{X}` and `AnyResource{Y}` restricted types, 
+as these contain no information about the type being restricted, and thus function only as restrictions on a generic type. 
+
+Existing references and capabilities could be migrated by replacing the restricted type with the outer type, i.e. converting `T{I}` to `T`. 
+In combination with the incoming entitlement changes, where the old "restricted" behavior would be recaptured with entitlements, this would
+be able to preserve existing behavior. 
+
```
turbolent, line 46: Maybe clarify that there is no behaviour change, and that the introduction of entitlements will play into here: we probably want to migrate interfaces in restricted references to authorized references with "equivalent" entitlements

cadence/2023-05-05-remove-restricted-types.md
```diff
@@ -0,0 +1,65 @@
+---
+status: proposed 
+flip: NNN (do not set)
+authors: Daniel Sainati (daniel.sainati@dapperlabs.com)
+sponsor: Daniel Sainati (daniel.sainati@dapperlabs.com)
+updated: 2023-05-05
+---
+
+# Replace Restricted Types with Interface Set Types
```
turbolent, line 9:
```suggestion
flip: 85
authors: Daniel Sainati (daniel.sainati@dapperlabs.com)
sponsor: Daniel Sainati (daniel.sainati@dapperlabs.com)
updated: 2023-05-05
---

# FLIP 85: Replace Restricted Types with Interface Set Types
```

cadence/2023-05-05-remove-restricted-types.md
turbolent, line 1: Rename to `20230505-remove-restricted-types.md`

dsainati1:
Keeping the feature definitely isn't a problem per se, but it is now a vestigial part of the language, and given our commitment to supporting the feature set of Stable Cadence, this would really be our last chance to remove it. Given that we're already planning to break every contract anyways, it doesn't seem like that much additional overhead to use this opportunity to remove this while we still can.

turbolent:
I think this change makes sense, and might even avoid confusion and security issues. Older documentation, tutorials, blog posts, code, etc. might still use restricted references for access control and developers might continue to follow these examples. A combination of entitlements (which "replace" restricted references for access control) and restricted type does not make a lot of sense.

@SupunS This change could actually be a solution to the current problem we ran into yesterday: Built-in members are not available on restricted types.

bjartek:
I am not against the feature, but there needs to be a clear migration path. If a contract stores something like

`access(self) let cap: Capability<&{MetadataViews.ResolverCollection}>
`
in a field there needs to be a way to migrate it

dsainati1:
> there needs to be a clear migration path

The migration path is pretty clear imo: the example you give with `Capability<&{MetadataViews.ResolverCollection}>` would not need to be migrated at all: this would simply be left as is. The only part of restricted types that is being removed is the outer `T` in a type of the form `T{I}`. So an existing `Capability<&Vault{Provider}>` reference would become invalid, but can be easily migrated to `Capability<&Vault>`.

bluesign:
> The only part of restricted types that is being removed is the outer T in a type of the form T{I}. So an existing Capability<&Vault{Provider}> reference would become invalid, but can be easily migrated to Capability<&Vault>.

nit: I think correct statement is `{I}` is removed no? ( except AnyResource case, where then just interface )

dsainati1:
Yes, I suppose for the purpose of the migration in this case it would be the `{I}` part of the annotation that is being removed. The point I was trying to make was that the `T` part of that type is being removed from the language, but it is perhaps a little unclear.

turbolent:
It just occurred to me that this is also great from a syntactical perspective.

The removal of `T` in the syntax `T{Us}` would remove the ambiguity for functions in interfaces, e.g. `fun foo(): T {}`, which might mean that the function returns a `T` and has an empty body, or which returns `T{}` and has no body. Currently this is disambiguated through whitespace, which is bad.

Maybe we could If we could further change the syntax and bring it closer to similar features in other languages? It would be great to remove the overloading of the `{` token, as it is also used for dictionaries. For example, when parsing `{T, U}`, until the `,` token occurs it is unclear if it is a restricted type or a dictionary. 

For example, TypeScript, Swift, and Scala have the syntax `T & U & V ...`, for [intersection types](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#intersection-types), [protocol composition](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/protocols/#Protocol-Composition), and [intersection types](https://docs.scala-lang.org/scala3/reference/new-types/intersection-types.html) respectively.

cadence/20230505-remove-restricted-types.md
```diff
@@ -0,0 +1,65 @@
...
+This would remove the "restricted type" portion of the restricted type. Users will no longer be able to write types with a `T` in 
+the `T{I}` position of an old restricted type. 
+
+For the new semantics of the interface set type, we can use the existing semantics of the `AnyStruct{X}` and `AnyResource{Y}` restricted types, 
+as these contain no information about the type being restricted, and thus function only as restrictions on a generic type. 
+
+Existing references and capabilities could be migrated by replacing the restricted type with the outer type, i.e. converting `T{I}` to `T`. 
+In combination with the incoming entitlement changes, where the old "restricted" behavior would be recaptured with entitlements, this would
+be able to preserve existing behavior. In particular, in the case of references, the entitlements present on the interfaces would be granted to the
+reference type.
```
SupunS, line 46: Maybe add a real-world example

cadence/20230505-remove-restricted-types.md
```diff
@@ -0,0 +1,65 @@
...
+
+This will simplify the language by removing a redundant and poorly understood feature. 
+
+## Design Proposal
+
+This would remove the "restricted type" portion of the restricted type. Users will no longer be able to write types with a `T` in 
+the `T{I}` position of an old restricted type. 
+
+For the new semantics of the interface set type, we can use the existing semantics of the `AnyStruct{X}` and `AnyResource{Y}` restricted types, 
+as these contain no information about the type being restricted, and thus function only as restrictions on a generic type.
```
SupunS, line 41: Maybe also mention how upcasting/downcasting works after this change (even if it stays the same way as now)

SupunS (approved):
Big 👍  from me

turbolent (approved):
👍

dsainati1:
This seems like it's reached consensus, so I am updating the FLIP to accepted and merging this PR.

cadence/20230505-remove-restricted-types.md
```diff
@@ -0,0 +1,69 @@
+---
+status: proposed
```
dsainati1, line 2:
```suggestion
status: accepted 
```

turbolent:
Could we maybe still discuss the opportunity for changing the syntax as notes above?

PR #86, opened 2023-05-12
Introducing built-in mutability entitlements

SupunS:
Depends on https://github.com/onflow/flips/pull/89

### DIFF
--- a/cadence/20230519-built-in-mutability-entitlements.md
+++ b/cadence/20230519-built-in-mutability-entitlements.md
@@ -17,7 +17,7 @@ dictionaries and composite types.
 
 A previous version of Cadence ("Secure Cadence") restricted the potential foot-gun of mutating container-typed
 `let` fields/variables via the
-[Cadence mutability restrictions FLIP](https://github.com/onflow/flips/blob/main/cadence/20211129-cadence-mutability-restrictions.md).
+[Cadence mutability restrictions FLIP (#703)](https://github.com/onflow/flips/blob/main/cadence/20211129-cadence-mutability-restrictions.md).
 
 However, there are still ways to mutate such fields by:
 - Directly mutating a nested composite typed field.
@@ -31,6 +31,11 @@ With an aim of moving one step closer to solving these existing problems, anothe
 
 Going another step further, this FLIP proposes introducing a set of built-in entitlements, that can be used along with
 the change of member-access semantics, to better control who can perform mutating operations through the references.
+With that, it is also proposed to remove the restrictions introduced in the previous
+[Cadence mutability restrictions FLIP (#703)](https://github.com/onflow/flips/blob/main/cadence/20211129-cadence-mutability-restrictions.md).
+
+The [Mutability Restrictions Vision](https://github.com/onflow/flips/pull/97) explains how this proposed changes
+contribute to solving the aforementioned problems, and how the final solution looks like, with examples.
 
 ## User Benefit
 
@@ -173,6 +178,39 @@ removableArrayRef[2] = "John"     // Static Error: doesn't have the required ent
 ```
 
 This would behave the same way for dictionaries as well.
+
+### Removing existing restrictions
+
+The changes (mutable annotation in particular) introduced in [FLIP #703](https://github.com/onflow/flips/blob/main/cadence/20211129-cadence-mutability-restrictions.md)
+are very similar to the ones that are suggested in this FLIP.
+The `mutable` annotation introduced there for built-in functions, can be replaced with the proposed entitlements.
+
+The major difference in the two approaches is that, changes in FLIP#703 also apply to 'owned' values,
+whereas the entitlement-based solution (this proposal) only applicable for references.
+
+However, for an owned value, since it is possible to take a reference to members with any desired entitlement,
+having those restrictions as in FLIP#703 doesn't make much sense.
+
+For an example:
+
+```cadence
+pub resource Collection {
+    pub var ownedNFTs: @{UInt64: NonFungibleToken.NFT}
+}
+
+// Consider an 'owned' value.
+var collection: @Collection <- ...
+
+// Mutating from outside of the resource is prohibitted as per existing type-checking rules.
+collection.ownedNFTRef[1234] <- nft
+
+// However, can take a mutable reference and mutate the field via the reference
+var mutableOwnedNFTRef = &collection as auth(Mutable) @{UInt64: NonFungibleToken.NFT}
+mutableOwnedNFTRef[1234] <- nft
+```
+
+Thus, the existing restrictions can be confusing, and would rather be better to keep a consistent behavior by undoing
+those changes, and leaning towards an entitlement-based solution.
 
 ### Putting It All Together
 
--- a/cadence/20230519-built-in-mutability-entitlements.md
+++ b/cadence/20230519-built-in-mutability-entitlements.md
@@ -1,9 +1,9 @@
 ---
-status: proposed
+status: accepted
 flip: 86
 authors: Supun Setunga (supun.setunga@dapperlabs.com)
 sponsor: Supun Setunga (supun.setunga@dapperlabs.com)
-updated: 2023-05-19
+updated: 2023-08-15
 ---
 
 # Introduce Built-in Mutability Entitlements
--- a/cadence/20230519-built-in-mutability-entitlements.md
+++ b/cadence/20230519-built-in-mutability-entitlements.md
@@ -53,28 +53,6 @@ The main change proposed in this FLIP is to introduce a set of built-in entitlem
   - `Removable`
   - `Mutable`
 
-<!--
-Having:
-
-```cadence
-struct Foo {
-    access(Insertable, Removable) fun someFunction()
-}
-
-var fooRef: auth{Insertable, Removable} Foo = ...
-```
-
-is equivalent to:
-
-```cadence
-struct Foo {
-    access(Mutable) fun someFunction()
-}
-
-var fooRef: auth{Insertable, Removable} Foo = ...
-```
--->
-
 Along with that, built-in array/dictionary mutating functions such as `append`, `remove`, etc., would get a
 corresponding entitlement access, instead of the `pub`/`access(all)` access that they currently possess.
 Thus, only a reference (to an array/dictionary) that has the correct entitlement can call such methods.
--- a/cadence/20230519-built-in-mutability-entitlements.md
+++ b/cadence/20230519-built-in-mutability-entitlements.md
@@ -1,0 +1,373 @@
+---
+status: proposed
+flip: 86
+authors: Supun Setunga (supun.setunga@dapperlabs.com)
+sponsor: Supun Setunga (supun.setunga@dapperlabs.com)
+updated: 2023-05-19
+---
+
+# Introduce Built-in Mutability Entitlements
+
+## Objective
+
+The objective of this proposal is to introduce a set of built-in entitlements to control mutable operations on arrays,
+dictionaries and composite types.
+
+## Motivation
+
+A previous version of Cadence ("Secure Cadence") restricted the potential foot-gun of mutating container-typed
+`let` fields/variables via the
+[Cadence mutability restrictions FLIP](https://github.com/onflow/flips/blob/main/cadence/20211129-cadence-mutability-restrictions.md).
+
+However, there are still ways to mutate such fields by:
+- Directly mutating a nested composite typed field.
+- Mutating a field by calling a mutating function on the field.
+- Mutating the field via a reference.
+
+More details on this problem is described in the [FLIP for improving mutability restrictions](https://github.com/onflow/flips/pull/58).
+
+With an aim of moving one step closer to solving these existing problems, another FLIP was proposed previous to this, to
+[change member access semantics](https://github.com/onflow/flips/pull/89) for container types.
+
+Going another step further, this FLIP proposes introducing a set of built-in entitlements, that can be used along with
+the change of member-access semantics, to better control who can perform mutating operations through the references.
+
+## User Benefit
+
+The changes suggested in this proposal eliminate the potential for users to accidentally expose mutable fields through
+references.
+
+## Design Proposal
+
+Before going into the details of the proposal, it is important to note that these proposed changes depend on the
+[change to member-access semantics](https://github.com/onflow/flips/pull/89) and the introduction of
+[Entitlements](https://github.com/onflow/flips/pull/54).
+
+The main change proposed in this FLIP is to introduce a set of built-in entitlements:
+  - `Insertable`
+  - `Removable`
+  - `Mutable`
+
+<!--
+Having:
+
+```cadence
+struct Foo {
+    access(Insertable, Removable) fun someFunction()
+}
+
+var fooRef: auth{Insertable, Removable} Foo = ...
+```
+
+is equivalent to:
+
+```cadence
+struct Foo {
+    access(Mutable) fun someFunction()
+}
+
+var fooRef: auth{Insertable, Removable} Foo = ...
+```
+-->
+
+Along with that, built-in array/dictionary mutating functions such as `append`, `remove`, etc., would get a
+corresponding entitlement access, instead of the `pub`/`access(all)` access that they currently possess.
+Thus, only a reference (to an array/dictionary) that has the correct entitlement can call such methods.
+
+### Arrays
+
+Array functions would now have the below entitlements:
+
+- `access(all) fun contains(_ element: T): Bool`
+- `access(all) fun firstIndex(of: T): Int?`
+- `access(all) fun slice(from: Int, upTo: Int): [T]`
+- `access(all) fun concat(_ array: T): T`
+- `access(Mutable | Insertable) fun append(_ element: T): Void`
+- `access(Mutable | Insertable) fun appendAll(_ element: T): Void`
+- `access(Mutable | Insertable) fun insert(at: Int, _ element: T): Void`
+- `access(Mutable | Removable) fun remove(at: Int): T`
+- `access(Mutable | Removable) fun removeFirst(at: Int): T`
+- `access(Mutable | Removable) fun removeLast(at: Int): T`
+
+Thus, for any arbitrary array:
+
+```cadence
+var arrayRef = &array as &[String]
+arrayRef.contains("John")          // OK
+arrayRef.append("John")            // Static Error: doesn't have the required entitlement
+arrayRef.remove(2)                 // Static Error: doesn't have the required entitlement
+
+var insertableArrayRef = &array as auth(Insertable) &[String]
+insertableArrayRef.contains("John") // OK
+insertableArrayRef.append("John")   // OK
+insertableArrayRef.remove(2)        // Static Error: doesn't have the required entitlement
+
+var removableArrayRef = &array as auth(Removable) &[String]
+removableArrayRef.contains("John")  // OK
+removableArrayRef.append("John")    // Static Error: doesn't have the required entitlement
+removableArrayRef.remove(2)         // OK
+
+var mutableArrayRef = &array as auth(Mutable) &[String]
+mutableArrayRef.contains("John")   // OK
+mutableArrayRef.append("John")     // OK
+mutableArrayRef.remove(2)          // OK
+```
+
+### Dictionaries
+
+Dictionary functions would now have the below entitlements:
+
+- `access(all) fun containsKey(key: K): Bool`
+- `access(all) fun forEachKey(_ function: ((K): Bool)): Void`
+- `access(Mutable | Insertable) fun insert(key: K, _ value: V): V?`
+- `access(Mutable | Removable) fun remove(key: K): V?`
+
+Thus, for any arbitrary dictionary:
+
+```cadence
+var dictionaryRef = &dictionary as &{String:AnyStruct}
+dictionaryRef.containsKey("John")               // OK
+dictionaryRef.insert("John", "Doe")             // Static Error: doesn't have the required entitlement
+dictionaryRef.remove("John")                    // Static Error: doesn't have the required entitlement
+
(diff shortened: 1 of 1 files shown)


cadence/2023-05-10-references-on-member-access.md
```diff
@@ -0,0 +1,322 @@
...
+thinking they are constants.
+
+## Design Proposal
+
+Before going into the details of the proposal, it is important to note that these proposed changes depend on
+[Entitlements](https://github.com/onflow/flips/pull/54).
+
+This proposal consist of two main changes:
+ - Member-access to return references.
+ - Introducing a built-in `Mutable` entitlement.
```
dete, line 42: By introducing this built-in entitlement, are setting a precedent that entitlements should be defined in title case (like types), and that they should be in adjective form ("Mutable"). I don't have problem with either of these choices, but I just wanted to draw attention to the fact that these decisions are being made here (whether we mean to or not!), and we might as well make them intentionally. If anyone has suggestions for the best way to style the names of entitlements otherwise, now is the best time to debate them!

Some ideas (I don't endorse any of these, just brainstorming):
- maybe entitlements are lower case like functions? `mutable`
- maybe they start with a special character (like underscore) `_mutable`
- verb form? `Mutating`
- noun form? `Mutator`
- adverb? `Mutably`

dete: On language forms, some other examples are worth thinking through:

| Entitlement | Noun | Verb | Adjective | Adverb |
| --- | --- | --- | --- | --- |
| Modifying a Collection | Mutator | Mutating | Mutable | Mutably |
| Vault Withdrawing | Provider | Providing | Providable | Providably |
| Kitty Renaming | Namer | Naming | Namable | Namably |
| Kitty Breeding | Breeder | Breeding | Breedable | Breedably |
| NFT Minting | Minter | Minting | Mintable | Mintably |

Going through these examples, I think that the noun and verb forms are the cleanest. (The adverb form has to be seen as a joke at this point!) I kind of like the noun form since that does differentiate it from functions, but could mean it's conflated with resource and struct types ("objects"). Without looking at the examples, the adjective form makes the most sense (it's a modifier of the reference).

SupunS: Great point! I don't think we ever had a conversation about the convention so far.

In terms of styling, I personally think people will "use" entitlements more frequently than they define them. So in my view, treating them as similar to access-controllers makes more sense, and hence should make them look a lot like access controllers.
i.e: a lower-case naming. e.g: `access(all)`, `access(self)` and similarly `access(mutable)`

For the language form (maybe I'm not the best person to argue about this, not my first language :) ), In this particular case, I felt it's natural to call a "mutable reference" for it implies the value is mutable via the reference.
But I agree, considering the other examples, noun sounds more suitable.

If we go with noun-form, one reason why we should maybe go for a lower-case naming is to avoid any confusion with naming: for e.g: Now there is a `Provider` interface, and we would need an entitlement to expose provider functionality, and having that also named `Provider` would be really confusing.

dsainati1: On the other hand, lower case in Cadence (and most other languages) are typically used for values, which entitlements are not. IMO they behave more like types and should be cased accordingly; this way its clear to users they can't use them like values. 

As for the part of speech, I like the adjectives. Nouns clearly denote objects and other values, while verbs denote functions. Entitlements "describe" the object that you have (in this case a reference) and the interactions that you can perform on that reference, so it seems to me like adjectives would be easiest to talk about; a `Mutable` reference to a `Vault` or a `Breedable` reference to a `Kitty` cleanly describe exactly what you are allowed to do with that reference.

dsainati1:
One other suggestion I had to improve this proposal would be to "break" the `Mutable` entitlement up a little bit more; instead of having one `Mutable` entitlement that allows any and all mutating operations on arrays and dictionaries, IMO we can have `Insertable` and `Removable` entitlements for these types, along with the `Mutable` entitlement. In this model the `Mutable` entitlement would allow access to any mutating function, but the `Insertable` would only allow inserting new values (while `Removable` would only allow removing). This doesn't need any support for "entitlement inheritance", as it can already be modeled today with entitlement sets; using type definitions for arrays like:

```
struct Array {
    access(all) let length: Int
    access(all)fun concat(_ array: T): T
    access(all) fun contains(_ element: T): Bool
    // ... other pub functions here 
    access(Mutable | Insertable) fun append(_ element: T): Void
    access(Mutable | Insertable) fun appendAll(_ array: T): Void
    access(Mutable | Insertable) fun insert(at: Int, _ element: T): Void
    
    access(Mutable | Removable) fun remove(at: Int): T
    access(Mutable | Removable) fun removeFirst(): T
    access(Mutable | Removable) fun removeLast(): T
 }
 ```
 
This way it will be possible to give someone only add/remove access to an array, while still also making it simple to give them full edit access as well.

File: cadence/2023-05-10-references-on-member-access.md [hunk 2]
```diff
@@ -0,0 +1,322 @@
...
+
+// However, `collection.id` would resturn the value (same as existing semantics),
+// since it is a prmitive type.
+var id: String = collection.id
+```
+
+It is also important to note that, the returned reference has no entitlements assigned to them.
+i.e: they are **non-**`auth` references.
+
+### Built-in `Mutable` Entitlement
```
turbolent, line 76: It might be better to propose this addition in a separate FLIP, and focus on just the change to member access semantics in this FLIP.

Such a follow-up FLIP could then propose to _replace_ the existing feature, [FLIP 703: Cadence Mutability Restrictions](https://github.com/onflow/flips/blob/main/flips/20211129-cadence-mutability-restrictions.md), by replacing the notion of "mutability" for functions with entitlements for them.

cadence/2023-05-10-references-on-member-access.md
```diff
@@ -0,0 +1,322 @@
...
+| **Updated**   | 2023-05-10                                   |
+
+## Objective
+
+This proposal suggest to return a reference when accessing a member of a container (field of a composite value, 
+key of a map, index of an array), if the accessed member is also a container type (i.e: non-primitive type).
+
+## Motivation
+
+A previous version of cadence (secure cadence release) restricted the potential foot-gun of mutating container-typed
```
turbolent, line 17:
```suggestion
A previous version of Cadence ("Secure Cadence") restricted the potential foot-gun of mutating container-typed
```

cadence/2023-05-10-references-on-member-access.md
```diff
@@ -0,0 +1,322 @@
...
+`let` fields/variables via the
+[Cadence mutability restrictions FLIP](https://github.com/onflow/flips/blob/main/flips/20211129-cadence-mutability-restrictions.md).
```
turbolent, line 19: Maybe add two point out that this solution was only a partial solution and only covered arrays and dictionaries, built-in types, and was not a general solution for user-defined types.

cadence/2023-05-10-references-on-member-access.md
```diff
@@ -0,0 +1,322 @@
...
+- Mutating the field via a reference.
+
+More details on this problem is described in the [FLIP for improving mutability restrictions](https://github.com/onflow/flips/pull/58).
+
+This proposal is aimed at solving this problem by returning references for member-access, and using 
+[Entitlements](https://github.com/onflow/flips/pull/54) to control who can perform mutating operations via the reference.
+
+## User Benefit
+
+Suggested changes of this proposal eliminate the potential of users accidentally exposing mutable fields,
```
turbolent, line 32:
```suggestion
The suggested changes of this proposal eliminate the potential for users to accidentally expose mutable fields,
```

cadence/2023-05-10-references-on-member-access.md
```diff
@@ -0,0 +1,322 @@
...
+thinking they are constants.
+
```
turbolent, line 34: Another benefit of this proposal is that it would allow replacing [FLIP 703: Cadence Mutability Restrictions](https://github.com/onflow/flips/blob/main/flips/20211129-cadence-mutability-restrictions.md), the special casing for built-in types (arrays and dictionaries) could be generalized to user-defined types, by introducing entitlements for built-in types (e.g. `Mutable` entitlement; but as mentioned below, maybe better in a separate FLIP)

cadence/2023-05-10-references-on-member-access.md
```diff
@@ -0,0 +1,322 @@
...
+Before going into the details of the proposal, it is important to note that these proposed changes depend on
+[Entitlements](https://github.com/onflow/flips/pull/54).
+
+This proposal consist of two main changes:
+ - Member-access to return references.
+ - Introducing a built-in `Mutable` entitlement.
+
+### Member-Access To Return References
+
+The main change proposed in this FLIP is that, field-access (i.e: `foo.bar`) and index-access (`foo[bar]`) would now
```
turbolent, line 46: It might be good to be more specific about the behaviour, i.e. clarify that member access change depends on two things:
  - What the type of the accessed value (base) is (e.g. composite, dictionary, array, optional), and
  - What the type of the element is (e.g. primitive (what exactly?) vs non-primitive (composite, 
dictionary, array, optional?))

cadence/2023-05-10-references-on-member-access.md
```diff
@@ -0,0 +1,322 @@
...
+// `ownedNFTsRef` is a dictionary reference with auth{Mutable} entitlement.
+
+var len = ownedNFTsRef.length        // OK
+
+ownedNFTsRef["someNFT"] <- someNft   // OK
+```
+
+### Drawbacks
+
+#### Mutating functions without `Mutable` entitlements.
```
turbolent, line 174: As discussed, it might be good to point out that inner mutations are not always bad, e.g. by discussing the Vault example, where deposits mutations are "safe", but with withdrawals are "unsafe".

Compared to previous proposals which were concerned with "physical mutation" (failing to find a good term for this concept), i.e. the plain fact if some value is mutated, e.g. by introducing a "mutating" modifier (cf. Swift), this proposal actually addresses the shortcoming of those proposals: Such a notion is probably too limiting, as typically,  programs actually have a notion of "logical mutation" (same here, making up this term for a lack of better words) as demonstrated by the Vault example. As a developer, I want to decide what access is granted to mutating functions.

cadence/2023-05-10-references-on-member-access.md
```diff
@@ -0,0 +1,322 @@
...
+var ownedNFTsRef: &{UInt64: NonFungibleToken.NFT} = collection.ownedNFTs
+
+// However, `collection.id` would resturn the value (same as existing semantics),
+// since it is a prmitive type.
+var id: String = collection.id
+```
+
+It is also important to note that, the returned reference has no entitlements assigned to them.
+i.e: they are **non-**`auth` references.
+
```
turbolent, line 75: It might be good to emphasize that this proposal only applies to access on _references_, and no changes are proposed to access on owned values.

turbolent (commented):
Great start! 👌

cadence/2023-05-10-references-on-member-access.md
dete: Well, aren't we just having a fine ol' bikeshedding moment... 😆 

Let me put my vote in for "noun" and "lowercase". (Sadly, I'm not joking.)

Why noun? Because it describes who you want to have the entitlement. We use nouns for the built-in "entitlements" `access(all)` and `access(self)`, so why not `access(mutator)`, `access(withdrawer)`, and `access(breeder)`?

In fact, I would argue that now that interfaces are all about polymorphism, the interfaces that should change to verb form. `Withdrawing` interface. `Depositing` interface. `Breeding` interface.

Why lower case? Because they _aren't_ types, and we don't have a third letter case. 😅 (Yeah, that's weak!) If you squint, I'd say they feel like values, not types or variables. That implies they are closest to enumerations. Unfortunately, Java, Python, and C seem encourage all-upper, Go and Rust like title case, while Swift and Pascal prefer all-lower for enums... 😈

turbolent: It is probably bike shedding, but still leaving my 2 cents here: 

A naming scheme for entitlements based on nouns, and especially one which describes roles, seems counter to the capability-based security scheme we are trying to encourage: it looks like access is granted to a role, rather than a granted capability. For example, `withdrawer` looks like access is granted to someone who is a withdrawer, but really it should be to something which has the capability to withdraw, i.e. the reference. I feel like verbs work better, but can live with nouns.

For naming in general, and e.g. for interfaces, we've been following [Swift's API design guidelines](https://www.swift.org/documentation/api-design-guidelines/). Its naming suggestions for protocols:

> - Protocols that describe what something is should read as nouns (e.g. Collection).
> - Protocols that describe a capability should be named using the suffixes able, ible, or ing (e.g. Equatable, ProgressReporting).

cadence/2023-05-10-references-on-member-access.md
SupunS: Open a separate pr https://github.com/onflow/flips/pull/89 to propose the member-access semantics change, and re-purpose this PR to propose the introduction of built-in entitlements.

cadence/2023-05-10-references-on-member-access.md
SupunS: This is added in the updated proposal

cadence/2023-05-10-references-on-member-access.md
SupunS: Included in the other proposal: https://github.com/onflow/flips/pull/89

cadence/2023-05-10-references-on-member-access.md
SupunS: I don't think we can completely replace [FLIP 703](https://github.com/onflow/flips/blob/main/flips/20211129-cadence-mutability-restrictions.md), since the changes added there apply to 'owned' values, where the entitlement-based solution (this proposal) only applicable for references.

turbolent: Oh interesting, I didn't realize / forgot it also applied to owned values.

Maybe we can make this FLIP propose to explicitly undo the changes that were introduced in FLIP 703?

SupunS: Added 19f16dc4ff18f3fe098c940af2e4089d2d408b68

turbolent:
👍 for the idea

We still might want to find consensus on the naming scheme. As Dete pointed out, it'll set a precedent, so we might want to discuss the naming scheme separately, as it also is a question/blocker for e.g. #92

cadence/2023-05-10-references-on-member-access.md
cybercent: I agree with @turbolent that using a noun might be confusing: `access(owner)`

bjartek:
I think it is a good idea to discuss naming scheme in more general and just have a documented way on how it is done. Naming things is hard.

cadence/2023-05-10-references-on-member-access.md
bjartek: I prefer lowercase and verbs, but dont have an issue with nouns. Allthough the crash with things like owner might be an issue.

Btw are owner/account/contract going to be entitlements now, or just like they are now?

dsainati1: `account` and `contract` are not entitlements, they are special cased like `all` and `self` that mean specific things when used as `access` modifiers. There is no built in `access(owner)`.

turbolent: @onflow/flow-smart-contracts what do you think about this?

dete: > Protocols that describe a capability should be named using the suffixes able, ible, or ing (e.g. Equatable, ProgressReporting).

This feels like as decent a tie-breaker as anything else! However, they do lump together verbs and adjectives. Should it be "mutating" or "mutable"?

Let me suggest "mutating". Think about using it in a sentence:

"I can modify this object because I have a _mutating reference_."

"I can modify this object because I have a _mutable reference_."

We are _used_ to the second structure, but if you actually think about what the words mean, "mutable reference" means "a reference that can be mutated", which isn't what we have at all. The reference itself is not being mutated! A "mutating reference" means "a reference that mutates". Which is exactly right. (I'll admit that "a reference that mutates" is somewhat ambiguous since it could mean "a reference for mutating" or "a reference that gets mutated", but at least _one_ of those meanings is correct! 😉 )

I think Bastian should flip a coin to decide between Title Case and lowercase. 😈

sisyphusSmiling: Looks like consensus is converging, but I'll add my $.02 in for Swift convention + capitalization.

Adjective/verb makes the most sense to my ears in relation to the reference in question. And capitalization creates distinction from values & types, further distinguished by noun special access modifiers - e.g. self, all, account, contract. 

```cadence
access(MUTATABLE) fun updateName(name: String)
// compared to...
access(Mutatable) fun updateName(name: String)
// and finally
access(mutatable) fun updateName(name: String)
```

That said, lowercase definitely reads better, so I could live with that as well.

dsainati1: You don't mention Title Case in your comment, does that mean it's your least preferred option?

sisyphusSmiling: Yes, title case would be least preferred out of the three IMO

turbolent: Sorry if there was some confusion: The reference to the Swift style guide was in response to Dete's note on naming interfaces, I didn't mean to suggest this for entitlements.

For the verb form, I don't quite get why only the `-ing` form is mentioned (e.g. mutating), and I find the infinite form (e.g. mutate) simpler, For example, a `auth(Mutate) &r` authorizes me to mutate `r`.

When working through the [Account type FLIP](https://github.com/onflow/flips/pull/92), which introduces many fine-grained and coarse-grained entitlements, it felt very natural to use:
- nouns for coarse-grained entitlements, they describe what is granted access to
- verbs for fine-grained entitlements, they describe what operation can be performed

For example, the fine-grained `AddKey` entitlement is added for allowing to add a key,
and the coarse-grained `Keys` entitlement is added to allow any key management operation.

The casing does not matter to me much, but like @dsainati1 pointed out, it might make sense to distinguish them from built-in non-entitlement access modifiers (`all`, `self`, etc.).
I'd prefer title-case over capitalization – that looks like SCREAMING and is a bit too much "in your face", haha.

dsainati1: I don't really feel strongly about the part of speech to be honest; I'd even be fine simply not adding any explicit guidelines here and encouraging developers to use whatever names they want for entitlements. 

However, I do feel very strongly that lowercase is not acceptable for entitlement names because as Bastian mentioned it looks too similar to the built-in access modifiers `all`, `self`, `account` and `contract`, which are explicitly NOT entitlements and cannot be used in conjunction with entitlements. I don't want to confuse anyone into thinking they can use `contract` in an `auth` modifier or in the output of a mapping, so best not even to invite the comparison and just case entitlement names differently than these builtins.

turbolent: Maybe we can simplify the naming further and just use "operation nouns" for fine-grained entitlements.
Basically, the fine-grained entitlement is named  "what operation" can be performed.
For example, `Insertion` instead of `Insert`, and `KeyAddition` instead of `AddKey`.

turbolent: @SupunS @dsainati1 Thoughts?

SupunS: Good point on the verb forms, Bastian. I actually like the 'root form' of the verb more.

Just went through the account type change FLIP and the entitlements that are proposed there. The following (what you previously suggested) actually makes a lot of sense.

> - nouns for coarse-grained entitlements, they describe what is granted access to
> - verbs for fine-grained entitlements, they describe what operation can be performed

Again, we are only proposing a 'guideline', and a single language form does not always fit all the use-case. So for simplicity, I would vote in for the 'root form' of verbs, and nouns, and propose the above two bullet points as the recommendation. Then leave it up to the developers to choose one as appropriate.

cadence/20230519-built-in-mutability-entitlements.md
```diff
@@ -0,0 +1,411 @@
...
+is equivalent to:
+
+```cadence
+struct Foo {
+    access(Mutable) fun someFunction()
+}
+
+var fooRef: auth{Insertable, Removable} Foo = ...
+```
+-->
```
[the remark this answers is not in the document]
SupunS: Removed. I initially added it, but then realized it doesn't make much sense in other contexts. e.g: even though for array/dictionaries `auth{Insertable, Removable}` is equivalent to `auth{Mutable}`, anyone can use these entitlements as they desired, such that the above equivalency may not hold. so Removed it.

turbolent:
@SupunS given the alignment on naming in https://github.com/onflow/flips/pull/86#discussion_r1254611245, maybe the proposed entitlements should be renamed to `Insert`, `Remove`, and `Mutate`?

Also, is setting with indexing going to require `Insert | Mutate`?

SupunS:
> given the alignment on naming in https://github.com/onflow/flips/pull/86#discussion_r1254611245, maybe the proposed entitlements should be renamed to Insert, Remove, and Mutate?

Yea, I was waiting till we reach a conclusion to update the naming. Will update once we have a consensus (looks like we are almost there).

> Also, is setting with indexing going to require Insert | Mutate?

Yes, this is explained in [assigment](https://github.com/onflow/flips/blob/759d9f9b98d6cc4595ad13ea83f834a7fd496bb5/cadence/20230519-built-in-mutability-entitlements.md#assignment) section

cadence/2023-05-10-references-on-member-access.md
SupunS: Updated to the root form: 778d2d5a7ec79932016a571642621df6651fd5e5

turbolent (approved):
👍

turbolent:
@SupunS Approved in the Cadence Language Design Meeting, let's merge this FLIP and the implementation 👏 🎉

PR #89, opened 2023-05-17
Change member access semantics

SupunS:


cadence/20230517-member-access-semnatics.md
```diff
@@ -0,0 +1,185 @@
...
+
+```cadence
+pub resource MasterCollection {
+    pub var kittyCollection: @Collection
+    pub var topshotCollection: @Collection
+}
+
+
+pub resource Collection {
+    pub var ownedNFTs: @{UInt64: NonFungibleToken.NFT}
```
dsainati1, line 114: One thing I think this should also propose: it should be possible to declare container typed fields with entitlement mapped access. Currently it is only possible to use mapped access if the member is either reference-typed (pretty much only useful in the case of the `Account` object), or an accessor function. With this change to have container accesses return references, it would be useful to be able to have these container-typed fields also be able to be declared with entitlement mappings that apply when the field is accessed on a reference.

dsainati1 (commented):
Nice, I am in favor of this change because it interacts well with the new entitlements changes and also has potential for helping out the mutability problem.

cadence/20230517-member-access-semnatics.md
```diff
@@ -0,0 +1,185 @@
...
+## Objective
+
+This proposal suggest to return a reference when accessing a member of a container (field of a composite value,
+key of a map, index of an array), if the parent value is also a reference and the accessed member is a container type.
+
+## Motivation   
+
+A previous version of Cadence ("Secure Cadence") restricted the potential foot-gun of mutating container-typed
+`let` fields/variables via the
+[Cadence mutability restrictions FLIP](https://github.com/onflow/flips/blob/main/flips/20211129-cadence-mutability-restrictions.md).
```
j1010001, line 20: The link is throwing 404.

```suggestion
[Cadence mutability restrictions FLIP](https://github.com/onflow/flips/blob/main/cadence/20211129-cadence-mutability-restrictions.md).
```

cadence/20230517-member-access-semnatics.md
SupunS: I actually thought that is already supported. 💯 agree, we should also support that to make this FLIP to be really useful. 

Probably should propose that as a separate FLIP though, to keep things simple. e.g. This FLIP is currently not dependent on the entitlements FLIP, and would want to keep it that way.

cadence/20230517-member-access-semnatics.md
```diff
@@ -0,0 +1,185 @@
+---
+status: proposed
+flip: 89
+authors: Supun Setunga (supun.setunga@dapperlabs.com)
+sponsor: Supun Setunga (supun.setunga@dapperlabs.com)
+updated: 2023-05-17
+---
+
+# Change Member Access Semantics
```
turbolent, line 9:
```suggestion
# FLIP 89: Change Member Access Semantics
```

cadence/20230517-member-access-semnatics.md
```diff
@@ -0,0 +1,185 @@
...
+
+Assume `collection` is of type `Collection`. i.e. the code owns the value.
+Then, there is no change to the current behavior.
+
+```cadence
+var collection: Collection = ...
+
+// `collection.ownedNFTs` would return the concrete value, which is of type `{UInt64: NFT}`.
+// This is same as existing semantics.
+var ownedNFTs: {UInt64: NFT} = collection.ownedNFTs
```
turbolent, line 76: This example is a bit unfortunate / potentially confusing: It attempts is to illustrate owned access in general, for structs and resources alike, but does so with a struct-type which is commonly implemented as a resource.

It might be good to avoid the confusion by:
- Adding an example for resources, maybe even rework the existing collection/owned NFTs example to a resource ( 
- Reworking the example for structs, avoiding the use of the collection/owned NFTs use-case

SupunS: I intentionally used a struct here to avoid the confusion between `@` and `&` in the reference vs owned scenario. But in the process, it may have created a different form of confusion. Will update the example 👍

SupunS: Updated the example to use resources [459b3b1](https://github.com/onflow/flips/pull/89/commits/459b3b185fdab3fa3b635e5ac6ebf1fdf2acf79d)

turbolent: Nice! Could you please also add an example for structs (which is ideally not based on the collection/owned NFT example to avoid confusion)?

SupunS: Added with 179c6b180211d9d8f7049cd44da2fce0641f796b

cadence/20230517-member-access-semnatics.md
SupunS: Proposed the change in https://github.com/onflow/flips/pull/95

cadence/20230517-member-access-semnatics.md
```diff
@@ -0,0 +1,243 @@
...
+user-defined types.
+
+There are still ways to mutate such fields by:
+- Directly mutating a nested composite typed field.
+- Mutating a field by calling a mutating function on the field.
+- Mutating the field via a reference.
+
+More details on this problem is described in the [FLIP for improving mutability restrictions](https://github.com/onflow/flips/pull/58).
+
+This proposal is aimed at moving one step closer to solving this problem by returning references for member-access,
```
bjartek, line 31: A strong misson statement here, and a good reason to do it. love it.

bjartek (approved):

turbolent:
@SupunS there has not been any additional feedback for multiple weeks, let's approve and merge 🎉

cadence/20230517-member-access-semnatics.md
```diff
@@ -0,0 +1,243 @@
+---
+status: proposed
+flip: 89
+authors: Supun Setunga (supun.setunga@dapperlabs.com)
+sponsor: Supun Setunga (supun.setunga@dapperlabs.com)
+updated: 2023-05-17
```
turbolent, line 6:
```suggestion
status: accepted
flip: 89
authors: Supun Setunga (supun.setunga@dapperlabs.com)
sponsor: Supun Setunga (supun.setunga@dapperlabs.com)
updated: 2023-07-05
```

PR #94, opened 2023-05-31
Improve entitlement mappings

SupunS:


cadence/20211129-entitlement-improvements.md
```diff
@@ -0,0 +1,169 @@
...
+  for composite fields.
+- Introducing an "Identity mapping", which is a shorthand syntax for mapping a given entitlement to the same entitlement.
+
+## Motivation
+
+### Motivation I: Cannot use entitlement mappings for fields
+
+With the [FLIP for changing member access semantics](https://github.com/onflow/flips/pull/89),
+fields of a composite value would return a non-auth reference when accessed via a reference to the composite value.
+However, currently, entitlement mappings can only be used for functions, but not for fields.
```
dsainati1, line 24: They can be used for reference-typed fields only (which is not very useful outside of the `Account` use case admittedly). i.e.

```
access(M) let refField: auth(M) &T
```

would be allowed

cadence/20211129-entitlement-improvements.md
```diff
@@ -0,0 +1,169 @@
...
+
+The proposal is to allow entitlement mappings on composite fields.
+For example, below code would now be valid.
+
+```cadence
+entitlement mapping M {
+    Mutable -> Insertable
+}
+
+pub struct Collection {
```
dsainati1, line 62: Probably this should be a resource, as should the `NFT` dictionary within.

cadence/20211129-entitlement-improvements.md
```diff
@@ -0,0 +1,169 @@
...
+    access(M) var ownedNFTs: {UInt64: NFT}
+}
+```
+
+Assume an `auth{Mutable}` reference to the collection `mutableCollectionRef` is available.
```
dsainati1, line 67: We should also include an explicit explanation of what happens when `ownedNFTs` is accessed on an owned `Collection` value, i.e. something of type `@Collection`. In that case it should behave as normal (i.e. give you the `{UInt64: NFT}` object, but it's worth pointing out in the FLIP.

cadence/20211129-entitlement-improvements.md
```diff
@@ -0,0 +1,169 @@
...
+The proposed syntax is:
+
+```cadence
+entitlement mapping M {
+    include Identity
+}
+```
+
+Here, `Identity` is a built-in entitlement mapping which maps any arbitrary entitlement to itself.
+The `include` keyword includes all entries from a given mapping (`Identity` in this case).
```
dsainati1, line 106: Technically the `Identity` mapping and the `include` feature are separate features (albeit ones designed to be used in tandem). IMO we should separate out these two features into separate sections; this explains the `Identity` mapping well, but it would be good to have a section dedicated to just the `include` feature, maybe also showing off how it can be used with other mappings as well, not just `Identity`.

SupunS: Good idea. Added this and the other improvements in https://github.com/onflow/flips/pull/94/commits/88a07698d14fc363728c3ab5114961ae356b15d6

dsainati1 (approved):

bjartek:
+1 i really like the usability improvements this gives!

turbolent:
@SupunS let's approve and merge 🎉

SupunS:
Thanks everyone for the suggestions and feedback!

This FLIP has been open for a while, and it seems we are all happy with the proposed improvement.
Thus, I'm going to consider this FLIP as accepted 🎉 🎉 🎉

You can follow how the implementation is progressing here ➡️ https://github.com/onflow/cadence/issues/2536

PR #96, opened 2023-06-02
Range Types

darkdrag00nv2:
Work towards https://github.com/onflow/developer-grants/issues/179


cadence/20230602-cadence-range-progression.md
```diff
@@ -0,0 +1,179 @@
...
+Presently, to loop over a start & end integer value, users have to either create an `Array` or use an imperative style `while` loop. A concise syntax for e.g. `1 .. 10` would enhance the readability of the language.
+
+## User Benefit
+
+Users will benefit from the concise syntax leading to a cleaner & readable code.
+
+## Design Proposal
+
+This proposal suggest adding two new types to Cadence:
+1. `Range<T: Integer`
```
dsainati1, line 29:
```suggestion
1. `Range<T: Integer>`
```

cadence/20230602-cadence-range-progression.md
```diff
@@ -0,0 +1,179 @@
...
+It can be instantiated using the `..` & `downTo` binary operators & can be used in a `for-in` loop as follows:
+
+```cadence
+// i will be start, start + 1, .. endInclusive
+for i in start .. endInclusive {
+    // do something with i
+}
+
+// i will be start, start - 1, .. endInclusive
+for i in start downTo endInclusive {
```
dsainati1, line 52: I personally am not a huge fan of the `downTo` operator, to me it looks a bit too much like a variable name. In Kotlin (where I'm guessing this idea came from?) this is a function on types, rather than a built-in language operator. This loop declaration, for example, is quite hard to parse at a glance because it's all identifiers.

cadence/20230602-cadence-range-progression.md
```diff
@@ -0,0 +1,179 @@
...
+let rangeValue = 11 .. 45
+
+let len1 = rangeValue.length // 35
+let isPresent1 = rangeValue.contains(24) // True
+
+for i in rangeValue {
+    // logic using i
+}
+
+let rangeValueBackwards = 132 .. 33
```
dsainati1, line 78: Did you mean `downTo` here? IMO It's pretty clear in this case that this is a decreasing range, so it doesn't seem like a separate operator for this case is all that necessary.

cadence/20230602-cadence-range-progression.md
```diff
@@ -0,0 +1,179 @@
...
+It can be instantiated with the usage of `..` & `downTo` along with a `step` operator & can be used in a `for-in` loop as follows:
+
+```cadence
+// i will be start, start + 2, step + 4, .. endInclusive
+for i in start .. endInclusive step 2 {
+    // do something with i
+}
+
+// i will be start, start - 3, start - 6, .. endInclusive
+for i in start downTo endInclusive step 3 {
```
dsainati1, line 107: I find this really really hard to read.

cadence/20230602-cadence-range-progression.md
```diff
@@ -0,0 +1,179 @@
...
+let progressionValue = 11 .. 45 step 2
+
+let len1 = progressionValue.length // 18
+let isPresent1 = progressionValue.contains(24) // False
+
+for i in progressionValue {
+    // logic using i
+}
+
+let progressionValueBackwards = 132 .. 33 step 2
```
dsainati1, line 134: As an example of the alternative syntax I'm proposing above, this might look like 

```
let progressionValueBackwards = 132, 129 .. 33
```

cadence/20230602-cadence-range-progression.md
```diff
@@ -0,0 +1,179 @@
...
+
+let rangeValueBackwards = 132 .. 33
+
+let len2 = rangeValueBackwards.length // 100
+let isPresent2 = rangeValueBackwards.contains(1) // False
+```
+
+### Progression
+
+A `Progression` type will essentially be a `Range` with an additional `step` which determines the step of the progression. In other words, `Range` will be a `Progression` with `step` set to 1 or -1.
```
dsainati1, line 86: Yea you point out exactly the same thing I mentioned in overall comment right here. These can definitely be combined into one type for simplicity.

cadence/20230602-cadence-range-progression.md
```diff
@@ -0,0 +1,179 @@
...
+
+Example usage of a `Progression` type:
+
+```cadence
+let progressionValue = 11 .. 45 step 2
+
+let len1 = progressionValue.length // 18
+let isPresent1 = progressionValue.contains(24) // False
+
+for i in progressionValue {
```
dsainati1, line 130: It's worth making explicit here what happens when looping over a range where the step overshoots the end of the range. E.g. if I had a range from `10` to `24` with a `step` of `5`, is the last value that `i` attains in the loop body `20` or `25`?

dsainati1 (commented):
Nice! I like the `Range` and `Progression` types because they are nicely extensible; we could later allow people to iterate over any kind of comparable type (strings, addresses, etc) if we lift the restriction that the type argument here be an `Integer`. A couple of suggestions though:

I don't think `Range` and `Progression` actually need to be different types, since a `Range` is just a special case of the more general `Progression` type where the `step` is a specific default value (in this case `1` or `-1`.  I suggest that we combine these into one `Range` type that can be given a `step` if the developer wishes, but otherwise uses these defaults. 

Additionally, I think we don't need the specific `downTo` operator; all this is really doing is making the `step` negative. 

What if instead the range was always defined with `..`, and the sign of the `step` is automatically inferred from whether the start or end value is greater. We could also imagine a syntax like `let range = x, y .. z` where the `start` of this range is `x`, the end is `z`, and the step is `y-x`. Rephrased, developers could define the sequence by giving it a start and end, and if they want the step to be something other than `+/- 1`, they can define the second element of the range and the step is automatically inferred based on that?

turbolent (commented):
Nice! Thank you for writing up this proposal 👍

cadence/20230602-cadence-range-progression.md
darkdrag00nv2: Yes, I meant to add a `downTo` here.

cadence/20230602-cadence-range-progression.md
darkdrag00nv2: @turbolent In the current proposal, it depends on whether `step` is provided or not. If step is provided, then `Progression` else `Range`.

darkdrag00nv2:
@turbolent @dsainati1 Thanks for your review comments. I agree with all of the suggestions and will modify the FLIP with the following new proposal:

We will not have operators to create a `*Range`. We can just have standard library functions. There will be two types & constructors: `InclusiveRange` and `ExclusiveRange`. I'll revert the lexer/parser changes I made in my draft PR: https://github.com/onflow/cadence/pull/2523. This would eliminate the need of operators & all the readability concerns with it.

The only source of confusion for me is whether we want to unify `Range` and `Progression` types as @dsainati1 proposed (originally proposed - https://github.com/onflow/cadence/issues/2482#issuecomment-1546977952) or keep them separate as @turbolent proposed (see https://github.com/onflow/cadence/issues/2482#issuecomment-1548703720).

For unified, the constructors will be `InclusiveRange`, `InclusiveRangeWithStep`, `ExclusiveRange` & `ExclusiveRangeWithStep`  due to lack of function overloading.

For separate, the constructors will `InclusiveRange`, `InclusiveProgression`, `ExclusiveRange` & `ExclusiveProgression`.

dsainati1:
IMO it's better to have them be one type, if only because the names are clearer. IMO the function name `InclusiveRangeWithStep` is much clearer than `InclusiveProgression`, so I prefer this naming scheme in general.

cadence/20230602-cadence-range-progression.md
darkdrag00nv2: It will be 20. I'll make it explicit.

cadence/20230602-cadence-range-progression.md
darkdrag00nv2: Not needed after constructor functions.

cadence/20230602-cadence-range-progression.md
darkdrag00nv2: Removed the concept of Progression. We only have ranges now.

cadence/20230602-cadence-range-progression.md
darkdrag00nv2: Made it explicit.

cadence/20230602-cadence-range-progression.md
```diff
@@ -0,0 +1,179 @@
...
+authors: darkdrag00n (darkdrag00n@proton.me)
+sponsor: Bastian Müller (bastian@dapperlabs.com)
+updated: 2023-06-02
+---
+
+# FLIP 96: Range & Progression Type
+
+## Objective
+
+This FLIP proposes adding the `Range` & `Progression` types to Cadence.
```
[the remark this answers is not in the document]
darkdrag00nv2: Done

cadence/20230602-cadence-range-progression.md
```diff
@@ -0,0 +1,179 @@
...
+Both of these types can then be used inside a `for-in` loop.
+
+## Motivation
+
+Modern langauges often provide a concise way for highly repetitive use cases. One of them is looping between a start & end integer possibly with a step.
+
+Presently, to loop over a start & end integer value, users have to either create an `Array` or use an imperative style `while` loop. A concise syntax for e.g. `1 .. 10` would enhance the readability of the language.
```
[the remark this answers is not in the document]
darkdrag00nv2: Nice, added in the "Objective" section based on the current proposal.

cadence/20230602-cadence-range-progression.md
```diff
@@ -0,0 +1,179 @@
...
+A `Range` will be defined as follows:
+
+```go
+struct Range<T: Integer> {
+   let start: T
+   let endInclusive: T
+}
+```
+
+It can be instantiated using the `..` & `downTo` binary operators & can be used in a `for-in` loop as follows:
```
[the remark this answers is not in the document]
darkdrag00nv2: Nice suggestion. Thanks! Have updated the proposal based on this.

cadence/20230602-cadence-range.md
```diff
@@ -0,0 +1,196 @@
...
+}
+
+let exclusiveRangeValueBackwards = ExclusiveRangeWithStep(132, 33, -3)
+exclusiveRangeValueBackwards.length // 15
+exclusiveRangeValueBackwards.contains(34) // True
+for i in exclusiveRangeValue {
+    // i will be 132, 125, 118, ... 41, 34
+}
+
+let exclusiveRangeValueEmpty = ExclusiveRangeWithStep(10, 10, 2) // Empty since start == endExclusive
```
dsainati1, line 148: IMO this should just be a runtime error; just like attempting to instantiate any other ranges with invalid parameters. Better to error on the creation than to silently create an empty range and confuse developers when their loop doesn't execute.

dsainati1 (commented):
Nice! I like this proposal a lot more after the changes, it seems like it's going to be a nice feature.

cadence/20230602-cadence-range.md
```diff
@@ -0,0 +1,196 @@
...
+2. In `ExclusiveRange`, `endExclusive` is guaranteed to not be part of the generated sequence. As a result, if `start == endExclusive`, the resulting sequence is empty.
+
+Both the `Range` types will be usable in a `for-in` loop.
+
+### Constructor Functions
+Constructor functions will be defined to instantiate variables of type `InclusiveRange` & `ExclusiveRange`. They are defined below.
+
+- `InclusiveRange(start: T, endInclusive: T)`: Creates an `InclusiveRange` with the provided start and end with a default value of step as 1 if `start <= endInclusive` or -1 if `start > endInclusive`.
+
+- `InclusiveRangeWithStep(start: T, endInclusive: T, step: T)`: Creates an `InclusiveRange` with the provided start, end and step.
```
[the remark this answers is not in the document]
darkdrag00nv2: I see, I got confused with the lack of overloading in Cadence with the lack of overloading in native functions. You point is correct. Thanks for sharing the example of assert as well.

I'll update the proposal.

cadence/20230602-cadence-range.md
darkdrag00nv2: Sure, makes sense.

This issue won't be there if we go with Supun's suggestion of having a single `Range` type.

cadence/20230602-cadence-range.md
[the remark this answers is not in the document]
turbolent: There isn't really support for overloading, but rather, built-in functions may have variable arguments, i.e. allow additional arguments (see `FunctionType.RequiredArgumentCount`)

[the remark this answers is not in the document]
darkdrag00nv2: Yes, was able to get it to work in my PR. Thanks!

cadence/20230602-cadence-range.md
```diff
@@ -0,0 +1,196 @@
...
+   let start: T
+   let endInclusive: T
+   let step: T
+}
+
+struct ExclusiveRange<T: Integer> {
+   let start: T
+   let endExclusive: T
+   let step: T
+}
```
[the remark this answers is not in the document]
turbolent: Like I described in the comment above, https://github.com/onflow/flips/pull/96#discussion_r1214973241, I still feel like it is a good idea to follow other languages and have separate types, one for inclusive end, and one for exclusive end. 

For example, Swift has [`Range`](https://developer.apple.com/documentation/swift/range) and [`ClosedRange`](https://developer.apple.com/documentation/swift/closedrange), and Rust has [`Range`](https://doc.rust-lang.org/std/ops/struct.Range.html) and [`RangeInclusive`](https://doc.rust-lang.org/std/ops/struct.RangeInclusive.html).

At the moment the type bound is `Integer`, in which case you are right @SupunS. 

However, in the future we might want to make the bound less restrictive, and e.g. just require the type to be comparable (e.g. using a new `Comparable` interface), like in Swift (type bound is [`Comparable`](https://developer.apple.com/documentation/swift/comparable)) and Rust (type bound is [`PartialOrd`](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html)). It might be good to future proof the API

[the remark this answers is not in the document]
SupunS: Ok, so I guess the problem is, if the bound type becomes something like `Comparable`, it would be difficult/impossible to adjust the `end` parameter (to make it inclusive/exclusive), because the user wouldn't know what the previous/next value of the range would be, because it solely depends on the implementation of the 'comparator' of the value.

In that case, makes sense to have two types 👍

cadence/20230602-cadence-range.md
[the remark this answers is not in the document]
SupunS: The FLIP still proposes the old API. Can you please update the APIs here as well?

cadence/20230602-cadence-range.md
```diff
@@ -0,0 +1,196 @@
...
+
+None
+
+### Dependencies
+
+None
+
+### Engineering Impact
+
+No engineering impact is expected.
```
SupunS, line 175: Just a minor knit: This would need engineering effort to implement the feature

cadence/20230602-cadence-range.md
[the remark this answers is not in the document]
darkdrag00nv2: Yes, on it now. Will update the FLIP shortly.

cadence/20230602-cadence-range.md
darkdrag00nv2: Done, I've updated this to be a runtime error.

cadence/20230602-cadence-range.md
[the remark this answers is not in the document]
darkdrag00nv2: @SupunS Updated the FLIP with the updated API.

cadence/20230602-cadence-range.md
darkdrag00nv2: Done, added a line about the engineering effort.

darkdrag00nv2:
Updated the FLIP

- Merged `InclusiveRange` and `InclusiveRangeWithStep` into a single constructor function `InclusiveRange` with `step` being a non-required argument. Did a similar change for `ExclusiveRange`.
- Renamed `endInclusive` and `endExclusive` in the two types to be `end` since the name of the type is indicative enough
- Made empty `ExclusiveRange` creation a runtime error.

cadence/20230602-cadence-range.md
```diff
@@ -0,0 +1,191 @@
...
+2. In `ExclusiveRange`, `end` is guaranteed to not be part of the generated sequence. As a result, if `start == end`, the resulting sequence is invalid & we disallow creation of such `ExclusiveRange`.
+
+Both the `Range` types will be usable in a `for-in` loop.
+
+### Constructor Functions
+Constructor functions will be defined to instantiate variables of type `InclusiveRange` & `ExclusiveRange`. They are defined below.
+
+- `InclusiveRange(start: T, end: T, step: T)`: Creates an `InclusiveRange` with the provided start, end and step. `step` will be an optional argument with a default value of step as 1 if `start <= end` or -1 if `start > end`.
+
+- `ExclusiveRange(start: T, end: T, step: T)`: Creates an `ExclusiveRange` with the provided start, end and step. `step` will be an optional argument with a default value of step as 1 if `start < end` or -1 if `start > end`.
```
turbolent, line 55: Here, the argument labels for `start`, `end`, and `step` are required. In the usages below, they are incorrectly omitted.

Maybe either fix the use-sites, e.g. `InclusiveRange(11, 21)`, to specify the required argument labels, e.g. `InclusiveRange(start: 11, end: 21)`, or do not require them:

```suggestion
- `InclusiveRange(_ start: T, _ end: T, step: T)`: Creates an `InclusiveRange` with the provided start, end and step. `step` will be an optional argument with a default value of step as 1 if `start <= end` or -1 if `start > end`.

- `ExclusiveRange(_ start: T, _ end: T, step: T)`: Creates an `ExclusiveRange` with the provided start, end and step. `step` will be an optional argument with a default value of step as 1 if `start < end` or -1 if `start > end`.
```

Maybe we can always require the argument label for step to make it clear the signature is not e.g. start/step/end

bjartek:
Can we have a helper method that has some defaults? For instance for IntRange having it as step 1 is a start. Also what about specifying length and not end?

cadence/20230602-cadence-range.md
```diff
@@ -0,0 +1,191 @@
...
+authors: darkdrag00n (darkdrag00n@proton.me)
+sponsor: Bastian Müller (bastian@dapperlabs.com)
+updated: 2023-06-19
+---
+
+# FLIP 96: Range Type
+
+## Objective
+
+This FLIP proposes adding the concept of a `Range` to Cadence. It proposes two new types, `InclusiveRange` and `ExclusiveRange` along with constructor functions to create them. It also proposes their usage inside a `for-in` loop.
```
turbolent, line 13: Small nit: The proposal now proposes multiple types

```suggestion
# FLIP 96: Range Types

## Objective

This FLIP proposes adding the concept of a range to Cadence. It proposes two new types, `InclusiveRange` and `ExclusiveRange` along with constructor functions to create them. It also proposes their usage inside a `for-in` loop.
```

cadence/20230602-cadence-range.md
```diff
@@ -0,0 +1,191 @@
...
+////////////////////
+
+let inclusiveRangeValue = InclusiveRange(11, 21)
+inclusiveRangeValue.count // 11
+inclusiveRangeValue.contains(20) // True
+for i in inclusiveRangeValue {
+    // i will be 11, 12, 13, ... 20, 21
+}
+
+let inclusiveRangeValueWithStep = InclusiveRange(11, 20, 2)
```
SupunS, line 87: The argument label `step` is required here and all functions below that take the `step` argument.

```suggestion
let inclusiveRangeValueWithStep = InclusiveRange(11, 20, step: 2)
```

SupunS (approved):
The overall design looks good tom me!

cadence/20230602-cadence-range.md
```diff
@@ -0,0 +1,191 @@
...
+1. `step` must be non-zero i.e. `abs(step) > 0`.
+
+2. The value of `step` must lead to the sequence moving towards the end.
+
+Violation of the above constraints will lead to runtime errors during instantiation.
+
+### Public Members
+Both `InclusiveRange` & `ExclusiveRange` will also provide the following public member variables and functions:
+
+1. `count`: Returns the count of integers included in the Range.
```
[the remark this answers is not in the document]
darkdrag00nv2: Sure, I'll remove.

darkdrag00nv2:
@bjartek 

> For instance for IntRange having it as step 1 is a start.

The step argument in the constructor is optional. It already supports step of 1 or -1.

> Also what about specifying length and not end?

Sounds useful but I think the argument against `count` from @turbolent  also applies here. This might not be useful if we drop the `Integer` bound. So may be we can revisit and add this at a later point if there is demand.

cadence/20230602-cadence-range.md
darkdrag00nv2: Thanks, updated everywhere.

cadence/20230602-cadence-range.md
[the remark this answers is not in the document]
darkdrag00nv2: Removed.

cadence/20230602-cadence-range.md
```diff
@@ -0,0 +1,191 @@
...
+struct InclusiveRange<T: Integer> {
+   let start: T
+   let end: T
+   let step: T
+}
+
+struct ExclusiveRange<T: Integer> {
+   let start: T
+   let end: T
+   let step: T
```
darkdrag00nv2, line 39: There is a small thing to note with these two definitions which I noticed while implementing this.

The type of `step` is `T` which means for `UInt*`, step cannot be negative. So, a backward range where `start > end` won't be possible in this representation if `T` is `UInt*`.

darkdrag00nv2: I guess that is okay since the user can cast the `UInt*` to `Int*` if they wish to have a backward range. Just sharing here for comments & I'll add this note in the FLIP as well once we have some discussion on this point.

turbolent: Agreed, I think it's OK.

Interestingly enough, in Kotlin, the equivalent progression type is not generic, but there's a dedicated type for each T, and the step is actually always integer, e.g. https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.ranges/-u-int-progression/

darkdrag00nv2: @dsainati1 @SupunS Thoughts? I am okay with keeping `step` as it is or converting it to a `Int` independent of `T` similar to Kotlin.

SupunS: Yeah, the existing one seems OK to me too.

If we ever feel that it is a limitation, we can make it an `Integer` in the future: change would be backward compatible. But again, I don't think we would have a need to do that.

SupunS (approved):
Nice, LGTM!

dsainati1 (approved):
Looks good to me! Thanks for the hard work

cadence/20230602-cadence-range.md
```diff
@@ -0,0 +1,191 @@
...
+
+Example usage of `InclusiveRange` and `ExclusiveRange` types:
+
+```cadence
+////////////////////
+// InclusiveRange //
+////////////////////
+
+let inclusiveRangeValue = InclusiveRange(11, 21)
+inclusiveRangeValue.count // 11
```
turbolent, line 80: The examples still mention the `count` field here and below, maybe remove them

turbolent (approved):
Nice! 🚀

cadence/20230602-cadence-range.md
darkdrag00nv2: Done, thanks for pointing out.

turbolent:
Let's leave this open for another week to invite additional feedback. If there's no additional feedback the FLIP is ready to get accepted and merged 🎉

cadence/20230602-cadence-range.md
```diff
@@ -0,0 +1,183 @@
+---
+status: draft
```
turbolent, line 2:
```suggestion
status: accepted
```

cadence/20230602-cadence-range.md
```diff
@@ -0,0 +1,183 @@
+---
+status: accepted
+flip: 96
+authors: darkdrag00n (darkdrag00n@proton.me)
+sponsor: Bastian Müller (bastian@dapperlabs.com)
+updated: 2023-06-19
```
turbolent, line 6:
```suggestion
updated: 2023-07-05
```

turbolent (approved):

PR #97, opened 2023-06-02
Add mutability restrictions vision

SupunS:
This is an explanation of how the several proposed mutability FLIPs can help to solve the existing mutability problems.


turbolent (approved):
Nice! 👍

PR #111, opened 2023-06-09
Emit events from function conditions

SupunS:
This FLIP is the same as: https://github.com/onflow/flips/pull/82, but updated with the new objectives.


cadence/20230417-events-emitted-from-interfaces.md
```diff
@@ -0,0 +1,110 @@
...
+https://github.com/onflow/cadence/issues/2069
+
+## Prior Art
+
+N/A
+
+## Questions and Discussion Topics
+
+- Is it still a requirement to allow interfaces to declare concrete events (not just type requirements),
+and being able to use those events in `emit` statements?
```
SupunS, line 110: Not sure if we still want to do this. For e.g. even if the `FungibleToken` doesn't need this, would it be needed by other interfaces out there, etc.?

cadence/20230417-events-emitted-from-interfaces.md
```diff
@@ -0,0 +1,110 @@
...
+This allows interfaces to guarantee the emission of events, when a function (defined in the interface) is called.
+
+### Example:
+
+Suppose the `FungibleToken` is a contract, and has a `TokenWithdrawal` event declared in it.
+
+With function pre/post conditions being able to emit events, the `withdraw` function would look like below.
+
+```cadence 
+pub contract FungibleToken {
```
turbolent, line 45: Should this be a contract interface?

SupunS: This has been discussed in the original FLIP #82, and the reason for keeping this as a contract (similar to the original FLIP) is, the `FungibleToken` going to be a contract in the new V2 standard.

This can be made an interface for the time being, but then it's still an open question whether we need to allow defining concrete events in interfaces. Already mentioned this in the "Questions and Discussion Topics" section: https://github.com/onflow/flips/pull/111#discussion_r1224447786

turbolent: I see, I assumed that changes was part of this proposal. We can maybe propose the removal of nested type requirements and allowing the declaration of concrete events in interfaces in a separate FLIP 👍

cadence/20230417-events-emitted-from-interfaces.md
turbolent: We can maybe propose it in a separate FLIP

turbolent (approved):
Simple and useful addition 👍

As Deniz had described in the initial draft, this FLIP just proposes "syntactic sugar" for something that is already currently possible: A condition may call a function that emits the event.

bitcoinfinancier (approved):

joshuahannan (approved):
Looks great to me!

cadence/20230417-events-emitted-from-interfaces.md
```diff
@@ -0,0 +1,110 @@
...
+   
+    pub event TokenWithdrawal(type: Type, amount: UFix64, from: Address?)
+   
+    pub resource interface Provider{ 
+        pub fun withdraw(amount: UFix64): @{FungibleToken.Vault} { 
+            post { 
+                result.balance == amount: "Withdrawal amount must be the same as the balance of the withdrawn Vault"
+                emit TokenWithdrawal(type:type, amount:amount, from:from)
```
satyamakgec, line 53: isn't it changing the definition of post and pre conditions? Because by nature these statements should be pure and not write anything to the blockchain while making then emit events we are writing logs in the blockchain.

SupunS: Added: https://github.com/onflow/flips/pull/111/commits/dfa753cb65ea7ac131d1c4371dec2ebd14858c56

cadence/20230417-events-emitted-from-interfaces.md
```diff
@@ -0,0 +1,115 @@
...
+    }
+}
+```
+
+Here, everytime the `withdraw` is invoked, an `TokenWithdrawal` event will be emitted.
+Thus, the interface can ensure the proper event is emitted, and can take that burden off the implementation.
+
+### Drawbacks
+
+In the [FLIP for purity/mutability analysis and restriction in function conditions](https://github.com/onflow/flow/pull/1056),
```
turbolent, line 65: Maybe link to the committed FLIP instead:

```suggestion
In the [FLIP for `view` functions](https://github.com/onflow/flips/blob/main/cadence/20220715-cadence-purity-analysis.md),
```

cadence/20230417-events-emitted-from-interfaces.md
```diff
@@ -0,0 +1,115 @@
...
+function conditions were considered to be `view`, which restricts the modification of the state inside a condition.
+Emitting of events has been categorized as an operation with side effects, and as such, emitting events is restricted
+in function conditions, in Stable Cadence.
+
+This proposal contradicts the previous FLIP, and suggests allowing events to be emitted in a `view` function context.
```
turbolent, line 70: This FLIP actually does not contradict the `view` functions FLIP: The previous FLIP noted that emitting events (and logging) are permitted side-effects, as they are not observable

```suggestion
```

SupunS: ah right, I misread the FLIP there

turbolent:
@SupunS Looks like there has not been any new feedback, so I think we can approve and merge 🎉 
We can propose the change from nested type requirements to allowing the declaration of concrete events in interfaces in a separate FLIP

SupunS:
Thanks everyone for the suggestions and feedback!

This FLIP (and the original FLIP #82) has been open for a while, and it seems we are all happy with the current proposal, and seems there are no any remaining concerns.
Thus, I'm going to consider this FLIP as accepted 🎉 🎉 🎉

You can follow how the implementation is progressing here ➡️ https://github.com/onflow/cadence/pull/2589

PR #118, opened 2023-07-11
FLIP for removing type requirements

dsainati1:
Adds a FLIP proposing to remove type requirements and change the semantics of event declarations in interfaces. 

Part of https://github.com/onflow/cadence/issues/1283

### DIFF
--- a/cadence/20230711-remove-type-requirements.md
+++ b/cadence/20230711-remove-type-requirements.md
@@ -6,7 +6,7 @@ sponsor: Daniel Sainati (daniel.sainati@dapperlabs.com)
 updated: 2023-07-24
 ---
 
-# Remove Type Requirements from Cadence
+# FLIP 118: Remove Type Requirements from Cadence
 
 ## Objective
 
--- a/cadence/20230711-remove-type-requirements.md
+++ b/cadence/20230711-remove-type-requirements.md
@@ -1,5 +1,5 @@
 ---
-status: draft 
+status: Approved 
 flip: 118
 authors: Daniel Sainati (daniel.sainati@dapperlabs.com)
 sponsor: Daniel Sainati (daniel.sainati@dapperlabs.com)
--- a/cadence/20230711-remove-type-requirements.md
+++ b/cadence/20230711-remove-type-requirements.md
@@ -3,7 +3,7 @@ status: draft
 flip: NNN (do not set)
 authors: Daniel Sainati (daniel.sainati@dapperlabs.com)
 sponsor: Daniel Sainati (daniel.sainati@dapperlabs.com)
-updated: 2023-07-11
+updated: 2023-07-24
 ---
 
 # Remove Type Requirements from Cadence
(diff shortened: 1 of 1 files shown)


austinkline:
Would it be possible to use NFT/FT definitions as examples and show how they'd change based on this FLIP?

dsainati1:
cc @joshuahannan are the contract rewrites for the FT/NFT contracts in a state where we can use them for this FLIP?

dsainati1:
Practically speaking what @bluesign is suggesting would mean that when an event is declared inside an interface, it would only be available inside that interface's definition. E.g. if I had some event

```
contract interface I {
   event Foo()
}
```

Then `Foo` would only be available inside of `I`. Effectively it would be a private type definition. Would it be reasonable then to require it to be declared with `access(self)` to reflect this?

bluesign:
> Would it be reasonable then to require it to be declared with `access(self)` to reflect this?

I think emit already limits to defined contract scope when called. I think `access(self)` is not necessary in this case.

SupunS:
That's an interesting point. I think previously the requirement of declaring events as `pub`/`access(all)` was to allow implementations to have access to it (especially in interfaces, for type requirements). But now that requirement is gone, maybe mandating events to be defined as non-public is a good move?

That way, there won't be a need to keep this special casing of `emit` statements (to not allow emitting out-of-scope events) as well. The usual access modifiers would handle it.

cadence/20230711-remove-type-requirements.md
```diff
@@ -0,0 +1,126 @@
...
+From the implementation side, 
+they require a great deal of special-cased behavior that complicates the Cadence codebase. 
+
+Removing this feature would simplify the language both for the users and the implementors, 
+and with the addition of the ability to emit events direclty from interfaces, 
+the primary use-case for this feature (requiring contracts to define certain events) has been removed. 
+
+## User Benefit
+
+This will simplify the language for users.
```
SupunS, line 58: Maybe also add a code example? as in, how a simple type requirement looks like, and how would the same code would look like after the change? 
i.e: if there is a function that accepts an `Interface.TypeRequirement`, then it would need to have an interface instead of the type requirement, and the function would need to accept a restricted type.
e.g.:
```cadence
contract interface Interface {
    struct NestedType {}

    fun foo(v: NestedType) {}
}
```

would eventually be something like:
```cadence
contract interface Interface {
    struct NestedInterface {}

    fun foo(v: {NestedInterface}) {}
}
```

cadence/20230711-remove-type-requirements.md
```diff
@@ -0,0 +1,146 @@
+---
+status: draft 
+flip: NNN (do not set)
```
turbolent, line 3:
```suggestion
flip: 118
```

joshuahannan (approved):
Feeling good about this. The token standards don't need to be able to define events in the interface anymore since the contract is a concrete type now instead of an interface

cadence/20230711-remove-type-requirements.md
```diff
@@ -0,0 +1,146 @@
+---
+status: draft
```
dsainati1, line 2:
```suggestion
status: Approved 
```

PR #120, opened 2023-07-13
[Cadence] FLIP - random function

tarakby:
Proposes a FLIP to:
- Update the `unsafeRandom` function name after the underlying implementation has been secured using Flow protocol native random beacon.
- Update the interface to a safer and more convenient one (generalized types and a modulo parameter)


cadence/20230713-random-function.md
```diff
@@ -0,0 +1,128 @@
...
+### Drawbacks
+
+Step (2) in [Design Proposal](#design-proposal) introduces a breaking change.
+
+### Alternatives Considered
+
+- `fun random<T: UnsignedInteger>(): T` can provide more flexibility
+to developers. However, any random output of type `T` can be built using `fun random(): UInt64`.
+Adding a new built-in function can still be considered.
+- An optional modulus can be added as a input `fun random([modulus; UInt64]): UInt64` to produce
```
tarakby, line 71: Here I'm assuming [this](https://github.com/onflow/flow-go/blob/master/crypto/random/rand.go#L75-L105) Go logic can be implemented on Cadence (basically using bit operations on Cadence).

dsainati1:
My only concern with this change is that I think it may communicate the wrong thing to Cadence users. I understand how from the protocol perspective our randomness is no longer "unsafe", but from the perspective of a Cadence user, I think they will probably not consider it to be such. As far as they are concerned, the randomness (used naively) is still "unsafe" because it is still exploitable by a transaction that reverts to postselect a favorable result. Changing the name of the function to remove the `unsafe` prefix I think would suggest to people that the function is "safe" in the sense that it is immune to this exploit, which it's not. 

Users can use certain patterns to make their randomness truly safe, but they need to structure their code in a specific way for this to work, and I am worried that people won't realize they need to do this without the built-in warning that a name like `unsafeRandom` provides. IMO we should have the name (or the API) here indicate that the user needs to do something additional beyond just calling the function to get a random number; something like `revertibleRandom`, for example.

tarakby:
Thanks @dsainati1

I see your point but I'll add counter-arguments to keep the discussion going:
- Although any transaction with a random call can be reverted, some applications require randomness and have no incentive in reverting the result. Flow-core contracts for instance [use randomness](https://github.com/onflow/flow-core-contracts/blob/2da88fa/contracts/epochs/FlowEpoch.cdc#L772) without reverting risks. A well-respected dapp also (assigning random NFTs for example) wouldn't want to add a reverting logic to their contract.  For these applications, the new exposed randomness is totally safe. Adding `unsafe` would even mean there is something wrong with the contract while there isn't.
- result abortion is an inherent property of atomic smart-contract platforms, rather than specific functions like randomness. The non-safety comes from the language property rather than the randomness itself. We can argue that any logic written with Cadence is revertible, and yet we do not add the prefix `unsafe` or `revertible` for all Cadence functions. Any developer should be aware that a state change in a transaction is revertible, not only when it uses randomness (a deterministic game transaction can abort if the player isn't happy with the result for instance) 
- a language should try to provide as safe as possible tools, but it can't guarantee that any program written in that language is 100% safe. There is always some responsibility that falls on the language developer. I guess it's possible to write very unsafe contracts and Cadence can't prevent that (I'm thinking of the cryptography tools as an example).

turbolent:
Great points @dsainati1 and @tarakby! @tarakby Could you please incorporate the concerns and your feedback into the FLIP itself? That way the reasoning is not lost

cadence/20230713-random-function.md
turbolent: That might be possible, but seems error-prone and currently probably inefficient. Could we maybe provide this to users for convenience reasons?

cadence/20230713-random-function.md
```diff
@@ -0,0 +1,128 @@
...
+2. As a second step, deprecate `fun unsafeRandom(): UInt64` as part of the
+stable cadence release (Cadence v1.0).
+
+### Drawbacks
+
+Step (2) in [Design Proposal](#design-proposal) introduces a breaking change.
+
+### Alternatives Considered
+
+- `fun random<T: UnsignedInteger>(): T` can provide more flexibility
```
turbolent, line 68: Same as below: Though it is possible to build this using a lower-level primitive, it seems convenient to have a function that simply returns the expected type without having to perform any arithmetic and type conversion

turbolent (commented):
Looks good, though it would be nice to provide a more convenient API (a combination of the alternatives?) which reduces potential misuse

cadence/20230713-random-function.md
tarakby: I know I have asked you before about adding "native" Cadence functions to provide:
 1. randoms module `n`.
 2. other integer types. 
 
 However, thinking more about this and learning more about what's possible with Cadence, I realized the functions are implementable on Cadence and the operations would look very similar performance-wise to their implementation in Go:
 - modulo function only requires bit-wise operations (shift, or, and)
 - supporting other big types requires big number shifting (Go big Int shifting underneath)
 
I thought in this case it's usually preferred to keep the "native" functions as thin as possible and implement more logic with Cadence (unless there is a big performance gap). What do you think?

joshuahannan (approved):
Looks good! I think I'm fine with accepting that developers need to know that random is revertible without reading the function name

janezpodhostnik:
Would it be better if FVM was injecting the `random` function (via `Environment.Declare(valueDeclaration stdlib.StandardLibraryValue)`) so that it would be entirely handled by the FVM this way we could remove the random from the runtime interface entirely. @turbolent?

turbolent:
@janezpodhostnik That could be an option. There's always the question what should be part of the built-in functionality / standard library, and what is only provided by a particular environment.

So far, the random function is part of the built-in functionality (standard library). I feel like every implementation of Cadence should provide one, so developers can be assured the function is available and reason about the behaviour. For now I'd keep it that way, however, I don't feel strongly about it.

(The same question / reasoning applies to other functions/APIs (like the block API).

cadence/20230713-random-function.md
turbolent: It is true that the higher-level API could be implemented in Cadence using the low-level function.

However, that would mean that developers would *have* to do so, and they might get it wrong.
One of the goals of Cadence is to provide safe and convenient APIs, so I feel like we as implementers should take on the "burden" of implementing the high-level API.

We could/would still keep the interface between Cadence and the host (embedder; e.g FVM) low-level.

tarakby: I updated the FLIP: https://github.com/onflow/flips/pull/120/commits/7452c6e3a20c0f58c3741cd1486bdf4f4b509b00

Please have a look at the Cadence example too.

cadence/20230713-random-function.md
```diff
@@ -0,0 +1,159 @@
...
+# FLIP 120: Rename unsafeRandom function to random
+
+## Objective
+
+The purpose is :
+- to rename the current `fun unsafeRandom(): UInt64` function to
+`fun random(): UInt64`.
+This can be done by introducing a new `random` function, and eventually
+deprecating `unsafeRandom` (breaking change).
+- to expand the current function to a more safe and convenient `fun random<T: UnsignedInteger>([modulo: T]): T` where `UnsignedInteger` covers more Cadence's unsigned integer types and `modulo` is an optional upper-bound argument.
```
turbolent, line 18:
```suggestion
The purpose of this FLIP is to:
- Rename the current `fun unsafeRandom(): UInt64` function to `fun random(): UInt64`.
  This can be done by introducing a new `random` function, and eventually deprecating `unsafeRandom` (breaking change).
- Expand the current function to a more safe and convenient `fun random<T: UnsignedInteger>([modulo: T]): T`,
   where `UnsignedInteger` covers any of Cadence's unsigned integer types, and `modulo` is an optional upper-bound argument.
```

cadence/20230713-random-function.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Motivation
+
+### function rename
```
turbolent, line 22: Use title case for headings, here and below:

```suggestion
### Function Rename
```

cadence/20230713-random-function.md
```diff
@@ -0,0 +1,159 @@
...
+
+The Flow Virtual Machine (FVM) provides the implementation of `unsafeRandom`
+as part of the Flow protocol. The FVM implementation has been using the block hash as
+a source of entropy. This source can be manipulated by miners (i.e consensus nodes)
+and should not be relied on to derive secure randomness, hence the `unsafe`
+suffix in the function name.
```
turbolent, line 28: Break lines at line endings and logical breaks (`,`, `and`, etc).

That makes it easier to review changes and does not affect rendering.

```suggestion
The Flow Virtual Machine (FVM) provides the implementation of `unsafeRandom` as part of the Flow protocol. 
The FVM implementation has been using the block hash as a source of entropy. 
This source can be manipulated by miners (i.e consensus nodes)
and should not be relied on to derive secure randomness, 
hence the `unsafe` suffix in the function name.
```

cadence/20230713-random-function.md
```diff
@@ -0,0 +1,159 @@
...
+
+A more convenient way of using `random` is to cover unsigned integer types `UInt8`, `UInt16`, `UInt32`, `UInt64`, `UInt128`, `UInt256`, `Word8`, `Word16`, `Word32`, `Word64`. The type applies to the optional argument `modulo` as well as the returned value. This would abstract the complexity of generating randoms of different types using 64-bits values as a building block. The new suggested function signature is therefore `fun random<T: UnsignedInteger>([modulo: T]): T`, where `T` can be any type from the above list.
```
turbolent, line 51:
```suggestion
A more convenient way of using `random` is to cover all unsigned integer types (currently, `UInt8`, `UInt16`, `UInt32`, `UInt64`, `UInt128`, `UInt256`, `Word8`, `Word16`, `Word32`, `Word64`). 
The type applies to the optional argument `modulo` as well as the returned value. 
This would abstract the complexity of generating randoms of different types using 64-bit values as a building block. 
The new suggested function signature is therefore `fun random<T: UnsignedInteger>([modulo: T]): T`, 
where `T` can be any type from the above list.
```

cadence/20230713-random-function.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Tutorials and Examples
+
+```
+let r1 = randomUInt8()                    // r1 is of type `UInt8`
+let r2 = randomWord16()                   // r2 is of type `Word16`
+let r3 = randomUInt128(UInt128(3))        // r3 is of type `UInt128` and is strictly less than 3
+let r4 = randomUInt128(UInt128(1 << 100)) // r4 is of type `UInt128` and is of at most 100 bits
+let r5 = randomWord64(Word64(r1))         // r5 is of type `Word64` and is strictly less than `r1`
+let r6 = randomUInt64(UInt64(0))          // this would panic
```
turbolent, line 120: - The `modulo` parameter requires an argument label
- There is only the single `random` function, and it is generic (it has a type parameter), therefore 
  - The called function is always `random`
  - A type argument must be provided, if it cannot be inferred (from the `modulo` argument)
  - There is no need to convert `Int` literals using conversion functions, the literals are typed accordingly through type inference. The explicit conversion `Word64(r1)` is needed, because `r1` is of type `UInt8`, but the expected type  is `Word64` (type conversion must be performed explicitly)

```suggestion
let r1 = random<UInt8>()                     // r1 is of type `UInt8`
let r2 = random<Word16>()                    // r2 is of type `Word16`
let r3 = random<UInt128>(modulo: 3)          // r3 is of type `UInt128` and is strictly less than 3
let r4 = random<Word128>(modulo: 1 << 100)   // r4 is of type `Word128` and is of at most 100 bits
let r5 = random<Word64>(modulo: Word64(r1))  // r5 is of type `Word64` and is strictly less than `r1`
let r6 = random<UInt64>(modulo: 0)           // panics
let r7 = random(modulo: 2)                   // r6 is of type `Int` and is strictly less than 2
```

cadence/20230713-random-function.md
```diff
@@ -0,0 +1,159 @@
...
+## Design Proposal
+
+1. As a first step:
+  - update Cadence's [runtime interface](https://github.com/onflow/cadence/blob/8a128022e0a5171f4c3a173911944a2f43548b98/runtime/interface.go#L107) `UnsafeRandom() (uint64, error)` to `ReadRandom(byte[]) error`.
+  - add a new Cadence function `fun random<T: UnsignedInteger>([modulo: T]): T`, backed
+  by a safe FVM implementation. `fun unsafeRandom(): UInt64` remains
+  available to avoid immediate breaking changes. Note that both functions
+  would be backed by the same safe FVM implementation.
+2. As a second step, deprecate `fun unsafeRandom(): UInt64` as part of the
+stable cadence release (Cadence v1.0).
```
turbolent, line 71:
```suggestion
1. As a first step:
  - Update Cadence's [runtime interface](https://github.com/onflow/cadence/blob/8a128022e0a5171f4c3a173911944a2f43548b98/runtime/interface.go#L107) `UnsafeRandom() (uint64, error)` to `ReadRandom(byte[]) error`.
  - Add a new Cadence function `fun random<T: UnsignedInteger>([modulo: T]): T`, backed by a safe FVM implementation. 
    `fun unsafeRandom(): UInt64` remains available to avoid immediate breaking changes. 
    Note that both functions will be backed by the same safe FVM implementation.
2. As a second step, deprecate `fun unsafeRandom(): UInt64` as part of the
Stable Cadence release (aka Cadence v1.0).
```

turbolent (commented):
Great improvements! Just some fixes for the examples, and some stylistic recommendations

cadence/20230713-random-function.md
tarakby: I didn't want to say "all unsigned integer types" because the list does not include `UInt`. 
The reason is that I had doubts with what `random<UInt>()` should return, as in how many random bits it should generate. 
How does Cadence consider `UInt` internally? (for instance compilers usually consider the host CPU word size)

cadence/20230713-random-function.md
tarakby: Same comment about "any of Cadence's unsigned integer types". I excluded `UInt` so far and I didn't want to include all types.

cadence/20230713-random-function.md
turbolent: Oh, good point! `UInt` is a variable-sized type, so has no max. I can't think of a good way in which `random` would be defined for it, so variable-sized types should be excluded.

Maybe make the text more specific, by saying "all *fixed-size* unsigned integer types".

We currently do not have a type that expresses this, maybe we should introduce `FixedSizeUnsignedInteger`, and use that as the type-bound.

That way, usage of e.g. `UInt` would be statically rejected, instead of a run-time error.

Good flag about `Int` and `UInt` missing from the documentation, will open an issue for that.

cadence/20230713-random-function.md
turbolent: Ah, good points! I should have tried these out before suggesting them.

- Yes, the `r7` should be removed. I wanted to demonstrate how the type of the argument (`2`) is used to infer the type argument, i.e. it does not have to be explicitly provided.
  However, as you pointed out, the function is not defined for `Int`, as it is not an unsigned integer type.
- Right, `r4` should be adjusted to e.g. `UInt128`, `Word128` does not exist at the moment.

cadence/20230713-random-function.md
tarakby: Thanks for clarifying and opening the doc issue 👌🏼

cadence/20230713-random-function.md
```diff
@@ -125,6 +125,17 @@ such as commit-reveal schemes can be proposed in a separate FLIP.
...
+
+We can define an alternative behavior for `random` in scripts. These following options are possible:
+ 1. panic
+ 2. return a constant (zero value of the type)
+ 3. return pseudo-random numbers using a weak source of entropy
+
+Option (1) can cause confusion because valid Cadence code would fail when run as a script. Option (3) can also cause confusion as the returned numbers may look "random" while they are different than the numbers returned by a transaction, and are therefore not useful for users. (2) avoids panicking and may remind users that randomness doesn't work properly on scripts. This FLIP suggests to choose option (2).
```
aishairzay, line 158: I prefer option 3 for a few reasons and a few use-cases:

1. When learning cadence, finding out that randomness does not work in scripts sounds like a very confusing and hard to fix developer journey.

2. Going with a constant value can possibly be more confusing than helpful, and not provide expected results to a developer. i.e. I want to test a function on my contract with 1,000 random numbers through a script, and i use random() to generate them

3. For flow sight, https://github.com/Zay-Codes-Lab/flow-sight, we used scripts to translate a TX before its run into a script and simulate it. One use-case we planned previously is to re-run the script simulation many times to see if there is any discrepancy in results based on use of the `random()` function. Similar to my previous point, without the script having randomness, it wouldn't be possible to detect any changes in behavior between scripts, like there would be for the TX.

tarakby: Thanks @aishairzay! I'm happy to switch the proposal to another option if there are enough use cases that require randomness in scripts. 

In option 3, the source of randomness is weak (for instance the block hash). Randoms would be verifiable but there is a degree of biasing them by some node operators. Not sure if the use-cases above require a secure non-biasable source?

For a secure source (Flow distributed randomness used by transactions), the source cannot be retrieved for old blocks. This means that `ExecuteScriptAtLatestBlock` would work but `ExecuteScriptAtBlock` wouldn't. That's why I didn't add it as Option 4.

AlexHentschel (approved):
thank you for this excellent write-up 👏

cadence/20230713-random-function.md
```diff
@@ -148,14 +148,9 @@ such as commit-reveal schemes can be proposed in a [separate FLIP](https://githu
...
- 1. panic
- 2. return a constant (zero value of the type)
- 3. return pseudo-random numbers using a weak source of entropy
-
-Option (1) can cause confusion because valid Cadence code would fail when run as a script. Option (3) can also cause confusion as the returned numbers may look "random" while they are different than the numbers returned by a transaction, and are therefore not useful for users. (2) avoids panicking and may remind users that randomness doesn't work properly on scripts. This FLIP suggests to choose option (2).
+`executeScriptAtBlock` and `ExecuteScriptAtLatestBlock` are used to execute Cadence read-only code against the execution state at a past sealed block or the latest sealed blocked, respectively. 
+The FVM implementation of `random` uses the transaction hash to diversify the random sequence per transaction. For this reason, it is not possible to replicate Cadence's `random` behavior in scripts. Although scrips would use the same randomness source as transactions to derive randoms. It is not possible to retrieve the same random numbers when calling `random` as a transaction or as a script.
```
AlexHentschel, line 152: I am worried readers might easily confuse / conflate  'randomness source' with diversifier. I think it might not be harmful to be as explicit as we can 

```suggestion
The FVM implementation of `random` uses the transaction hash to diversify the random sequence per transaction. For this reason, it is not possible to replicate Cadence's `random` behavior in scripts. Although scrips would use the same randomness source as transactions to derive randoms, the additional diversifier is different (transaction hash vs script hash). Therefore, it is not possible to retrieve the same random numbers when calling `random` as a transaction or as a script.
```

cadence/20230713-random-function.md
tarakby: Update after talking further to the protocol team. It should be possible to have scripts use the protocol beacon as a source. Nodes should have enough past block history to retrieve the beacon source. If not the script execution may be relayed to another node.

cadence/20230713-random-function.md
tarakby: Thanks for the feedback 👌🏼  added clarification in https://github.com/onflow/flips/pull/120/commits/3cf30923ec1a3fb0dec9c868d09230ab678e0b0c

cadence/20230713-random-function.md
```diff
@@ -44,26 +46,62 @@ deterministic sequence of randoms.
 - FVM does not expose the PRG seed or state to the execution environment.
 
 These measures make the `unsafeRandom` unpredictable to the transaction execution environment
-and unbiasable by all transaction code prior to the random function call.
+and unbiasable by all transaction code prior to the random function call (for instance a transaction that calls
+`revertibleRandom()` multiple times in order to predict the future random output).
+
+All the reasons above support dropping the `unsafe` prefix.
+
+#### Revertible randomness
```
turbolent, line 54: This is a great explanation:ok_hand:

Once merged, we should put this into the documentation and an abridged version into the docstring.

turbolent (approved):
:+1:

turbolent:
The plan is to decide on this FLIP in the next Cadence Language Design Meeting on Tue 12th, Aug. 

So far there is positive sentiment. Unless there are any new insights or significant reasons to reject are brought forward, the plan is to approve the FLIP.

turbolent:
@tarakby The FLIP got approved 🎉 Please update the status and merge :-)

PR #123, opened 2023-07-27
[protocol] commit-reveal scheme for non-reverted randomness

tarakby:
WIP


protocol/20230725-commit-reveal.md
```diff
@@ -0,0 +1,124 @@
...
+
+Once the changes above are available, a commit-reveal scheme can be implemented as follows. The coin toss example described [earlier](#motivation) will be used for illustration:
+
+ - when a user submits a bidding transaction, the coin toss contract stores the block height where the bid was made and stores the bid amount on the application contract. This is a commitment by the user to use the SoR at the current block. Note that the current block's `SoR_A` isn't known to the transaction execution environment, and therefore the transaction has no way to inspect the random outcome and predict the coin toss result. The current block's `SoR_A` is only available once added to the history core-contract, which only happens at the end of the block's execution. 
+ - the coin toss contract grants the user a limited window of time (i.e a block height range) to reveal the results and claim any winnings. Failing to do so, the user loses the bid amount in favour of the coin toss contract.
```
janezpodhostnik, line 46: Might be good to reiterate here that this window is needed because the SoR history contract has a limited view into the past.

protocol/20230725-commit-reveal.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Design Proposal
+
+The proposed design is a commit-reveal pattern. 
+
+The solution requires infrastructure changes to provide new data to the transaction execution environment:
+
+ 1. a new FVM function that exposes the current block's SoR (more precisely a derived value from the protocol `SoR_A`) to Cadence runtime. Note that `unsafeRandom` only exposes randoms derived from `SoR_A` through a pseudo-random generator (PRG) but not the `SoR_A` itself. 
+ 2. a new system core-contract that stores a limited history of SoRs for the past `N` blocks. The new FVM function in (1) is only available to the history contract, and is not available to other non-system transactions. Note that system-transactions in Flow are executed at the end of each block, after all non-system transactions of the block are executed. The proposal suggests to add a new system transaction that adds the current block's `SoR_A` to the SoR history contract (it also removes the oldest block's `SoR_A`). The contract indexes the SoR history by block height.
```
janezpodhostnik, line 40:
```suggestion
 2. a new system core-contract that stores a limited history of SoRs for the past `N` blocks. The new FVM function in (1) is only available to the history contract, and is not available to other non-system transactions. Note that system-transactions in Flow are executed at the end of each block, after all non-system transactions of the block are executed. The proposal suggests to change the system transaction so that it also adds the current block's `SoR_A` to the SoR history contract (it also removes the oldest block's `SoR_A`). The contract indexes the SoR history by block height. 
```

there is one system transaction per block, but it has multiple functions/responsibilities

protocol/20230725-commit-reveal.md
```diff
@@ -0,0 +1,124 @@
...
+		destroy receipt
+
+		if coin == 1 {
+			return <-FlowToken.createEmptyVault()
+		}
+		
+		return <-self.reserve.withdraw(amount: receipt.betAmount * 2)
+}
+
+fun randomCoin(atBlock: UInt64, salt: [UInt8]) : UInt8 {
```
janezpodhostnik, line 118:
```suggestion
fun randomCoin(atBlock: UInt64, salt: UInt64) : UInt8 {
```

protocol/20230725-commit-reveal.md
```diff
@@ -0,0 +1,124 @@
...
+const let expiryWindowLength = 1000000
+
+cointossReveal(receipt: @Receipt) : @FlowToken.Vault {
+		let currentBlock = getCurrentBlock().height
+		if receipt.commitBlock >= currentBlock {
+			panic("cannot reveal yet")
+		}
+        if receipt.commitBlock + expiryWindowLength > currentBlock {
+            return <-FlowToken.createEmptyVault()
+        }
```
janezpodhostnik, line 106: I have some thoughts about this part.

the window is mainly needed because of the `SoRHistory` history limit. But I think the correct implementation of `SoRHistory.getSoR(atBlock)` should just panic if called for a block to far in the past. I think that makes this check redundant.

I think its also important to point out to the developers applying this pattern that the second call may never come.

protocol/20230725-commit-reveal.md
tarakby: Good point, though I think that's the second reason. The primary reason is that the block's SoR will become public and the user can know easily whether they won after the commit transaction. Even if we had unlimited history of SoR, the user may decide to never send the second transaction and never be "charged" the losses. In that case, an expiry window is needed to force "charging" the bet into the application account. Is that correct or am I not understanding how the escrow works?

Either way I'll add a word about the SoR history being limited.

protocol/20230725-commit-reveal.md
tarakby: The expiry window defined by the application isn't necessary equal to the history window of the core-contract history. It can be smaller if the application wishes, that's why I added that check. I will add a comment to clarify.

tarakby: I can also remove it if this is not an important detail.

protocol/20230728-commit-reveal.md
```diff
@@ -0,0 +1,125 @@
...
+		self.reserve.deposit(from: <-bet)
+    return <- receipt
+}
+
+cointossReveal(receipt: @Receipt) : @FlowToken.Vault {
+		let currentBlock = getCurrentBlock().height
+		if receipt.commitBlock >= currentBlock {
+			panic("cannot reveal yet")
+		}
+        // `expiryWindowLength` is optionally defined by the application to be less than `N`.
```
Kay-Zee, line 102: seems to be some spacing issues with the example code?

protocol/20230728-commit-reveal.md
```diff
@@ -0,0 +1,125 @@
...
+
+		if coin == 1 {
+			return <-FlowToken.createEmptyVault()
+		}
+		
+		return <-self.reserve.withdraw(amount: receipt.betAmount * 2)
+}
+
+fun randomCoin(atBlock: UInt64, salt: UInt64) : UInt8 {
+    let sor = SoRHistory.getSoR(atBlock)
```
Kay-Zee, line 120: this SOR is exposed to the user/dapp developer? Any concerns with that?

protocol/20230728-commit-reveal.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Tutorials and Examples
+
+Here is an example of a coin toss contract with one function to commit a bid, and another function to resolve the bid:
+
+```
+import SoRHistory from 0xFLOWSORHISTORY
+
+cointossCommit(bet: @FlowToken.Vault) : @Receipt {
+		let receipt <- create Receipt(
```
Kay-Zee, line 89: Almost feels like this Receipt object should have a "prg" function that is call-able, is that the intention?

tarakby: There are many ways are possible to organize the code. Here, I wanted to explicitly call `SoRHistory.getSoR` to show how to query the history SoR core-contract, and use it to seed a PRG.

protocol/20230728-commit-reveal.md
tarakby: I don't think so. That is the SoR of a past block. The current block's SoR isn't available in the contract `SoRHistory`.

protocol/20230725-commit-reveal.md
janezpodhostnik: I would maybe mention that a`SoRHistory.getSoR` will panic if called for a block to far in the past, and omit this check from the example, to make it more minimal.

To add to that two things are already omitted from the example pattern that should probably be in the final instructions we put on the doc site, but maybe don't have to be mentioned here:

- instead of collecting the bet into the reserve pool `self.reserve.deposit(from: <-bet)` I would instead have an array of pending rewards: 
    ```cadence
   let potentialWinnings <- self.reserve.withdraw(amount: bet.amount)
   potentialWinnings.deposit(from: bet)
   self.storePotentialWinnings( potentialWinnings, /* somehow not the block and the receipt this was stored for */) 
    ```
  this ensures that the users wont be able to bet so much that they  cannot be payed out. This also ensures that reserve can be emptied without affecting bets that are in flight. And it is also safer to separate the potential user funds out from the big pool.
- prevent bets of minuscule  amounts, to prevent any sort of storage bloating attacks

tarakby: sounds good, it makes sense to omit the expiry check and clarify the behaviour of `SoRHistory.getSoR`. 

Regarding the two other points, I'm wondering if that's also too detailed. The reason is that this FLIP should be about "how to use randomness in a commit-reveal" rather than "how to design a betting contract". In that sense, I wanted focus on the 2-steps pattern and querying the new core-contract. The coin-toss contract (including bets) is only an example and should not represent all usage of randomness.

janezpodhostnik (approved):

protocol/20230728-commit-reveal.md
```diff
@@ -0,0 +1,129 @@
...
+The points 1, 2 and 3 in [the design proposal](#design-proposal) need to be built as part of this proposal.
+
+### Tutorials and Examples
+
+Here is an example of a coin toss contract with one function to commit a bid, and another function to resolve the bid:
+
+```
+import SoRHistory from 0xFLOWSORHISTORY
+
+cointossCommit(bet: @FlowToken.Vault) : @Receipt {
```
turbolent, line 88:
```suggestion
fun commitCointoss(bet: @FlowToken.Vault): @Receipt {
```

protocol/20230728-commit-reveal.md
```diff
@@ -0,0 +1,129 @@
...
+				betAmount: bet.balance,
+				// commit to use randomness at the current block (still unknown)
+				commitBlock: getCurrentBlock().height
+	)
+	// commit the bet
+	self.reserve.deposit(from: <-bet)
+    return <- receipt
+}
+
+cointossReveal(receipt: @Receipt) : @FlowToken.Vault {
```
turbolent, line 99:
```suggestion
fun revealCointoss(receipt: @Receipt): @FlowToken.Vault {
```

protocol/20230728-commit-reveal.md
```diff
@@ -0,0 +1,129 @@
...
+	let currentBlock = getCurrentBlock().height
+	if receipt.commitBlock >= currentBlock {
+		panic("cannot reveal yet")
+	}
+	// `expiryWindowLength` is optionally defined by the application to be less than `N`.
+	//  if not `N` would be implicitly used as an expiry window.
+	const let expiryWindowLength = 1000000
```
turbolent, line 106:
```suggestion
	let expiryWindowLength = 1000000
```

protocol/20230728-commit-reveal.md
```diff
@@ -0,0 +1,129 @@
...
+	if receipt.commitBlock + expiryWindowLength > currentBlock {
+		return <-FlowToken.createEmptyVault()
```
turbolent, line 108: `receipt` is lost / not properly invalidated. Maybe read `receipt.commitBlock` into a temporary, perform the `destroy receipt`, then have this if block here.

protocol/20230728-commit-reveal.md
```diff
@@ -0,0 +1,129 @@
...
+import SoRHistory from 0xFLOWSORHISTORY
+
+cointossCommit(bet: @FlowToken.Vault) : @Receipt {
+	let receipt <- create Receipt(
+				betAmount: bet.balance,
+				// commit to use randomness at the current block (still unknown)
+				commitBlock: getCurrentBlock().height
+	)
+	// commit the bet
+	self.reserve.deposit(from: <-bet)
```
turbolent, line 95: What is `self.reserve`?

protocol/20230728-commit-reveal.md
```diff
@@ -0,0 +1,129 @@
...
+	
+	return <-self.reserve.withdraw(amount: receipt.betAmount * 2)
+}
+
+fun randomCoin(atBlock: UInt64, salt: UInt64) : UInt8 {
+	// query the SoR history core-contract
+	let sor = SoRHistory.getSoR(atBlock)
+	// instantiate a PRG
+	var prg = createPRG(sor, salt)
+	let rand = prg.Uint64()
```
turbolent, line 126: Maybe define `createPRG` and the `PRG` type, incl. the `Uint64` function, and leave the implementation out. 

Also, the variable is never re-assigned, so can be `let`
```suggestion
	let prg = createPRG(sor, salt)
	let rand = prg.Uint64()
```

protocol/20230728-commit-reveal.md
```diff
@@ -0,0 +1,129 @@
...
+	destroy receipt
+
+	if coin == 1 {
+		return <-FlowToken.createEmptyVault()
+	}
+	
+	return <-self.reserve.withdraw(amount: receipt.betAmount * 2)
+}
+
+fun randomCoin(atBlock: UInt64, salt: UInt64) : UInt8 {
```
turbolent, line 121:
```suggestion
fun randomCoin(atBlock: UInt64, salt: UInt64): UInt8 {
```

protocol/20230728-commit-reveal.md
```diff
@@ -0,0 +1,129 @@
...
+	// query the SoR history core-contract
+	let sor = SoRHistory.getSoR(atBlock)
+	// instantiate a PRG
+	var prg = createPRG(sor, salt)
+	let rand = prg.Uint64()
+	return rand & 1
```
turbolent, line 127: Need to explicitly convert `UInt64` to return type `UInt8`:
```suggestion
	return UInt8(rand & 1)
```

turbolent (commented):
Mainly just reviewed the code sample

protocol/20230728-commit-reveal.md
tarakby: @janezpodhostnik can you help me define `reserve` in the example please?

protocol/20230728-commit-reveal.md
tarakby: I was assuming the `PRG` object and its method are defined by another contract (which is supposed to be implemented as part of this proposal). 
I imported another contract in the example and added comments to clarify, but happy to make further adjustments.

protocol/20230728-commit-reveal.md
janezpodhostnik: Its basicaly the main pool of funds of the coin-toss contract. As it is written in the example it would be a `@FlowToken.Vault` field on the contract itself. It would be seeded with some initial money so that the contract could function.

turbolent: Thank you for the great explanation? Could you please add this information somehow to the FLIP itself? For example, maybe add comments to the code and/or the explanation as text next to the code block.

protocol/20230728-commit-reveal.md
turbolent: Maybe document in the FLIP that only past SoR are available

protocol/20230728-commit-reveal.md
```diff
@@ -0,0 +1,128 @@
...
+
+The points 1, 2 and 3 in [the design proposal](#design-proposal) need to be built as part of this proposal.
+
+### Tutorials and Examples
+
+Here is an example of a coin toss contract with one function to commit a bid, and another function to resolve the bid:
+
+```
+import SoRHistory from 0xFLOWSORHISTORY
+import PRG from 0xPRGIMPLEMENTATION
```
turbolent, line 87: Tooling convention is to use string literal imports, which get rewritten to the actual address depending on network:

```suggestion
import SoRHistory from "SoRHistory"
import PRG from "PRG"
```

protocol/20230728-commit-reveal.md
tarakby: I added a comment to the example (https://github.com/onflow/flips/pull/123/commits/c648ffa0a5bb09341d2305143a550eb2687bc84d) but Janez feel free to add a proper field if you prefer.

janezpodhostnik: I don't think Its necessary to add the field to the cadence example. I think its better to keep the focus on the usage and not on how this theoretical contract would function, but let me know if you disagree.

protocol/20230728-commit-reveal.md
```diff
@@ -0,0 +1,132 @@
...
+ 2. a new system core-contract that stores a limited history of SoRs for the past `N` blocks. The new FVM function in (1) is only available to the history contract, and is not available to other non-system transactions. Note that system-transactions in Flow are executed at the end of each block, after all non-system transactions of the block are executed. The proposal suggests to change the system transaction so that it adds the current block's `SoR_A` to the SoR history contract (it also removes the oldest block's `SoR_A`). The contract indexes the SoR history by block height. 
+ 3. an on-chain implementation of a PRG is required. A PRG is initialized with an SoR and can generate a sequence of random numbers for an application. It's up to the application to use a suitable PRG instance. At least one recommended PRG implementation should be provided as part of the FLIP implementation (precise instance to be determined).
```
sisyphusSmiling, line 41: Curious of the status and anticipated effort required for the recommended PRG implementation.

sisyphusSmiling (approved):
Great writeup!

dete:
Thanks for putting this FLIP together, @tarakby! Unlocking secure randomness for smart contract developers is super exciting!

Some suggestions:
1. It seems to me that the cost of storing 32 bytes for all blocks indefinitely may actually acceptable! It removes the need for devs to handle the (rare) edge case where the random seed for their use-case is more than two weeks in the past. Based on your ~20Mb estimate per week, it's ~1 GB per year. I think we can cope with that level of state growth today, and it's only going to be shrink relative to the total state space size as we go into the future.
2. I think "random beacon" is an accurate description of the block randomness, and will be more meaningful for most folks than "SoR". As such, I suggest the function should be `RandomBeaconHistory.getBeaconValue(atBlockHeight: x)`. It should throw if x is greater than _or equal to_ the current block height, or less than the block height at which random beacon history was recorded.
3. I would probably go so far as to make `RandomBeaconHistory` private to the Service Account, and then ensure we ship a contract on the Service Account which wraps it into something easily used by most devs. I'll include a first-draft of what I mean below.

Note: I didn't see any docs for the PRG object. Is that proposed or existing? My assumption below is that it's either new, or it can be modified to work with byte arrays instead of UInt64.

```Cadence
import PRG

resource SecureRandom {
  access(self) let initHeight: UInt64
  access(self) let prg: PRG
  access(self var fetchedRandomBeacon = false

  init() {
    let currentBlock = getCurrentBlock()

    initHeight = currentBlock.height
    prg = create PRG()

    prg.addEntropy(currentBlock.timeStamp.toBigEndianBytes())
    prg.addEntropy(currentBlock.id.toBigEndianBytes())
    prg.addEntropy(self.uuid.toBigEndianBytes())
  }

  access(all) fun addEntropy(_ entropyBytes: [UInt8) {
    prg.addEntroy(entropyBytes)
  }

  access(all) fun nextBytes(count: UInt64): [UInt8] {
    pre {
      getCurrentBlock().height > self.initHeight:
            "SecureRandom objects can't be used within the same block they are created to avoid reversion attacks."
    }
    if (!self.fetchedRandomBeacon) {
      prg.addEntropy(RandomBeaconHistory.getBeaconValue(atBlockHeight: self.initHeight))
    }

    return prg.nextBytes(count: count)
  }

  access(all) fun nextUInt64(): UInt64 {
    return UInt64.fromBigEndianBytes(self.nextBytes(count: 8))
  }

  // Returns a random value in the range min <= result < max
  access(all) fun rangedUInt64(min: UInt64, max: UInt64) {
    pre {
      min < max:
        "Max must be strictly greater than min."
    }

    let rangeLength = max - min
    let offset = self.nextUInt64() % rangeLength

    return min + offset
  }

  access(all) fun nextUInt32 ...
  access(all) fun nextUInt16 ...
  access(all) fun nextUInt8 ...

  access(all) fun nextInt64 ...
  access(all) fun nextInt32 ...
  access(all) fun nextInt16 ...
  access(all) fun nextInt8 ...

  access(all) fun nextBool(): Bool {
    let aByte = self.nextBytes(count: 1)
    return ((aByte[0] & 1) == 1)
}
```

AlexHentschel:
In the code snippet Dete posted, I think we are missing the statement setting `fetchedRandomBeacon` to `true`. I think the `access(all)` function should read something like:

```Cadence
  access(all) fun nextBytes(count: UInt64): [UInt8] {
    pre {
      getCurrentBlock().height > self.initHeight:
            "SecureRandom objects can't be used within the same block they are created to avoid reversion attacks."
    }
    if (!self.fetchedRandomBeacon) {
      prg.addEntropy(RandomBeaconHistory.getBeaconValue(atBlockHeight: self.initHeight))
      self.fetchedRandomBeacon = true // ADDED THIS LINE
    }

    return prg.nextBytes(count: count)
  }
```

AlexHentschel:
From my perspective:
* Tarak suggested function naming `SoRHistory.getSoR(UInt64)`
* Dete's criticism: too cryptic and not using terminology people are familiar with
Dete's suggestion:
* RandomBeaconHistory.getBeaconValue(UInt64)
* Tarak's and my concern: inaccurate terminology (there is a established tech term, for the output of the random beacon: 'SourceOfRandomness'). I think it would really benefit clarity and precision of our work, if we used that term. 

I think the following would be a good middle ground: `RandomBeaconHistory.SourceOfRandomnessForBlock(UInt64)`

Reasoning:
* the random beacon produces the SoR. Therefore, I feel RandomBeaconHistory is technically reasonably accurate to use as the first word. It will help devs with familiar terminology. 
* non cryptic (no abbreviations)
* we use the technically correct term Source Of Randomness for the random beacon output

protocol/20230728-commit-reveal.md
tarakby: The PRG used in the protocol is a ChaCha20 based PRG. This is a crypto-secure PRG that is not required for the on-chain case. It also might be too heavy to implement on-chain. 

A lighter version and easy to implement is XorShift128+. IMO, implementing this FLIP can include one PRG instance (for instance XorShift128+), but developers can implement other instances and are free to choose the PRG they want in their contracts.

tarakby:
Thanks @dete for taking the time and providing a feedback. 

I answer your suggestions to keep the discussion going and get other people's opinion: 

1. This sounds reasonable, I'll update. 
2. I agree `SoR` isn't clear, it should be at least expanded into `SourceOfRandomness`. I also agree that `RandomBeacon` is a more common term, although it doesn't mean the exact same thing as source of randomness. My understanding is that "Random Beacon" is the sub-protocol generating randomness and its output is a "source of randomness".  I believe the naming suggested by Alex addresses your concern while keeping the naming accurate `RandomBeaconHistory.SourceOfRandomnessAtBlock(UInt64)`. Please let me know if it doesn't. 
3. I'll comment on your suggestion in different points:
 - To answer your question about the PRG. No this FLIP did not provide a PRG implementation. It only provides a way to store previous block's sources, to make on-chain commit-reveal schemes possible (not possible without the proposed protocol update). Once this is unlocked, developers are able to implement any PRG they need, and can use the random sources as seeds for their PRGs. On this FLIP, I didn't want to tie developers to a single PRG instance. The FLIP implementation can propose one PRG implementation of course, but I thought it should be possible for developers to freely seed other PRGs of their choice. 
 - To continue with my argument above, I think including the PRG into the `SecureRandom` resource does not allow users to choose their PRG freely. Making `RandomBeaconHistory` private may also restrict the ways to use the random sources. The FLIP aims to provide the missing tools for developers (past random seeds), and allow them to implement the rest of the on-chain logic. 
 - `SecureRandom`: the main issue this FLIP wants to solve is the random output being revertible. It seems to me that the methods of `SecureRandom` (considered on their own) are still revertible, as in randoms can always be post-selected). My understanding is that the full commit-reveal design pattern is what makes randomness secure, and not a single resource. Specifically, the financial commitment step in a contract design, along with committing to a certain random (not known at the time of committing), is what makes post-selecting obsolete. A non-trusted user can always revert the transaction result, but they can't revert the financial commitment. It seems to me that the `SecureRandom` resource only commits to a certain random. I hope I understood your suggestion correctly and didn't miss an important point. 
 
 I'm curious to know what other Cadence developers think of Dete's suggestion and my answer (specifially #3) 
 cc @janezpodhostnik @sisyphusSmiling

protocol/20230728-commit-reveal.md
```diff
@@ -0,0 +1,138 @@
...
+	let sor = RandomBeaconHistory.SourceOfRandomnessAtBlock(atBlock)
+
+	// instantiate a PRG object using external `createPRG` that takes a `seed` 
+	// and `salt` and returns a pseudo-random-generator object.
+	let prg = PRG.createPRG(sor, salt)
+
+	// derive a 64-bit random using the object `prg`
+	let rand = prg.Uint64()
+	return UInt8(rand & 1)
+}
```
turbolent, line 137:
```suggestion
fun randomCoin(atBlockHeight blockHeight: UInt64, salt: UInt64): UInt8 {
	// query the Random Beacon history core-contract.
	// if `blockHeight` is the current block height, `sourceOfRandomness` errors.
	let sourceOfRandomness = RandomBeaconHistory.sourceOfRandomness(atBlockHeight: blockHeight)

	// instantiate a PRG object using external `createPRG` that takes a `seed` 
	// and `salt` and returns a pseudo-random-generator object.
	let prg = PRG.createPRG(sourceOfRandomness, salt)

	// derive a 64-bit random using the object `prg`
	let rand = prg.Uint64()
	return UInt8(rand & 1)
}
```

turbolent (commented):
Cadence pretty much follows [Swift's API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines) so
- Agreed on [avoiding abbreviations](https://www.swift.org/documentation/api-design-guidelines/#use-terminology-well) (`SoR` -> `SourceOfRandomness` / `RandomBeaconHistory`)
- [Function should be `lowerCamelCase`](https://www.swift.org/documentation/api-design-guidelines/#general-conventions) (`SourceOfRandomness` -> `sourceOfRandomness`)
- [Add an argument label to clarify the parameter](https://www.swift.org/documentation/api-design-guidelines/#promote-clear-usage). For the source of randomness function, it is not a _block_ that is passed, but a _block height_, e.g. `sourceOfRandomnessAtBlock(UInt64)` -> e.g.`sourceOfRandomness(atBlockHeight: UInt64)`

bluesign:
Can't we just store this on protocol state ? in block especially, Cadence has access to `getBlock`

AlexHentschel:
The following is subtle, but an important detail that hopefully addresses your question @bluesign:
- Nodes are not required to have block history beyond the start of the current epoch. While we could change this in theory, in practise this is non-negotiable (intractable engineering effort and undermining core value propositions of Flow for node operators). Implications:
    - The protocol currently does not guarantee availability of past Source of Randomness [SoR] (consider an EN joining and executing the first block of an Epoch). Therefore, transactions cannot assume that _past_ SoRs can be made available via the FVM.

- Consider that we are running cadence code on the state of block X. If cadence can interact with the state of block X, the QC certifying block X is guaranteed to be known.

  However, the commit and reveal scheme generally interacts with commitments that were stored in ancestors of blocks X! For example:
    - in the state of block X, there is a commitment to run some gambling code.
    - The commitment was made by a transaction in bock Y = X - 100.
    - The gambling contract specifies that when making a commitment in block Y, the SoR of the future block Z = Y+1 is to be used.
    - Meaning: the code is running on the state of block X. While executing, the cadence code needs access to the SoR for block X - 99. While we guarantee the knowledge of the QC for X, we do not guarantee knowledge of QCs for X's ancestor (including X-99).

In other words, we need a solution to store a long history of SoRs for each block. The execution state is the best place to store this history in my opinion, because: (i) persists across sporks, (ii) it is already fork aware, meaning we can store a different SoR history for each fork out of the box.

tarakby:
Thanks @AlexHentschel. 
@bluesign, Alex's answer is briefly explained in [this FLIP section](https://github.com/onflow/flips/blob/tarak/commit-reveal/protocol/20230728-commit-reveal.md#design-proposal) under notes (the first one). Nodes are not guaranteed to keep history of past block's data other than the current epoch, and hence the use of execution state (via a contract).

Regarding Cadence's `getBlock`, the block returned has [minimal information](https://developers.flow.com/cadence/language/environment-information#block-information), but not the consensus Qorum Certificate (which contains the source of randomness).

bluesign:
thanks @AlexHentschel this is pretty good argument, I didn't know that we couldn't get previous spork block information within Cadence. 

@tarakby yeah I was thinking exposing SoR there at first ( or just cadence exposing SecureRandom struct ) but I missed the part we don't have access to past block.

tarakby:
> I still think that exposing `PRG.createPRG(sourceOfRandomness, salt)` is little dangerous, instead exposing some `SecureRandom` cadence type would be much easier for developers.

I agree `PRG.createPRG(sourceOfRandomness, salt)` still requires developers to be careful and know what they are doing. 
IMO, the main added value of this FLIP is:
- without this FLIP, even a careful developer cannot work around revertible randomness by a non-trusted user.
- with this FLIP implemented, a careful developer is able to develop secure randomness.

In the current form, the FLIP doesn't prevent developers from making mistakes unfortunately, but it makes secure randomness implementable (which is already a good progress).
If we can update this FLIP to also prevent misuses, that would be even better! suggestions are welcome!  Dete's suggested a `SecureRandom` resource above, but randomness seems revertible to me (or I completely missed the idea).

bluesign:
>  Dete's suggested a SecureRandom resource above, but randomness seems revertible to me (or I completely missed the idea).

It is creating the resource, but value you can get on next block. ( so technically you cannot know the result yet in the same transaction )

sisyphusSmiling (commented):
Great discussion here, following up on some points made above:

> sourceOfRandomness(atBlockHeight blockHeight: UInt64), so a call-site looks like sourceOfRandomness(atBlockHeight: 1)

Agreed with @turbolent on including the named parameter for clarity. I also think `RandomBeaconHistory` is a well-named.

> Once this is unlocked, developers are able to implement any PRG they need, and can use the random sources as seeds for their PRGs. On this FLIP, I didn't want to tie developers to a single PRG instance.

This is a valuable feature here. The modularity afforded is really great to have, but I also think providing a sample PRG implementation(s) with the feature rollout is equally valuable. As mentioned, ultimately the quality of the application-level randomness will be contingent on PRG implementation, and most developers will want a set of plug & play options.

>  I think including the PRG into the SecureRandom resource does not allow users to choose their PRG freely. Making RandomBeaconHistory private may also restrict the ways to use the random sources.

I see what you're saying here @tarakby. It's not clear to me why we would want to obfuscate or limit access to `RandomBeaconHistory`, but I can see the advantage of developer-friendly methods to call for random values of different types. That's a pattern we can adopt into the initial implementation of the sample PRG and perhaps even into a util contract that converts byte arrays to alternative values as @dete implemented in `SecureRandom`.

Great FLIP! Just a small question about the FVM method below.

protocol/20230728-commit-reveal.md
```diff
@@ -0,0 +1,138 @@
...
+
+Adding a safe pattern to reveal randomness without the transaction results being reverted unlocks application relying on randomness (for instance the coin toss contract described above). Not providing such pattern pushes developers to rely on risky solutions with possible security flaws.
+
+## Design Proposal
+
+The proposed design is a commit-reveal pattern. 
+
+The solution requires infrastructure changes to provide new data to the transaction execution environment:
+
+ 1. a new FVM function that exposes the current block's SoR (more precisely a derived value from the protocol `SoR_A`) to Cadence runtime. Note that `unsafeRandom` only exposes randoms derived from SoR through a pseudo-random generator (PRG) but not the SoR itself.
```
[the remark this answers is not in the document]
tarakby: I think that's what is done in this [PR](https://github.com/onflow/flow-go/pull/4582): FVM exposes the source of randomness needed by the contract, and it is called in the service transaction at the end of the block.
Let me know if there is a conflict with what the FLIP describes.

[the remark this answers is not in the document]
sisyphusSmiling: Ah right, my mistake. Found the method in the [system chunk transaction](https://github.com/onflow/flow-go/blob/1231763cb306d6d81cdf7c922c5761097d1dcfb9/fvm/blueprints/scripts/systemChunkTransactionTemplate.cdc#L19)

janezpodhostnik:
> I think including the PRG into the SecureRandom resource does not allow users to choose their PRG freely. Making RandomBeaconHistory private may also restrict the ways to use the random sources.

I agree. I thing making `RandomBeaconHistory` as concise and minimal as possible would be the best as it can then serve as a reusable building block for the community to build their own tools on top of, if they are not sattisfied with the default PRG that will be provided. I also don't see an issue with making it public.

tarakby:
Thank you for all the comments! 

The FLIP is taking the following direction so far:

- It is about protocol changes (FVM update - system transaction update) that make a commit-reveal scheme for safe randomness possible, by recording all the random beacon history.
- The protocol data history above is stored in the execution state using a new `RandomBeaconHistory` contract. The contract is public and minimal, offering simple queries of random sources per block height. 
- This FLIP is not tight to a precise Cadence implementations of safe randomness. A simple example is given, but developers are free to use the history contract as a building block for different designs. Specific Cadence designs can be presented in future FLIPs if needed (the current proposal is a protocol FLIP rather than a Cadence FLIP).
- A Cadence implementation of one PRG instance (algorithm and interface) will be provided as part of this FLIP implementation, but is not tight to the history contract. The history contract can be used with other PRGs or other forms of processing.

Please feel free to add further comments or concerns about the above.

tarakby:
There seems to be no concerns about the current FLIP state. I will go ahead and update the status to `approved` unless there are new objections.

PR #179, opened 2023-08-30
FLIP 179: Staged Contract Updates

sisyphusSmiling:
FLIP Issue: #203 

This proposal outlines a mechanism for automated and efficient contract updates to take effect at or beyond a specified
block height with the goal of minimizing recovery time following breaking improvements. Included in this FLIP is a
design enabling contract developers to pre-define a sequence of contract updates across an arbitrary number of contracts
and accounts, and either execute these updates themselves or delegate deployment authority to some trusted party.


bjartek:
@JeffreyDoyle this is very interesting for flix.

application/20230809-staged-contract-updates.md
```diff
@@ -0,0 +1,463 @@
+---
+status: proposed
+flip: [179](https://github.com/onflow/flips/pull/179)
+authors: Giovanni Sanchez (giovanni.sanchez@dapperlabs.com)
+sponsor: Giovanni Sanchez (giovanni.sanchez@dapperlabs.com)
+updated: 08-31-2023
```
turbolent, line 6: https://xkcd.com/1179/

```suggestion
updated: 2023-08-31
```

sideninja:
Nice idea! 👍  
The part that worries me a bit (but is not really improving or making it worse to the current state) is what happens when one contract update fails, but maybe there can be a CLI tool supporting this and it can leverage mainnet/testnet emulator simulation to dry-run the updates. This is however out of the FLIP scope but I feel related to its integration.

application/20230809-staged-contract-updates.md
```diff
@@ -0,0 +1,463 @@
...
+}
+```
+
+With respect to the topic of this FLIP, the `Updater` can check `DeploymentResult.success` and emit any failed updates
+within the executed stage as well as preserve the failure status within its fields. Direct follow up and intervention by
+the `Updater` owner would then be to correct the issue and manually execute the contract updates.
+
+### Considerations
+
+#### Self-Defined Deployments
```
franklywatson, line 328: Supposing that 4000 contracts are queued for updates, and most of them depend on the latest FT contract update version. So we expect that within the context of one account and its individual contracts, the update sequence shall be specified by the account owner. 

However, how will be certain that the global update sequence will ensure that the FT contract update has preceded all the many contracts which depend on it globally? What about if someone depends on the Flovatar contract, but their contract update happen to be scheduled before the Flovatar contract gets updated?

application/20230809-staged-contract-updates.md
```diff
@@ -0,0 +1,463 @@
...
+updated contracts, then simulate the update execution locally. Guides and best practices on this front should follow if
+this FLIP is approved and implemented.
+
+Regarding 3/, a helpful tool here would be an easy to use dashboard where the `Updater` owner can authenticate and view
+both the current status of their `Updater` resource and any related events and transaction details where their updates
+were executed. This would make it easy for developers to identify issues with their update deployments and
+minimize follow up time.
+
+These gaps are building opportunities beyond the scope of this design, and contributions on these fronts are very
+welcome.
```
franklywatson, line 378: Are we saying we would want others to build this? Seems tenuous?

Without at least some of the above tooling seems like a pretty risky scenario?

franklywatson (approved):

application/20230809-staged-contract-updates.md
joshuahannan: We're planning on updating all the core contracts as part of the spork separate from this construct, so we don't need to worry about those. As for contracts like Flowvatar, I think we would probably just do multiple passes of this updater to catch any that might have happened in the wrong order

joshuahannan (approved):
Really well designed and written! I hope we can have something like this

application/20230809-staged-contract-updates.md
sisyphusSmiling: My hope is that if implemented, there would be some integration with @onflow/flow-develop-experience-tools with things like @sideninja mentioned above.

As proposed, simulated configuration and updates will be possible with existing/planned dev tooling via Cadence 1.0 local Mainnet mirroring AFAIK. The real addition here would be 3/ as a nice to have IMHO, and, if widely adopted, could be a grant candidate that I don't see taking too long to implement - likely a simple FE + indexer with event views filtered to the authenticated `Updater`.

sisyphusSmiling:
> maybe there can be a CLI tool supporting this and it can leverage mainnet/testnet emulator simulation to dry-run the updates

Love this idea @sideninja! That would be really helpful to prevent failed updates.

bjartek (commented):
There is a stable cadence emulator now so the suggested code snippets and repo should be updated to use stable-cadence clmpatible syntax. In particilar the entitlements on Account permissions.

bjartek:
I would slso like to add that in order to have a stable migration path to stable-cadence we need a way to test this process. I belive @joshuahannan created an issue about adding an invalid contract to emulator, but i could not find it now.

joshuahannan:
@bjartek https://github.com/onflow/flow-emulator/issues/468

franklywatson:
> > maybe there can be a CLI tool supporting this and it can leverage mainnet/testnet emulator simulation to dry-run the updates
> 
> Love this idea @sideninja! That would be really helpful to prevent failed updates.

cc @m-Peter

sisyphusSmiling:
Thanks for the callouts @bjartek! I've got a draft PR here: https://github.com/sisyphusSmiling/contract-updater/pull/6

It's currently in draft bc there's a [bug](https://github.com/onflow/flow-cli/issues/1208) in the CLI pre-release related to passing transaction args that prevents running through manual testing.

If we're leaning toward approving this mechanism, though, I can start working on a test suite for pre- and post-Cadence 1.0 since we'll need both versions

sisyphusSmiling:
Closing the comms loop here to point to the [updated Cadence 1.0 migration plan](https://forum.flow.com/t/update-on-cadence-1-0-upgrade-plan/5597). The contract detailed in this FLIP may still be useful for the purposes of coordinated contract upgrades whether it's Flow-managed or otherwise. Deployment of this contract is deprioritized in favor of a contract supporting network migration and will be revisited post-Cadence 1.0

See the latest changes to the staging contract interface here - https://github.com/onflow/contract-updater/pull/14

PR #195, opened 2023-09-08
[Networking] FLIP: Message Forensic (MF) System

yhassanzadeh13:
## Summary
This FLIP discusses and compares two potential solutions for the Message Forensic (MF) system in the Flow protocol — a system that identifies and attributes protocol violations to the original malicious sender. The two solutions under consideration are: (1) GossipSub Message Forensic (GMF), and (2) Enforced Flow-level Signing Policy For All Messages. We delve into both, listing their pros and cons, to determine which would be more feasible given the considerations of ease of implementation, performance efficiency, and security guarantees.

Our analysis finds the "Enforced Flow-level Signing Policy For All Messages" to be the more promising option, offering a generalized solution that doesn’t hinge on the protocol utilized to send the message, steering clear of the complexities tied to maintaining GossipSub envelopes and dodging the necessity of duplicating GossipSub router’s signature verification procedure at the engine level. Furthermore, it meshes well with the Flow protocol’s existing state.

## Review Guide
This FLIP is presented as a Pull Request (PR) in the `flow-go` repository. We welcome reviewers to express their opinions and share feedback directly on the PR page, aiming for a structured and productive discussion. To aid this, please adhere to one of the following response frameworks:
1. I favor the "Enforced Flow-level Signing Policy For All Messages" and here are my thoughts: 
2. I support the "GossipSub Message Forensic (GMF)" approach, articulating my views as follows.
3. I find both propositions unsatisfactory, elucidating my stance with.

### DIFF
--- a/protocol/20230908-message-forensic.md
+++ b/protocol/20230908-message-forensic.md
@@ -1,0 +1,269 @@
+| Status        | Proposed                                            |
+:-------------- |:----------------------------------------------------|
+| **FLIP #**    | [4697](https://github.com/onflow/flow-go/pull/4697) |
+| **Author(s)** | Yahya Hassanzadeh (yahya@dapperlabs.com)            |
+| **Sponsor**   | Yahya Hassanzadeh (yahya@dapperlabs.com)            |
+| **Updated**   | 8 September 2023                                    |
+
+# Message Forensic (MF) System
+
+## Summary
+This FLIP discusses and compares two potential solutions for the Message Forensic (MF) system in the Flow protocol 
+— a system that enables Flow protocol to attribute protocol violations to the original malicious sender. 
+The two solutions under consideration are: (1) GossipSub Message Forensic (GMF), and (2) Enforced Flow-level Signing Policy For All Messages. 
+We delve into both, listing their pros and cons, to determine which would be more feasible given the considerations of ease of implementation, 
+performance efficiency, and security guarantees.
+
+Our analysis finds the "Enforced Flow-level Signing Policy For All Messages" to be the more promising option, 
+offering a generalized solution that doesn’t hinge on a specific external protocol to send the message (e.g., GossipSub),
+steering clear of the complexities tied to maintaining GossipSub envelopes and dodging the necessity of duplicating GossipSub 
+router’s signature verification procedure at the engine level. Furthermore, it meshes well with the Flow protocol’s existing state.
+
+## Review Guide
+This FLIP is presented as a Pull Request (PR) in the `flow-go` repository. We welcome reviewers to express their opinions and share feedback directly on the PR page, aiming for a structured and productive discussion. To aid this, please adhere to one of the following response frameworks:
+1. I favor the "Enforced Flow-level Signing Policy For All Messages" and here are my thoughts:
+2. I support the "GossipSub Message Forensic (GMF)" approach, articulating my views as follows.
+3. I find both propositions unsatisfactory, elucidating my stance with.
+
+## Problem Overview
+Within the Flow protocol, nodes converse through a networking layer, a medium which undertakes the dispatch and receipt of messages among nodes
+via different communication methods: unicast, multicast, and publish. Unicast messages are transmitted to a single recipient over a direct connection.
+The multicast and publish, on the other hand, utilize a pub-sub network constructed with the LibP2P GossipSub protocol for message 
+dissemination. However, this protocol encounters challenges in message attribution, particularly in determining the 
+initial sender of a message, which is critical for identifying protocol violations and penalizing the malicious initiators. 
+Below, we break down the issues seen in different communication strategies and their implications.
+
+### Single Recipient (Unicast) Message Attribution
+Unicast messages have an implicit attribution to the sender, i.e., they are attributed to the remote node of the connection
+on which the message is received. Albeit, the situation differs when the unicast message carries the explicit signature of the
+sender or other attribution data (e.g., the signature of the original sender in case of a forwarded unicast).
+In such cases, the message is attributed to the sender that is specified in the attribution data. Nevertheless, the current state of
+Flow protocol does not enforce the attribution data to be present in the unicast messages. 
+A prime example is the [`ChunkDataResponse` messages](https://github.com/onflow/flow-go/blob/master/engine/execution/provider/engine.go#L326-L337) sent by the Execution Nodes to the Verification Nodes over a unicast. 
+Hence a node in Flow protocol cannot prove that a received unicast message is sent by a specific node, unless the message carries the explicit attribution data.
+
+### Group and All-Node (Multicast and Publish) Message Attribution
+When dealing with multicast or publish messages, the GossipSub router initially [signs it](https://github.com/libp2p/go-libp2p-pubsub/blob/master/sign.go#L109-L134) with a local node's networking key 
+before releasing it into the pub-sub network. The routers are structured to verify message signatures before forwarding them or making 
+them accessible to the application layer (i.e., Flow protocol engines). This method does indeed block invalid messages from progressing through the network, 
+also [penalizing](https://github.com/onflow/flow-go/blob/master/insecure/integration/functional/test/gossipsub/scoring/scoring_test.go#L26-L188) the immediate sender of an invalid message through a built-in scoring mechanism that influences network connections and message forwarding decisions.
+Yet, a significant loophole remains: the elimination of authentication data (like signatures) by the GossipSub router during message delivery to the application layer. 
+This eradication obstructs the Flow protocol's ability to trace a message back to its original sender in cases of protocol violations that happen at a level above the GossipSub layer.
+
+
+## Proposal-1: GossipSub Message Forensic (GMF)
+As the first option to (partially) mitigate the gap in the network to hold the malicious senders accountable,
+we introduce the GossipSub Message Forensic (GMF) mechanism designed to integrate GossipSub authentication data seamlessly with the 
+Flow protocol engines. The principal aim is to enhance message authenticity verification through multicast and publish,
+focusing on origin identification and event association. Here, we elucidate the method, its interface, and delve into the operational specifics, 
+weighing its pros and cons against potential alternatives.
+In this proposal, we propose a mechanism to share the GossipSub authentication data with the Flow protocol. We call this mechanism
+GossipSub Message Forensic (GMF). The idea is to add a new interface method to the [`EngineRegistry` interface](https://github.com/onflow/flow-go/blob/master/network/network.go#L37-L56). 
+The `EngineRegistry` was previously known as the `Network` interface, and is exposed to individual Flow protocol engines who are interested in receiving the
+messages from the networking layer including the GossipSub protocol. In this approach, the `EngineRegistry` interface is extended
+by two new methods: `GetGossipSubMessageForEvent` and `VerifyGossipSubMessage`. The `GetGossipSubMessageForEvent` method takes an origin 
+identifier as well as an event as input, and returns the GossipSub envelope of the message that is associated with the event and 
+is sent by the origin identifier.
+The `VerifyGossipSubMessage` method takes a origin identifier as well as a GossipSub message as input, and verifies the signature of
+the message against the networking public key of the origin id. 
+The method returns true if the signature is valid, and false otherwise. 
+The event is the scrapped message that is delivered to the application layer by the GossipSub router. The GossipSub envelope
+contains the signature of the message, as well as all the other GossipSub metadata that is needed to verify the signature. 
+```go
+
+type EngineRegistry interface {
+	// GetGossipSubMessageForEvent takes a Flow identifier as well as an event as input, and returns the GossipSub envelope of the
+	// message that is associated with the event and is sent by the origin identifier. The event is the scrapped message that is delivered to the application layer by
+	// the GossipSub router. The GossipSub envelope contains the signature of the message, as well as all the other GossipSub metadata
+	// that is needed to verify the signature.
+	// Args:
+	//   - originId: The Flow identifier of the node that originally sent the message.
+	//  - event: The scrapped message that is delivered to the application layer by the GossipSub router.
+	// Returns:
+	//  - message: The GossipSub envelope of the message that is associated with the event.
+	//  - error: An error if the message is not found.
+	GetGossipSubMessageForEvent(originId flow.Identifier, event interface{}) (*pb.Message, error)
+
+	// VerifyGossipSubMessage takes a Flow identifier as well as a GossipSub message as input, and verifies the signature of the
+	// message. The method returns true if the signature is valid, and false otherwise.
+	// Args:
+	//  - originId: The Flow identifier of the node that originally sent the message.
+	// - message: The GossipSub message that is associated with the event.
+	// Returns:
+	//  - bool: True if the signature is valid, and false otherwise.
+	//  - error: An error if the message is not found.
+	VerifyGossipSubMessage(originId flow.Identifier, message *pb.Message) (bool, error)
+
+	// Other methods are omitted for brevity
+	// ...
+}
+```
+
+### Implementation Complexities
+- This proposal requires the `LibP2PNode` implementation extract the individual [envelopes (i.e., `*pb.Message`)](https://github.com/libp2p/go-libp2p-pubsub/blob/master/pb/rpc.pb.go#L143-L153) from the incoming GossipSub RPCs. 
+  This is done using the [RPC inspector](https://github.com/libp2p/go-libp2p-pubsub/blob/master/pubsub.go#L1031-L1037) dependency injection. We need to implement a new inspector that extract the envelopes in a non-blocking approach. 
+  We should not block an incoming RPC to extract its messages, as it is interfering with the GossipSub router operation and can harm the message delivery rate of the router. 
+- The extracted envelopes are then required to stored persistently by their event id in a forensic component with proper ejection mechanism, 
+  which is exposed to the `GetGossipSubMessageForEvent` to query the envelope corresponding to a `(event, originId)` pair. 
+- This solution also requires replicating the signature verification logic of the GossipSub in `VerifyGossipSubMessage` so that it is accessible to the engines. 
+  We need to extend the signature verification mechanism to account for translation of `originId` from `flow.Identifier` to networking key and `peer.ID` (i.e., LibP2P level identifier).
+  As the engines are operating based on the `flow.Identifier`, while the GossipSub signatures are generated using the Networking Key of the node.
+
+### Advantages
+1. The GossipSub authentication data is shared with the Flow protocol, providing attribution for `Multicast` and `Publish` messages.
+2. The interface is easy to use for the engines, as it abstracts the complexity of translating the origin Flow identifier to the GossipSub peer id, and
+   verifying the signature of the message against the networking public key of the origin id.
+3. The implementation is not a breaking change and is backward compatible with the current state of the Flow protocol.
+
+### Disadvantages
+1. The GossipSub envelope is extractable through an RPC inspector dependency injection, which must be non-blocking and fast (a principle condition
+   imposed by GossipSub). This means that the GossipSub envelope extraction must be done in a separate goroutine _asynchronously_ to the message delivery to the
+   application layer. This entails that there can be a glitch between the message delivery to the application layer and the GossipSub envelope extraction. The potential glitch
+   means that the GossipSub envelope may not be available at the time of the message delivery to the application layer. Accounting for the glitch implies higher complexity in
+   the implementation of the Flow protocol engines (e.g., timeout queries, etc.). Moreover, the glitch may be exploited by the malicious nodes to perform timing
+   attacks on the Flow protocol engines and get away from detection. If an attacker can time the message delivery to the application layer in a way that the 
+   GossipSub envelope is not available at the time of the message delivery, then the attacker can send an invalid message and get away from detection, as when
+   there is no GossipSub envelope available, the Flow protocol engines cannot have forensic evidence to attribute the protocol-level violation to the malicious
+   sender.
+2. Maintaining the GossipSub envelopes poses several challenges:
+   - There must be an eviction policy to remove the GossipSub envelopes from the memory (or disk) after a certain period of time. Lack of an eviction policy may
+     lead to memory (or disk) exhaustion attacks.
+   - Existence of eviction policy may lead to data loss, as the GossipSub envelope may be evicted from the memory (or disk) before the Flow protocol engines
+     can use it to attribute a protocol-level violation to the malicious sender.
+   - Existence of eviction policy may also lead to attacks; if an attacker can time the message delivery to the application layer in a way that the GossipSub
+     envelope is evicted from the memory (or disk) before the Flow protocol engines can use it to attribute a protocol-level violation to the malicious sender,
+     then the attacker can send an invalid message and get away from attribution, as when there is no GossipSub envelope available, the Flow protocol engines cannot
+     have forensic evidence to attribute the protocol-level violation to the malicious sender.
+   - On a happy path when all nodes are honest, the GossipSub envelopes are not needed, and extracting and maintaining them poses a performance overhead.
+3. To construct the GossipSub-level signature on an event, this solution requires maintaining the entire [GossipSub message (i.e., envelope)](https://github.com/libp2p/go-libp2p-pubsub/blob/master/pb/rpc.pb.go#L143-L153) which is more than the event itself. 
+   This is because the GossipSub-level signature is generated on the entire message (i.e., envelope) and not just the event. Hence, this solution requires more memory and disk space.
+4. This solution requires replicating the signature verification logic of the GossipSub router at the engine level (i.e., `VerifyGossipSubMessage`). Changes to the GossipSub signature 
+    verification procedure may pose as breaking changes and be a blocker for upgrading and keeping up with GossipSub upgrades.
+5. The Flow node may receive the same message from multiple senders, and this solution requires all those message envelopes to be stored for forensic purposes. 
+6. This solution does not cover the unicast messages, as they are not sent through the GossipSub protocol.
+
+## Proposal-2: Enforced Flow-level Signing Policy For All Messages
+In this proposal, we introduce the policy to enforce a Flow-level signature for all messages using the staking key of the node. 
+The idea is to refactor the [`Conduit` interface](https://github.com/onflow/flow-go/blob/master/network/conduit.go#L26-L57), so that instead of taking an `interface{}` type event, it takes an `Envelope` type event. 
+The `Envelope` type event is a wrapper around the `interface{}` type event, and contains
+the staking key signature of the sender on the event. The `Envelope` type is defined as follows:
+```go
+type Envelope struct {
+    // The event that is wrapped by the envelope.
+    Event interface{}
+    // The Staking Key signature of the sender on the event.
+    Signature []byte
+}
+```
+
+Accordingly, the `Conduit` interface is refactored as follows:
+```go
+type Conduit interface {
+	Publish(envelope *Envelope, targetIDs ...flow.Identifier) error
+
+	Unicast(envelope *Envelope, targetID flow.Identifier) error
+
+	Multicast(event *Envelope, num uint, targetIDs ...flow.Identifier) error
+
+	// Other methods are omitted for brevity
+	// ...
+}
+```
+
+In this design, the engine on the sender side is required to sign the messages that are sent through the `Conduit` interface. To impose the safety mechanism, 
+the `Conduit` interface rejects any message that is not signed by the sender by returning an error. 
+On the other hand, the [`Engine` (aka `MessageProcessor`)](https://github.com/onflow/flow-go/blob/master/network/engine.go#L14-L60) interface is refactored to receive an `Envelope` type event instead of an `interface{}` type event. 
+The `Engine` interface is refactored as follows:
+```go
+// MessageProcessor represents a component which receives messages from the
+// networking layer. Since these messages come from other nodes, which may
+// be Byzantine, implementations must expect and handle arbitrary message inputs
+// (including invalid message types, malformed messages, etc.). Because of this,
+// node-internal messages should NEVER be submitted to a component using Process.
+type MessageProcessor interface {
+	// Process is exposed by engines to accept messages from the networking layer.
+	// Implementations of Process should be non-blocking. In general, Process should
+	// only queue the message internally by the engine for later async processing.
+	Process(channel channels.Channel, originID flow.Identifier, envelope *Envelope) error
+}
+```
+Prior to delivering the message to the engine (i.e., `MessageProcessor`), the networking layer will verify the signature of the `Envelope` against the staking key of the
+sender. If the signature is valid, the message will be delivered to the engine. Otherwise, the message will be rejected and reported to the Application Layer Spam Prevention (ALSP system) as an impersonation attempt, hence penalizing the
+node who sent it over the network (the remote node in unicast case, and the RPC sender in multicast and publish cases).
+In this way, when an engine's `Process` method is called, the engine can be sure that the message is signed by the sender. 
+Hence, the engine can attribute a protocol-level violation to the malicious sender that originally sent the message.
+
+### Implementation Complexities
+- This proposal requires refactoring the `Conduit` interface, which is used by all the Flow protocol engines to send messages to the networking layer. 
+  Currently, engines are passing arbitrary `interface{}` types as events to the conduits. This proposal requires refactoring the `Conduit` interface and 
+  engines to pass `Envelope` type events to the conduits. 
+- This proposal requires refactoring the `Engine` interface, which is used by all the Flow protocol engines to receive messages from the networking layer. 
+  Currently, engines are receiving arbitrary `interface{}` types as events from the networking layer. This proposal requires refactoring the `Engine` interface and 
+  engines to receive `Envelope` type events from the networking layer. This is a breaking change and is not backward compatible with the current state of the Flow protocol.
+- This proposal requires refactoring the networking layer of the Flow blockchain to verify the signature of the `Envelope` against the staking key of the sender prior to 
+  delivering the message to the engine.
+
+### Advantages
+1. This solution enables Flow protocol with attribution of a protocol-level violation to the malicious sender that originally sent the message.
+2. The implementation is simple; contrary to the GMF proposal, there is no need to extract and maintain the GossipSub envelopes, no extra asynchronous processing, storage, and memory management is needed.
+3. The signature and event are encapsulated together and are passed to the engine as a single `Envelope` object. Hence, the engine does not need to worry about the signature and event
+   being out of sync. The lifecycle of the signature and event are managed by the engine itself as an atomic entity, than dividing the responsibility between the engine and the networking layer.
+4. This solution not only covers the GossipSub messages, but also covers all the messages that are sent through the `Conduit` interface. Hence, it is a more general solution
+   that can be used to attribute a protocol-level violation to the malicious sender that originally sent the message, regardless of the protocol that is used to send the message.
+   For example, the current state of codebase does not enforce signature policy for the `ChunkDataPack` responses that Execution Nodes send to the Verification Nodes over unicast. Hence,
+   a verification node does not have any forensic evidence to attribute a protocol-level violation to the malicious Execution Node that originally sent the message.
+5. The entire authentication data is at the Flow protocol level (i.e., Staking Key) and is not dependent on the GossipSub protocol. Hence, the Flow protocol can attribute a
+   protocol-level violation to the malicious sender that originally sent the message, regardless of the protocol that is used to send the message, i.e., GossipSub in this case.
+6. `Envelope`-based message representation provides nested authentication, i.e., an envelope generated by sender A can be wrapped as an authenticated evidence by sender B, and so on.
+
+### Disadvantages
+1. The implementation is a breaking change and is not backward compatible with the current state of the Flow protocol. We change the `Conduit` interface, which is used by
+   all the Flow protocol engines to send messages to the networking layer. Hence, all the Flow protocol engines must be refactored to sign the messages that are sent through
+   the `Conduit` interface. 
+2. This approach adds a computation overhead to the Flow protocol engines, as the engines must sign all the messages that are sent through the `Conduit` interface, and on the receiving side, the 
+   networking layer must verify the signature of the message against the Staking Key of the sender. This overhead is not negligible, as the Flow protocol engines are the most performance critical components of the Flow blockchain. 
+   Hence, we must carefully evaluate the performance overhead of this approach. Moreover, with this approach, we are extending the size of data sent over the wire by piggybacking the signature.
+   Assuming that we are using ECDSA with `secp521r1` curve and SHA-512 for hashing, the signature size is ~100 bytes (in theory). Hence, we are adding ~100 bytes to the size of data sent over the wire.
+   Nevertheless, the performance overhead and the size of data sent over the wire may be an acceptable trade-off for the security guarantees that this approach provides.
+
+### Is the GossipSub Router Signature Still Necessary?
+Yes, maintaining the GossipSub router signature is essential, even while adopting Proposal-2. 
+The chief purpose of this signature is to authenticate messages at the GossipSub layer, serving as a robust mechanism to weed out invalid messages early in the transmission process,
+before reaching the application layer or being relayed to other nodes. Moreover, GossipSub needs its own signature to penalize the immediate sender of an invalid message through a built-in 
+scoring mechanism that influences network connections and message forwarding decisions.
+Drawing a parallel, it is akin to the ephemeral signatures employed in secure sessions like TLS and SSL. 
+These signatures work diligently to verify the authenticity of the session-level communications. 
+However, they are transient and do not function to attribute the message to its original sender after the session concludes. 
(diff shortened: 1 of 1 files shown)


protocol/20230908-message-forensic.md
```diff
@@ -0,0 +1,276 @@
+| Status        | Proposed                                            |
+:-------------- |:----------------------------------------------------|
+| **FLIP #**    | [4697](https://github.com/onflow/flow-go/pull/4697) |
```
gomisha, line 3:
```suggestion
| **FLIP #**    | [195](https://github.com/onflow/flips/pull/195) |
```

protocol/20230908-message-forensic.md
```diff
@@ -0,0 +1,276 @@
...
+We delve into both, listing their pros and cons, to determine which would be more feasible given the considerations of ease of implementation, 
+performance efficiency, and security guarantees.
+
+Our analysis finds the "Enforced Flow-level Signing Policy For All Messages" to be the more promising option, 
+offering a generalized solution that doesn’t hinge on a specific external protocol to send the message (e.g., GossipSub),
+steering clear of the complexities tied to maintaining GossipSub envelopes and dodging the necessity of duplicating GossipSub 
+router’s signature verification procedure at the engine level. Furthermore, it meshes well with the Flow protocol’s existing state.
+
+## Review Guide
+This FLIP is presented as a Pull Request (PR) in the `flow-go` repository. We welcome reviewers to express their opinions and share feedback directly on the PR page, aiming for a structured and productive discussion. To aid this, please adhere to one of the following response frameworks:
```
gomisha, line 23: The linked PR is closed. Should this be updated to ask for feedback on the current PR ([195](https://github.com/onflow/flips/pull/195)) instead?

protocol/20230908-message-forensic.md
```diff
@@ -0,0 +1,276 @@
...
+initial sender of a message, which is critical for identifying protocol violations and penalizing the malicious initiators. 
+Below, we break down the issues seen in different communication strategies and their implications.
+
+### Single Recipient (Unicast) Message Attribution
+Unicast messages have an implicit attribution to the sender, i.e., they are attributed to the remote node of the connection
+on which the message is received. Albeit, the situation differs when the unicast message carries the explicit signature of the
+sender or other attribution data (e.g., the signature of the original sender in case of a forwarded unicast).
+In such cases, the message is attributed to the sender that is specified in the attribution data. Nevertheless, the current state of
+Flow protocol does not enforce the attribution data to be present in the unicast messages. 
+A prime example is the [`ChunkDataResponse` messages](https://github.com/onflow/flow-go/blob/master/engine/execution/provider/engine.go#L326-L337) sent by the Execution Nodes to the Verification Nodes over a unicast.
```
gomisha, line 42:
```suggestion
A prime example is the [`ChunkDataResponse` messages](https://github.com/onflow/flow-go/blob/a85341d949c8a4e42e3d1d62bcf6fbf3d7f76bd9/engine/execution/provider/engine.go#L326-L337) sent by the Execution Nodes to the Verification Nodes over a unicast. 
```

protocol/20230908-message-forensic.md
```diff
@@ -0,0 +1,276 @@
...
+on which the message is received. Albeit, the situation differs when the unicast message carries the explicit signature of the
+sender or other attribution data (e.g., the signature of the original sender in case of a forwarded unicast).
+In such cases, the message is attributed to the sender that is specified in the attribution data. Nevertheless, the current state of
+Flow protocol does not enforce the attribution data to be present in the unicast messages. 
+A prime example is the [`ChunkDataResponse` messages](https://github.com/onflow/flow-go/blob/master/engine/execution/provider/engine.go#L326-L337) sent by the Execution Nodes to the Verification Nodes over a unicast. 
+Hence a node in Flow protocol cannot prove that a received unicast message is sent by a specific node, unless the message carries the explicit attribution data.
+
+### Group and All-Node (Multicast and Publish) Message Attribution
+When dealing with multicast or publish messages, the GossipSub router initially [signs it](https://github.com/libp2p/go-libp2p-pubsub/blob/master/sign.go#L109-L134) with a local node's networking key
```
gomisha, line 46: Suggestion: Use permalink instead of link to `master`, in case file gets updated.

protocol/20230908-message-forensic.md
```diff
@@ -0,0 +1,276 @@
...
+## Proposal-1: GossipSub Message Forensic (GMF)
+As the first option to (partially) mitigate the gap in the network to hold the malicious senders accountable,
+we introduce the GossipSub Message Forensic (GMF) mechanism designed to integrate GossipSub authentication data seamlessly with the 
+Flow protocol engines. The principal aim is to enhance message authenticity verification through multicast and publish,
+focusing on origin identification and event association. Here, we elucidate the method, its interface, and delve into the operational specifics, 
+weighing its pros and cons against potential alternatives.
+In this proposal, we propose a mechanism to share the GossipSub authentication data with the Flow protocol. We call this mechanism
+GossipSub Message Forensic (GMF). 
+
+At the current state of the code, the networking layer receives the [raw GossipSub envelop](https://github.com/onflow/flow-go/blob/master/network/p2p/p2pnet/internal/readSubscription.go#L48) from its underlying pubsub router,
```
gomisha, line 80:
```suggestion
At the current state of the code, the networking layer receives the [raw GossipSub envelope](https://github.com/onflow/flow-go/blob/master/network/p2p/p2pnet/internal/readSubscription.go#L48) from its underlying pubsub router, 
```

protocol/20230908-message-forensic.md
```diff
@@ -0,0 +1,276 @@
...
+- For the unicast cases, the networking layer reads the bytes from the stream, and decodes the `messages.Message` protobuf type from the bytes. Then, the `messages.Message.Payload`
+is decoded to an `interface{}` type event using CBOR, and is delivered to the Flow-protocol engines.
+- For the pubsub cases, the networking layer receives the `pb.Message` type from GossipSub, and decodes the `messages.Message` protobuf type from the `pb.Message.Data` bytes. 
+Then, the `messages.Message.Payload` is decoded to an `interface{}` type event using CBOR, and is delivered to the Flow-protocol engines.
+
+In the rest of this FLIP, we refer to `pb.Message` as the GossipSub envelop, and to the `message.Message` as the Flow message. Hence, in a nutshell:
+- For the unicast case, the event is wrapped in a Flow message type.
+- For the pubsub case, the event is wrapped in a Flow message type, which is then wrapped in a GossipSub envelope type.
+- The GossipSub envelop contains the signature of the sender over the entire envelope. The signature is generated using the networking key of the sender by the GossipSub router.
```
gomisha, line 69:
```suggestion
- The GossipSub envelope contains the signature of the sender over the entire envelope. The signature is generated using the networking key of the sender by the GossipSub router.
```

protocol/20230908-message-forensic.md
```diff
@@ -0,0 +1,276 @@
...
+	//  - error: An error if the message is not found.
+	VerifyGossipSubMessage(originId flow.Identifier, event interface{}, message *pb.Message) (bool, error)
+
+	// Other methods are omitted for brevity
+	// ...
+}
+```
+
+### Implementation Complexities
+- This proposal requires refactoring the message processing pipeline of the networking layer to pass along the GossipSub envelop together with the event to the Flow protocol engines. We need to
```
gomisha, line 126:
```suggestion
- This proposal requires refactoring the message processing pipeline of the networking layer to pass along the GossipSub envelope together with the event to the Flow protocol engines. We need to
```

protocol/20230908-message-forensic.md
```diff
@@ -0,0 +1,276 @@
...
+  also refactor the `Engine`'s `Process` method in a way that it can accept the GossipSub envelop together with the event. The implementation should account for proper encapsulation and abstraction of the interface,
```
gomisha, line 127:
```suggestion
  also refactor the `Engine`'s `Process` method in a way that it can accept the GossipSub envelope together with the event. The implementation should account for proper encapsulation and abstraction of the interface,
```

protocol/20230908-message-forensic.md
```diff
@@ -0,0 +1,276 @@
...
+
+At the current state of the code, the networking layer receives the [raw GossipSub envelop](https://github.com/onflow/flow-go/blob/master/network/p2p/p2pnet/internal/readSubscription.go#L48) from its underlying pubsub router,
+validates it, and passes the incorporated Flow message to for further decoding. The Flow message is then decoded to an `interface{}` type event,
+and is delivered to the Flow protocol engines. The idea is to pass along the GossipSub envelop to the Flow protocol engines together with their `interface{}` type event.
+In this way, the Flow protocol engines can have access to the GossipSub authentication data. The rough idea is depicted in the following figure. 
+Note that it is not a production-ready code, we need to architect the solution in an Object Oriented Manner with proper abstraction and encapsulation of the interface. In this figure,
+we intentionally pass along the envelope together with the event for sake of simplicity. In the actual implementation, we need to abstract the envelope and event in a single object 
+and pass it to the Flow protocol engines. We have to also account for the unicast case that does not have a GossipSub envelop.
```
gomisha, line 86:
```suggestion
and pass it to the Flow protocol engines. We have to also account for the unicast case that does not have a GossipSub envelope.
```

protocol/20230908-message-forensic.md
```diff
@@ -0,0 +1,276 @@
...
+
+	// Other methods are omitted for brevity
+	// ...
+}
+```
+
+### Implementation Complexities
+- This proposal requires refactoring the message processing pipeline of the networking layer to pass along the GossipSub envelop together with the event to the Flow protocol engines. We need to
+  also refactor the `Engine`'s `Process` method in a way that it can accept the GossipSub envelop together with the event. The implementation should account for proper encapsulation and abstraction of the interface, 
+  to also support the unicast case that does not have a GossipSub envelop.
```
gomisha, line 128:
```suggestion
  to also support the unicast case that does not have a GossipSub envelope.
```

protocol/20230908-message-forensic.md
```diff
@@ -0,0 +1,276 @@
...
+However, this stipulation does not apply universally across all data structures present in Flow, with `ChunkDataResponse` being a notable exception that does not demand the signature 
+of its originating execution node.
+
+Preserving the internal signature of a data structure holds considerable importance for future reference and validation processes.
+This retention is especially critical when introducing data structures to new nodes, which may either be lagging or freshly joining the network; 
+the internal signature by the creator facilitates the verification process, allowing the new node to establish trust in the data structure.
+On the other hand, the envelope-level signature characterizes the node wrapping the data structure, diverging from the creator’s signature found in the data structure itself.
+This approach substantially enhances forensic capabilities. Taking the instance of a scenario where a sender manipulates an execution receipt, compromising its integrity, 
+the system is designed to not just detect the violation via the executor's internal signature but to robustly attribute the breach to the malicious sender ensuring non-repudiation.
```
gomisha, line 269:
```suggestion
the system is designed to not just detect the violation via the executor's internal signature but to robustly attribute the breach to the malicious sender (via `Envelope` validation) ensuring non-repudiation.
```

protocol/20230908-message-forensic.md
```diff
@@ -0,0 +1,276 @@
...
+    verification procedure may pose as breaking changes and be a blocker for upgrading and keeping up with GossipSub upgrades.
+4. Verification logic is not straightforward, as it requires at least two steps:
+   - The first step is to ensure the event is wrapped in a GossipSub envelope. If not, the verification fails. For this we need to replicate the entire encoding path down to the GossipSub level
+      as wrapping the Flow message in the GossipSub envelope is done internally at the GossipSub and is not exposed to the Flow codebase. The replication may also cause another layer of coupling 
+      that causes breaking changes in the future upgrades of GossipSub.
+   - The second step is to verify the signature of the GossipSub envelope against the networking public key of the origin id. This requires translating the origin id from `flow.Identifier` 
+      to networking key and `peer.ID` (i.e., LibP2P level identifier), and replicating the signature verification logic of the GossipSub router, which is another layer of coupling that 
+      causes breaking changes in the future upgrades of GossipSub.
+
+## Proposal-2: Enforced Flow-level Signing Policy For All Messages
```
gomisha, line 153:
```suggestion
## Proposal-2: Flow-level Signing (FLS)
```

gomisha:
I favor the "Enforced Flow-level Signing Policy For All Messages" (I have suggested to rename this to FLS) and here are my thoughts:

- more generic solution vs GMF - I like that the solution can be used for generic forensics that aren't only for GossipSub messages 
- less overhead from having to carry around the GossipSub envelope by the engines as well as having to duplicate GossipSub verification logic in the engines

My biggest concern in adopting FLS is the backward compatibility issues this will cause as it will be a major breaking change. But from a design and maintainability perspective, I like this appraach over GMF.

protocol/20230908-message-forensic.md
yhassanzadeh13: https://github.com/onflow/flips/pull/195/commits/544c59f3de4fa83982b35dc40a8263f46f0e1ba3

protocol/20230908-message-forensic.md
yhassanzadeh13: https://github.com/onflow/flips/pull/195/commits/b44a9db5495e54191392bc9a862be2b8d931c4a0

bluesign:
I support the "GossipSub Message Forensic (GMF)" approach, articulating my views as follows: 

- Network layer is clearly separated from internal messaging logic. 
- By carrying GossipSub envelope without data we can prevent all the overhead. (without data GossipSub envelope should be pretty small ) 
- I think slashing / protocol violation is a very rare case, signing in gossipSub is pretty basic, duplicate GossipSub verification is only needed when protocol violation is raised and only required for the parts that are responsible for deciding on protocol violations. ( otherwise we have 2 signature verifications, one on network layer, one on application layer ) 

I raised before (on discord) about moving all communication on GossipSub ( unicast messages too ), but turned out there are some messages assumed to be sent on a direct 1:1 connection. I think solving that is also important. ( I don't have much information about current topology, but I think it maybe possible to have direct peering between components that require 1:1 connections )

yhassanzadeh13:
> I support the "GossipSub Message Forensic (GMF)" approach, articulating my views as follows:
> 
> * Network layer is clearly separated from internal messaging logic.
> * By carrying GossipSub envelope without data we can prevent all the overhead. (without data GossipSub envelope should be pretty small )
> * I think slashing / protocol violation is a very rare case, signing in gossipSub is pretty basic, duplicate GossipSub verification is only needed when protocol violation is raised and only required for the parts that are responsible for deciding on protocol violations. ( otherwise we have 2 signature verifications, one on network layer, one on application layer )
> 
> I raised before (on discord) about moving all communication on GossipSub ( unicast messages too ), but turned out there are some messages assumed to be sent on a direct 1:1 connection. I think solving that is also important. ( I don't have much information about current topology, but I think it maybe possible to have direct peering between components that require 1:1 connections ) Eventually I think objective is to allow anyone to join network (let's say any AN), making it over GossipSub security and scalability guarantees and would benefit in the long run.


@bluesign, it is imperative in GMF for an engine to retain the full integrity of the GossipSub envelope; the function hinges fundamentally on the wholeness of the data present. This is not merely a procedural stipulation but a requisite grounded in the necessity of verifying the signature, which encapsulates the entire envelope, thereby making it infeasible to relay the envelope devoid of its data contents.

⚠️ The perspective you depicted seems to bypass a critical phase of the verification process that involves:
1. Ascertaining the correlation between the event and the envelope, i.e., the event belongs to the envelope (requires data part of the envelope).
2. Authenticating the signature of the envelope, i.e., the envelope is attributable to the sender through its networking key (requires the entire envelope).

This is not about entrusting the networking layer blindly; it is about endorsing a methodologically sound practice that employs cryptographic primitives to furnish verifiable proofs, fostering a trust environment among Flow nodes that should not build trust but rather on cryptographically verifiable primitives. Otherwise, we may not even need a forensic mechanism in the first place.

By reducing the envelope to a fragmentary state, the engine's self-sufficient capacity for verification is _impeded_ and essentially _undermines_ the core objective of establishing irrefutable proofs. It is important to not overlook the cardinal principles that govern the GMF, which pivot on the complete and unaltered state of the data encompassed in the envelope. Thus, the GMF solution must consider the integral role of the intact envelope in sustaining the very foundation of the system we are discussing, ensuring the accurate verification of both the event’s association with the envelope and the envelope's signature.

bluesign:
What I was proposing is something like this:

Some struct like message.Message as the Flow message. ( enriched with seqno, topic. signature, key from pb.Message, and decoded event as interface{} ) passed to engine. So technically engine will have everything to prove (with little overhead). But as gossipSub is defending against impersonation etc, I don't think engine needs to do a signature check here. 

In case of conflict, this new struct is self sufficient to raise a claim. Then who is responsible checking for this violation can get the this struct  from the node, reconstruct pb.Message and  do the signature verification. if signature verification succeed, then it can punish the offender ( if not can punish the claimer )

yhassanzadeh13:
> What I was proposing is something like this:
> 
> Some struct like message.Message as the Flow message. ( enriched with seqno, topic. signature, key from pb.Message, and decoded event as interface{} ) passed to engine. So technically engine will have everything to prove (with little overhead). But as gossipSub is defending against impersonation etc, I don't think engine needs to do a signature check here.
> 
> In case of conflict, this new struct is self sufficient to raise a claim. Then who is responsible checking for this violation can get the this struct from the node, reconstruct pb.Message and do the signature verification. if signature verification succeed, then it can punish the offender ( if not can punish the claimer )

@bluesign doesn't this implicitly imply maintaining the `pb.Messages` by the node? As they are entirely required in order for the message to be verifiable.

peterargue (commented):
I have concerns with both options. 

I agree that that GossipSub specific option isn't universal enough to cover all of our usecases, but I worry about the overhead of adding an extra layer of signatures to every message. I wonder if there is a hybrid approach we could take that avoids generating and verifying 2 signatures for every/most messages, plus the additional network data.

At its core, the requirement is to collect cryptographic evidence that can be used by a 3rd party to prove that a protocol violation occurred, and that a specific node sent the message. The evidence itself is raw bytes of data making up the original message and the signature signed by the misbehaving node (and probably a block id/height to get the node's public keys from protocol state).

What if instead of adding a new signature to all messages, we use a usecase specific method, and encapsulate it in a shared interface to reduce complexity.

something like:
* Have a high level envelope interface with methods to get the underlying event and forensic data.
* Implementations hold references to the decoded event plus source data/signature needed for attribution
* Engines are updated to accept this new interface
* If a misbehavior is detected, the envelope is passed along to a slashing system

Then, there could be different implementations for each of the message types.
* Gossibsub would include the `pb.Message`.
* Unicast would include the new mechanism.
* Future optimizations could use the actual event itself as the evidence if it already contained a signature from the sender.
* The implementation could produce some consistent format that the slashing system used. I assume that would be something like `NodeID` plus 2 `[]byte`, one for the raw data and one for the signature.

I think this addresses most of the disadvantages for both options (except for being a breaking change), while also avoiding additional signatures.

AlexHentschel:
This is going to be a bit of a longer reply, sorry 😅. I think there is quite a few nuances and applications patterns that need to be considered.

### Central design goals
* We design for a scenario, where over longer periods of time, 99.9% messages are honest (or higher). In the ideal case, none of the code paths leading to slashing of a node are ever executed. Resource consumption and runtime impact should be optimized for the happy path scenario. We are willing to accept significant performance deterioration, in case of a attackers committing slashable protocol violations -  as long as on the order of some minutes, the network has slashed the offending nodes and ejected them. By virtue of being implemented, slashing will be enough of a deterrent. 
* For the overwhelming amount of messages, the additional information needed for message forensics is emphemeral. More concretely, for nearly all messages, their validity and protocol compliance will be confirmed within milliseconds. At this point, the additional message forensics information can be discarded. 

   Under normal operational scenarios (overwhelmingly dominant), there are zero to few unchecked messages in the engine's inbound queues. It is completely tractable to keep the additional information needed for message forensics temporarily in memory and let the garbage collector clean it up, once we know the message is honest and can drop the respective in-memory reference to the forensics data. My gut feeling is that this might cost us less than 200MB of extra ram per node, for almost all messages combined. There might be 1 or 2 exceptions (most prominently [ChunkDataResponse messages](https://github.com/onflow/flow-go/blob/master/engine/execution/provider/engine.go#L326-L337)), but there we can add special purpose optimizations. But for the large majority of messages, we can temporarily keep both the deserialized event in ram as well as `pb.Message` (that is including the event again in a serialized form). We can always optimize later. 

* It is generally highly desired to support the rotation of keys. We don't have to implement it now, but we also want to avoid designs making key rotation harder or impossible. 



### Thoughts on the metrics to rank designs by
* runtime impact
* reliability and strength of protection 
* complexity (implementation modularity, depth of required context to maintain code, probability of future extensions introducing subtle security vulnerabilities)

# Analyzing the proposals
## Regarding proposal-2: Flow-level Signing Policy (FSP)

* **runtime impact:**
   BLS signatures (staking key is BSL key) are computationally much more costly to generate verify an ECDSA signatures (networking key). 

   With the FSP proposal, we would incur the additional cost of an added BSL signature consistently on the happy path of the protocol. 

* **strength of protection:** 
    It is important to note that the FSP reduces the surface for attributable but not provable protocol violations, but does not eliminate it. Essentially, we are still (with a very small surface) violating Moxie Marlinspike's Cryptographic Doom Principle [[1]](https://moxie.org/2011/12/13/the-cryptographic-doom-principle.html) -- we are leaving some surface, where a node expends resources but cannot prove protocol violations. 

 I would argue that with the peer ranking system, we already have a a good foundational defense. But it leaves that gap of attributable but not provable protocol violations. FSP makes this gap smaller, but FSP doesn't close it entirely. Therefore, I questioning whether FSP is impactful enough compared to the required engineering time and the runtime cost -- especially since I believe there are ways to close this gap entirely. 

* **complexity:** I agree that that the engineering work is probably not too big of a lift. 
 
## Regarding proposal-1: GossipSub Message Forensic (GMF)

Generally I think this is the direction to go.  Lets look at our ranking criteria:
* **runtime impact:**
   If we keep `pb.Message` in memory for the short time we need it and then just garbage collect it, we would only expend a bit of ram. No latency or significant computational cost. 

* **strength of protection:** 

   No gap in security surface. Every attributable protocol violation is conceptually provable. Very strong security guarantee!.  

* **complexity:**  
   While I understand the concerns about [Implementation Complexities](https://github.com/onflow/flips/blob/544c59f3de4fa83982b35dc40a8263f46f0e1ba3/protocol/20230908-message-forensic.md#advantages) and [Disadvantages](https://github.com/onflow/flips/blob/544c59f3de4fa83982b35dc40a8263f46f0e1ba3/protocol/20230908-message-forensic.md#disadvantages), I think there are quite pragmatic solutions for those concerns. Overall, I am optimistic that we can achieve a more comprehensive solution without significantly extending the engineering cost compared to FSP.

### My thoughts on complexity

* In my mind, the responsibility to adjudicate slashing challenges is completely at the protocol layer. Slashing challenges are separate messages that are exchanged within the protocol. They are self-contained and should either reference other message, which the protocol has already embedded in blocks (e.g. execution receipts). Or the slashing challenge should contain the offending messages itself (the _entire_ message, incl. all envelopes). 

* I am of the opinion, that adjudicating slashing evidence is entirely out of the scope of the networking layer or the engines. They are focused on processing the _newest_ messages necessary for _extending_ the chain. The networking layer and the core protocol logic underlying the engines must be modular enough and provide the primitives allowing to validate slashing evidence outside of the happy path logic (engines). In other words, the networking layer and core protocol layer should expose very low-level functions that can be called by the adjudication logic when processing slashing challenges and the evidence contained therein.  

### Suggestions:
* I think the networking layer should continue what it is doing already: authenticates inbound messages before handing them to the engines (protocol layer). 

* separate `VerifyGossipSubMessage` from `EngineRegistry` . Reasoning:
   
   The adjudication logic should provide the networking key (maybe better [`Identity`](https://github.com/onflow/flow-go/blob/45e4885035530cc48e9913a4c295e2cf3e5bd156/model/flow/identity.go#L31)?) for each message it is requesting to be verified. This is necessary to allow key rotation (e.g. at epoch boundaries), but still allowing to verify authenticity of a message from the past epoch. 

    Therefore, the API of interfacing with the engines is very different from interfacing with the adjudication logic. Therefore, I think both should be separate interfaces. 


* Encapsulate all the auxiliary forensics information and sub cases for unicast vs  multicast into a `ForensicsContext`. The following sketch compares the current implementation to my flavour of the GMF proposal Message Forensics (7)


  The `ForensicsContext` could look something like this:
    ```golang
    type ForensicsContext interface {
    	Channel() channels.Channel
    	OriginID() flow.Identifier

    	// BinaryMessage returns the binary representation of the message from the perspective of the networking layer.
    	// This representation should contain the necessary information to:
    	//  * deserialize the Protocol-Layer message (e.g. for proving a protocol violation to another staked node)
    	//  * cryptographically confirm authenticity and integrity of the Protocol-Layer message
    	//    via the origin's networking key.
    	//  * The origin's public networking key is _not_ contained in the BinaryMessage
    	// In a nutshell, the returned value should present the necessary evidence to prove to a third party that the
    	// message was really sent by the origin. The protocol guarantees that the party inspecting this evidence knows
    	// the origin's public networking key. We cannot include the origin's public networking key here, 
    	// as this would _not_ be BFT.
    	BinaryMessage() []byte // TBD: suitable return type here. Not sure whether byte is the best.
    }
    ```

### My take on the [Disadvantages](https://github.com/onflow/flips/blob/544c59f3de4fa83982b35dc40a8263f46f0e1ba3/protocol/20230908-message-forensic.md#disadvantages)
> We need to extend the signature verification mechanism to account for translation of originId from `flow.Identifier` to networking key and `peer.ID` (i.e., LibP2P level identifier). As the engines are operating based on the flow.Identifier, while the GossipSub signatures are generated using the Networking Key of the node.

*  I think verifying networking signatures is not the responsibility of engines. The adjudication logic _might_ need to bridge `flow.Identifier` and `peer.ID`. Though, the adjudication logic presumably already knows which node is being accused of a protocol violation (that information would be part of the slashing challenge) and has looked up its networking key. 

   Not really sure I follow why this is outside of the implementation that we already have. Potentially our existing code needs to be refactored to be more modular, but it is hard for me to see why this would be super complex. 

> The first step is to ensure the event is wrapped in a GossipSub envelope. If not, the verification fails. For this we need to replicate the entire encoding path down to the GossipSub level as wrapping the Flow message in the GossipSub envelope is done internally at the GossipSub and is not exposed to the Flow codebase. The replication may also cause another layer of coupling that causes breaking changes in the future upgrades of GossipSub.

* I don't understand why we need to "replicate the entire encoding path down to the GossipSub level". The way I understood the description, we have the [raw GossipSub message](https://github.com/onflow/flow-go/blob/master/network/p2p/p2pnet/internal/readSubscription.go#L48), which contains all information. We can decode the protocol-layer event from that raw GossipSub message, can't we? 

   In a nutshell, this is all that Charlie needs to do when he want to adjudicate a slashing challenge raised by Alice (see my picture above for context) : 
    1. verify that `pb.Message` (contained in Alice's slashing challenge) is in fact originating from Edward. From what I understood, I think that is possible. 
    2.  from `pb.Message`, decode the event  and verify that it violates the protocol in accordance with Alice's complaint

jordanschalm:
I support the "GossipSub Message Forensic (GMF)" approach, with some of the extensions suggested by Peter & Alex. 

### Interface Changes
To accommodate a `ForensicsContext` accompanying messages, we need to modify the `MessageProcessor` interface. The FLIP proposes this interface, where the envelope is added as a separate parameter.
```
Process(channel channels.Channel, originID flow.Identifier, event interface{}, envelope *pb.Message) error
```
I agree with Peter's suggestion of instead including the event itself and forensic data within a single higher-level structure:
> Have a high level envelope interface with methods to get the underlying event and forensic data.
There are benefits to passing a `Message` or envelope interface type rather than an `any` type, for example:
- We can uniformly associate metric IDs, resource names, etc. to the `Message` type, rather than manually selecting the label at each callsite, which is much more susceptible to human error ([example](https://github.com/onflow/flow-go/blob/24cbe582986ad9e024df68cba7a9ffbf43899bbb/engine/common/synchronization/engine.go#L219))
- We can uniformly assign `Message`s identifiers to be able to trace their progress through different components. 
Since we need to change the interface for this proposal, we should change it in a way that opens up some of these options in the future without further API changes.
```go
Process(channel channels.Channel, originID flow.Identifier, message flow.Message) error

type Message interface {
    Event() any
    ForensicsCtx() ForensicsContext
    // ...
}
```

**Side Note:** As the `ForensicsContext` will include the origin ID, we could remove this parameter from `Process` as well.

### Adding some colour to the comparison of runtime impact
To explicitly state what Alex touched on in his comment: To enable linking messages to their corresponding network signature and envelope, we need to _retain a reference to already-allocated memory_ for (slightly) longer. We do not need to _allocate additional memory_ for each message. 

#### What is the memory impact of retaining `pb.Message` references for the complete processing duration of a message?
Here's some back-of-the-envelope calculations:

Average GC cycles on Mainnet averages around 0.5-2 per minute, depending on the node role. ([metrics link](https://dapperlabs.grafana.net/explore?orgId=1&panes=%7B%223kr%22:%7B%22datasource%22:%22rKhoxtiMz%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22rKhoxtiMz%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22avg%20by%20%28role%29%20%28rate%28go_gc_duration_seconds_count%7Bnetwork%3D%5C%22mainnet23%5C%22,job%3D%5C%22flow_app%5C%22%7D%5B2m%5D%29%29%20%2A%2060%22,%22legendFormat%22:%22__auto%22,%22range%22:true,%22instant%22:true%7D%5D,%22range%22:%7B%22from%22:%22now-7d%22,%22to%22:%22now%22%7D%7D%7D&schemaVersion=1))
image

Suppose it takes 100ms on average to process a message. Then the memory impact would be upper-bounded by the proportion of messages we receive within 100ms of a GC cycle. If we assume the larger 2 GCs/min, then about 0.3% of the memory associated with `pb.Message` allocations would be retained for an extra GC cycle as a result of the longer retention. Even if all the suggested 16GB RAM were used for `pb.Message`s, we would incur only a **maximum of 50MB** additional cost by retaining the message reference as suggested. In practice, this should be much much lower (**likely <10MB in practice**)

kc1116:
I support the "GossipSub Message Forensic (GMF)" approach, articulating my views as follows.
- I am afraid enforcing the message signing policy at the application layer will add `non-negligible` overhead to message processing.
- Although the flow signing policy will add support for signature verification of messages via Unicast the functionality will not be used effectively since we are not doing any "message forwarding" with Unicast. 
- Alex suggestion to keep a reference in memory to the original gossip sub-envelope solves the issue of additional overhead due to "passing the full original envelope" to the engine layer.

tarakby:
I do not support the "Enforced Flow-level Signing Policy For All Messages", I support the idea of "GossipSub Message Forensic (GMF)" but some changes may be necessary.

First, I wanted to clarify two concepts or cryptographic services, since they will be used twice in my comment:
- authentication: the ability to attribute a message to a certain party. For instance Alice is convinced that a message it received was sent by Bob.
- non-repudiation: the ability to prove to other parties the origin of a message. For instance, Alice is able to prove to Oscar that a message she received was sent by Bob.

These definitions may have social/legal aspects so we can assume for simplicity that we are not attributing messages to Bob as a party, but we are attributing messages to a party controlling a private key corresponding to some public key shared with the public. 
Non-repudiation is a stronger concept than authentication, and it prevents Bob from denying being the origin of some message. I believe @AlexHentschel has mentioned the same concepts and called them attributability (for authentication) and provability (for non-repudiation). 
As an example in cryptographic primitives, signature schemes offer both properties while message authentication code (MAC) between 2 parties only offers authentication.
The purpose of this FLIP is to provide protocol-level non-repudiation, since authentication seems to be already implemented on the network layer.

I do not support "Enforced Flow-level Signing Policy For All Messages" because:
1. It seems to me that it does not provide non-repudiation (I believe @AlexHentschel also pointed this out). If a message has a valid network authentication, but invalid protocol level signature, there is no way the protocol can attribute the message to the original sender, because the protocol does not recognize the network level authentication. I believe this is the issue we wanted to solve in the first place.
2. It is a redundant level of authentication IMO as @yhassanzadeh13 pointed out. The networking layer is already  authenticating messages, and some protocol messages are already being authenticated using the staking key (consensus votes for instance). We could disable the extra Flow-level signature for those "already-signed" messages but that may mess up with the engine modularity. Note that the protocol signatures are not always signing the message envelope (some signed messages are even omitted from the payload itself).
3. a less important point is about the signature scheme chosen. Protocol level authentication is currently using BLS. BLS is only relevant when multi-signature is needed (for instance aggregation and batch verification), but is not optimized for basic signatures (on my laptop, one BLS verification is 13x slower than ECDSA and 16.5x slower than EdDSA - considering our new fast BLS implementation). 

While I support "GossipSub Message Forensic (GMF)", I wanted to clarify a few points, in particular about our current implementation using libp2p:
1. I think it makes sense to differentiate the protocol specs from its implementation. We may want to base our specs on libp2p (since it is used in the current and only implementation of Flow nodes), but we should be able to describe the attribution data in a spec. The attribution data is part of a challenge that would eventually be posted in blocks and the protocol state, and should therefore be described regardless of libp2p or any other implementation. 
2. we are using lip2p2 unicast (for 1-1) and PubSub (for 1-many). I am going to only consider unicast messages (I didn't dig deep in libp2p's pubsub):
   1. libp2p isn't simply signing every network payload by the sender's networking private key and then verifying it on the receiving end by the sender public key. This is the complexity I believe @yhassanzadeh13 mentioned (also answering your question @AlexHentschel). The networking key isn't used to sign any network message, it only signs a second level key called a "static key", and therefore delegates authentication to the static key. This is a choice libp2p took to allow multiple signature schemes for the networking key (Flow doesn't need this choice). The static keys are then used in a Noise [handshake](https://github.com/libp2p/specs/blob/master/noise/README.md#libp2p-data-in-handshake-messages) involving other ephemeral keys to perform a key agreement. The symmetric keys are then used in an [authenticated encryption](https://github.com/libp2p/specs/blob/master/noise/README.md#encryption-and-io) (Flow doesn't need encryption here, but it is enabled by default in libp2p). The symmetric keys are what provide authentication, not the networking keys node share when staking. I have discussed this [long time ago](https://github.com/dapperlabs/flow-go/issues/5634#issuecomment-871750392) when we switched from SECIO to Noise (low level security protocols used by libp2p). Ok all this sounds complicated, but can we still make libp2p export all the data needed for attribution and trace it back to the networking key?
   2. the key agreement above is a Diffie-Hellman (DH) involving both static and ephemeral keys. Proving correctness of the shared key to third parties includes exporting the static private key of the node ⚠️ 
   3. the shared key, as it names says, is shared between both ends of the communication. Authentication is provided by authenticated encryption (AEAD). You can think of this as encryption mixed with a MAC (if you're not familiar with a MAC, it is the symmetric version of a signatures). As I mentioned in the beginning of my message, MAC does not provide non-repudiation. Evil Bob can always claim they never signed the payload (even though honest Alice knows he did), and that because the signing key is shared ⚠️ 
  
I didn't look at PubSub in details, it seems that the networking key (not libp2p static key) is [used to sign original payloads](https://github.com/libp2p/specs/blob/master/pubsub/README.md#message-signing) (which is great news) but we still need to confirm it.

If we want to implement GMF with libp2p we would need to make some changes to libp2p and its use of Noise for 1-1 (maybe through a fork). I will stop my reply here and we can get into ideas on how to update libp2p later.

protocol/20230908-message-forensic.md
```diff
@@ -0,0 +1,276 @@
...
+This FLIP is presented as a Pull Request (PR) in the `onflow/flips` repository. We welcome reviewers to express their opinions and share feedback directly on the PR page, aiming for a structured and productive discussion. To aid this, please adhere to one of the following response frameworks:
+1. I favor the "Flow-level Signing Policy (FSP)" and here are my thoughts:
+2. I support the "GossipSub Message Forensic (GMF)" approach, articulating my views as follows.
+3. I find both propositions unsatisfactory, elucidating my stance with.
+
+## Problem Overview
+Within the Flow protocol, nodes converse through a networking layer, a medium which undertakes the dispatch and receipt of messages among nodes
+via different communication methods: unicast, multicast, and publish. Unicast messages are transmitted to a single recipient over a direct connection.
+The multicast and publish, on the other hand, utilize a pub-sub network constructed with the LibP2P GossipSub protocol for message 
+dissemination. However, this protocol encounters challenges in message attribution, particularly in determining the
```
tarakby, line 32: The context here is great. It would be also clearer if we recall the Flow protocol state assumptions (for each node, the protocol state tracks 2 public keys, a staking and a networking key)

protocol/20230908-message-forensic.md
```diff
@@ -0,0 +1,276 @@
...
+   all the Flow protocol engines to send messages to the networking layer. Hence, all the Flow protocol engines must be refactored to sign the messages that are sent through
+   the `Conduit` interface. 
+2. This approach adds a computation overhead to the Flow protocol engines, as the engines must sign all the messages that are sent through the `Conduit` interface, and on the receiving side, the 
+   networking layer must verify the signature of the message against the Staking Key of the sender. This overhead is not negligible, as the Flow protocol engines are the most performance critical components of the Flow blockchain. 
+   Hence, we must carefully evaluate the performance overhead of this approach. Moreover, with this approach, we are extending the size of data sent over the wire by piggybacking the signature.
+   Assuming that we are using ECDSA with `secp521r1` curve and SHA-512 for hashing, the signature size is ~100 bytes (in theory). Hence, we are adding ~100 bytes to the size of data sent over the wire.
```
tarakby, line 233: This example is correct, but using `secp256r1` as an example gives a closer number to what we have in Flow (`secp521` provides higher security than that we require for Flow, while `sec256r1` is what the networking layer is using). 
We can also omit the hashing algorithm because it doesn't impact the signature length. 
```suggestion
   Assuming that we are using ECDSA with `secp256r1` curve, the signature size is 64 bytes. Hence, we are adding 64 bytes to the size of data sent over the wire.
```

tarakby (commented):
Nice write-up 👏🏼

AlexHentschel:
thanks @tarakby for your detailed comment. Lets try to use the established terminology of **authentication** and **non-repudiation** going forward. Thanks Tarak for clearly explaining the concepts.

KshitijChaudhary666:
Hi @yhassanzadeh13 - this FLIP is not reflected on FLIP project tracker. Did you follow the process outlined in https://github.com/onflow/flips? Specifically please remember to do the following without which the FLIP won't get visibility on the project tracker-

Create an [issue](https://github.com/onflow/flips/issues/new/choose) by using one of the FLIP issue templates based on the type of the FLIP - application, governance, cadence or protocol. The title of the issue should be the title of your FLIP, e.g., "Dynamic Inclusion fees". Submit the issue. Note the issue number that gets assigned. 
Then, Create your FLIP as a pull request to this repository ([onflow/flips](https://github.com/onflow/flips)). Use the issue number generated in step 2 as the FLIP number. And mention the FLIP issue by copying the GitHub URL or the issue in the comment section.

Thank you!

KshitijChaudhary666:
Hi @yhassanzadeh13 - this FLIP is not reflected on FLIP project tracker. Can you pls follow the process outlined in https://github.com/onflow/flips? Specifically the following without which the FLIP won't get visibility on the project tracker-

- Create an [issue](https://github.com/onflow/flips/issues/new/choose) by using one of the FLIP issue templates based on the type of the FLIP - application, governance, cadence or protocol
- The title of the issue should be the title of your FLIP, e.g., "Dynamic Inclusion fees". Submit the issue. Note the issue number that gets assigned
- Use the issue number generated in step 2 as the FLIP number. And mention the FLIP issue by copying the GitHub URL or the issue in the comment section
-

AlexHentschel:
@KshitijChaudhary666 Yahya has moved on to a new professional opportunity. 
This Flip is currently iceboxed (at least from the perspective of Flow foundation resourcing it - which does jot preclude community contributors from picking it up, though they would need to coordinate with us). In summary it is up to us now to do with this flip whatever we do with flips currently on ice (please let me know).

protocol/20230908-message-forensic.md
```diff
@@ -0,0 +1,276 @@
+| Status        | Proposed                                            |
+:-------------- |:----------------------------------------------------|
+| **FLIP #**    | [195](https://github.com/onflow/flips/pull/195) |
```
yhassanzadeh13, line 3:
```suggestion
| **FLIP #**    | [259](https://github.com/onflow/flips/issues/259) |
```
@kc1116

yhassanzadeh13: @KshitijChaudhary666 as @AlexHentschel mentioned, I have moved out of Flow, and this FLIP has been handed off to @kc1116. I followed your guidelines on creating an issue and updating the FLIP number. However, as I am not part of the Flow org, I cannot amend the PR.

vishalchangrani: thank you for making the change Yahya! I will commit the change

AlexHentschel (approved):

AlexHentschel (commented):
I am wondering what state we should be targeting for this PR. There is a lot of important and valuable discussion here in the PR, which might still lead to changes in the proposed FLIP.

KshitijChaudhary666 (approved):

vishalchangrani:
> I am wondering what state we should be targeting for this PR. There is a lot of important and valuable discussion here in the PR, which might still lead to changes in the proposed FLIP.

hey @AlexHentschel - we will not close the PR.

Just to be clear,
The issue to track this FLIP is: https://github.com/onflow/flips/issues/259
The FLIP ID is the same as the issue ID (as per the new process): 259.
This PR can remain open till there is agreement on the FLIP.

PR #197, opened 2023-09-13
FLIP 196: Restrict Capabilities Publish

bluesign:


turbolent (approved):
Nice, thank you for opening this FLIP Deniz! 👍

cadence/20230913-restrict-capabilities-publish.md
```diff
@@ -0,0 +1,46 @@
...
+flip: 196 
+authors: Deniz Mert Edincik (deniz@edincik.com)
+updated: 2023-09-13
+---
+
+# FLIP 196: Restrict capabilities.publish to account's own capabilities
+
+## Objective
+
+With current controller capabilities, it is possible to re-publish someone else's public capability; as this shift in behavior can cause problems, this FLIP strives to address this issue with minimal disruption.
```
turbolent, line 12:
```suggestion
With the new controller-based capability API, introduced in [FLIP 798](https://github.com/onflow/flips/blob/main/cadence/20220203-capability-controllers.md), it is possible to re-publish someone else's capability (with `Account.capabilities.publish`).

This was not possible in the previous/currently available linking-based capability API, the function `AuthAccount.link` does not allow targeting another account.

As this shift in behavior can cause problems, this FLIP strives to address this issue with minimal disruption.
```

cadence/20230913-restrict-capabilities-publish.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Motivation
+
+Before the introduction of new capabilities API, public capabilities were guaranteed to point to storage at the same address. However, with the introduction of this API, it is now possible to obtain another account's public capability and republish it as one's own, while still maintaining the link to the other account's storage.
```
turbolent, line 16:
```suggestion
Before the introduction of the new controller-based capabilities API, public capabilities were guaranteed to point to storage at the same address. However, with the introduction of the API, it is now possible to obtain another account's public capability and republish it as one's own, while still maintaining the link to the other account's storage. 
```

cadence/20230913-restrict-capabilities-publish.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Objective
+
+With current controller capabilities, it is possible to re-publish someone else's public capability; as this shift in behavior can cause problems, this FLIP strives to address this issue with minimal disruption.
+
+## Motivation
+
+Before the introduction of new capabilities API, public capabilities were guaranteed to point to storage at the same address. However, with the introduction of this API, it is now possible to obtain another account's public capability and republish it as one's own, while still maintaining the link to the other account's storage. 
+
+A lot of scenarios such as voting and gating using the proof of Non-Fungible Token (NFT) ownership, usually involve checking if an account owns a certain balance or resource by verifying the public path capability. Now, the responsibility of protection falls on the developers, which is an extra burden and holds the potential of introducing errors.
```
turbolent, line 18:
```suggestion
A lot of scenarios, such as voting and gating using the proof of Non-Fungible Token (NFT) ownership, usually involve checking if an account owns a certain balance or resource by verifying the public path capability. Now, the responsibility of protection falls on the developers, which is an extra burden and holds the potential of introducing bugs and security issues.
```

cadence/20230913-restrict-capabilities-publish.md
```diff
@@ -0,0 +1,46 @@
...
+
+The current suggested method of defence against this issue is always checking for the `address` of the capability or the `owner` of the resource after borrowing. Unfortunately, this is an error-prone approach that developers can easily forget.
```
turbolent, line 20:
```suggestion
The current suggested method of defence against this issue is always checking for the `address` of the capability or the `owner` of the resource after borrowing. 
Unfortunately, this is an error-prone approach that developers can easily forget.
In addition, the new `capabilities.borrow` convenience function gets the capability and borrows it in one call.
Developers will likely forget to still `get` and check the address of the capability, when needed. 
```

cadence/20230913-restrict-capabilities-publish.md
```diff
@@ -0,0 +1,46 @@
...
+
+## User Benefit
+
+This proposal aims to add certain restrictions in order to decrease developer burden.
```
turbolent, line 24:
```suggestion
This proposal aims to add a restriction, in order to decrease developer burden and decrease the risk of bugs and security issues. 
The restriction already exists in the currently available linking-based capability API, so this proposal simply proposes to keep the existing restriction, and prevent a regression in the API in terms of safety and usability.
```

SupunS:
What if someone else sends a capability to the inbox of my account, and once I claim it, I want to publish that? I guess that would not be possible with this change?
Not sure if that is even a valid use-case, I'm just asking out of curiosity.

turbolent:
@SupunS 

> What if someone else sends a capability to the inbox of my account, and once I claim it, I want to publish that? I guess that would not be possible with this change?

That would not be possible, but is also not possible with the linking-based capability API.

SupunS (approved):
Being restrictive to start with, is always a good idea!

turbolent (approved):
Accepted 🎉

PR #218, opened 2023-11-07
Update FLIP 204

jordanschalm:
- Add more detailed deployment plan, based on timing of Mainnet24 spork. 
- Add link to GitHub epic.

### DIFF
--- a/protocol/20231003-service-account-specified-epoch-switchover.md.md
+++ b/protocol/20231003-service-account-specified-epoch-switchover.md.md
@@ -131,6 +131,7 @@ pub fun getTargetEndTimeForEpoch(
 
 
 ### Implementation Plan
+See also [ZenHub epic](https://github.com/onflow/flow-go/issues/4948).
 
 #### Smart Contract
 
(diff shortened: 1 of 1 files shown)


joshuahannan (approved):
Since it looks like we are wanting to do the [epoch smart contract performance upgrades](https://github.com/onflow/flow-core-contracts/pull/379) sooner rather than later (maybe within the next 5 weeks, meaning canary upgrade would likely have to be in 3 weeks), do you think you'll have time to get this change into that upgrade, or should we just plan on doing it during a future upgrade?

jordanschalm:
> do you think you'll have time to get this change into that upgrade, or should we just plan on doing it during a future upgrade?

I should have time to implement the change. However, it might be tight to complete the full deployment plan in that time. (I'm not sure how long it takes to get all node operators to upgrade 😬 and this would be a breaking change, including for Observer nodes etc.) I'll aim to have the code ready and will ask around about how long it takes to get upgrades installed by all operators.

PR #224, opened 2023-11-22
Update FLIP-123 status

tarakby:
🥳

### DIFF
--- a/protocol/20230728-commit-reveal.md
+++ b/protocol/20230728-commit-reveal.md
@@ -1,5 +1,5 @@
 ---
-status: approved
+status: implemented
 flip: 123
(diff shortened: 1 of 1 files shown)


protocol/20230728-commit-reveal.md
```diff
@@ -1,5 +1,5 @@
 ---
-status: approved
+status: implemented
```
sisyphusSmiling, line 2: 🚀 Since the PRG is on Mainnet, can the status can be updated to `released`?

tarakby: good flag, I missed that there is the "released" status.

sisyphusSmiling (approved):

PR #225, opened 2023-11-27
EVM support

ramtinms:
## 👉 Announcement 👈 

Thank you for the insightful discussions on this FLIP. We feel there is an overwhelming support from the active developer community and the broader ecosystem. 

On **May 15th**, 2024 **10-11am PT (17-18:00 UTC)** we'll hold a joint **meeting** of the [Core Protocol Working Group](https://www.google.com/url?q=https://github.com/onflow/Flow-Working-Groups/tree/main/core_protocol_working_group&sa=D&source=calendar&ust=1715626269987743&usg=AOvVaw3xNADeHWNuPnhG2Dc4JdkJ) and the [Cadence Language and Execution Working Group](https://www.google.com/url?q=https://github.com/onflow/Flow-Working-Groups/tree/main/cadence_language_and_execution_working_group&sa=D&source=calendar&ust=1715626269987743&usg=AOvVaw0cAlXS0qgEpWxMydqQNwmY), where we would like to briefly go over it together and approve this FLIP unless there are any major concerns remaining. 

You can add the meeting to your calendar via [this link](https://calendar.google.com/calendar/event?action=TEMPLATE&tmeid=MmoyN2Q0bGx0Y3Q1ZnFib2dsNWIyNnM1bnRfMjAyNDA1MjNUMTcwMDAwWiBjXzQ3OTc4ZjVjZDlkYTYzNmNhZGM2Yjg0NzMxMDJiNTA5MmMxYTg2NWRkMDEwNTU4MzkzZWNiN2Y5ZmQwYzlhZDBAZw&tmsrc=c_47978f5cd9da636cadc6b8473102b5092c1a865dd010558393ecb7f9fd0c9ad0%40group.calendar.google.com&scp=ALL) or find it on our 📆 [Flow Events & Working Groups](https://bit.ly/flow-events-calendar) calendar along with all the other working group meetings. Everyone is welcome 💚


protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,268 @@
...
+
+**What about other fungible and non-tokens?**
+
+The main reason we call these account bridged accounts is their design makes it easy to build bridges. A Cadence smart contract can control a bridged account which means it could facilitate transactional operation across two environments. For example, this smart contract can accept a Fungible on the cadence side and remit the equivalence on the EVM side to a target address in a single Flow transaction. More about this would come in follow-up Flips. 
+
+#### Safety and Reproducibility of the EVM state
+
+EVM state is not accessible by Cadence outside of the defined interfaces above, and Cadence state is also protected against raw access from the EVM. 
+
+At the start, the EVM state is empty (empty root hash), and all the state is stored under a FLOW account that is not controlled by anyone (network-owned). Any interaction with this environment emits a transactionExecuted event (direct calls for bridged accounts use `255` as transaction type).
```
franklywatson, line 180:
```suggestion
At the start, the EVM state is empty (empty root hash), and all the state is stored under a Flow account that is not controlled by anyone (network-owned). Any interaction with this environment emits a transactionExecuted event (direct calls for bridged accounts use `255` as transaction type).  
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,268 @@
...
+* There should be end-to-end tests and benchmarks. If there are not 
+(since this is still a design), how will you track that these will be created?
+
+### Dependencies
+
+* Dependencies: does this proposal add any new dependencies to Flow?
+* Dependent projects: are there other areas of Flow or things that use Flow 
+(Access API, Wallets, SDKs, etc.) that this affects? 
+How have you identified these dependencies and are you sure they are complete? 
+If there are dependencies, how are you managing those changes?
```
franklywatson, line 207:
```suggestion
The project introduces no new dependencies from the Ethereum codebase since those required are already included in flow-go
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,268 @@
...
+* Do you expect changes to binary size / build time / test times?
+* Who will maintain this code? Is this code in its own buildable unit? 
+Can this code be tested in its own? 
+Is visibility suitably restricted to only a small API surface for others to use?
+
+### Best Practices
+
+* Does this proposal change best practices for some aspect of using/developing Flow? 
+How will these changes be communicated/enforced?
+
```
franklywatson, line 220:
```suggestion
```

Can drop this section since is best covered in the docs not the FLIP

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,268 @@
...
+Does the proposed idea/feature exist in other systems and 
+what experience has their community had?
+
+This section is intended to encourage you as an author to think about the 
+lessons learned from other projects and provide readers of the proposal 
+with a fuller picture.
+
+It's fine if there is no prior art; your ideas are interesting regardless of 
+whether or not they are based on existing work.
+
```
franklywatson, line 263:
```suggestion
```
Can drop this section since it's not really relevant

franklywatson (commented):
Overall I think we'll need to provide more colour and detail, perhaps context, where needed since there's a degree of assumed knowledge that might escape some readers. Otherwise mostly some smaller copy edits and tidy ups

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,268 @@
...
+Every Flow transaction has access to this environment similar to the way it has access to other standard contracts like RLP encoding etc.
+
+```
+import EVM from <ServiceAddress>
+```
+
+Every interaction with this smart contract that changes the state results in the formation of a new block (specific chain IDs), which is then stored on-chain. It also emits several FLOW event types (`evm.BlockExecuted`, `evm.TransactionExecuted`) that can be consumed to track the chain progress. For more details, refer to the documentation [here](https://github.com/onflow/flow-go/blob/master/fvm/evm/types/events.go).
+
+Since we use Flow transactions to interact with the Flow EVM environment, there is no need for complex block formation logic, such as handling uncle chains and reorgs. These transactions are executed and verified on Flow just like all other transactions and smart contracts. They are also protected from malicious MEV behaviours using the same mechanism that safeguards all other Flow transactions.
```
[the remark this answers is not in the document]
ramtinms: fixed and added more details.

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,268 @@
...
+
+Every Flow transaction has access to this environment similar to the way it has access to other standard contracts like RLP encoding etc.
+
+```
+import EVM from <ServiceAddress>
+```
+
+Every interaction with this smart contract that changes the state results in the formation of a new block (specific chain IDs), which is then stored on-chain. It also emits several FLOW event types (`evm.BlockExecuted`, `evm.TransactionExecuted`) that can be consumed to track the chain progress. For more details, refer to the documentation [here](https://github.com/onflow/flow-go/blob/master/fvm/evm/types/events.go).
```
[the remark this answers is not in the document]
ramtinms: refactored it.

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,197 @@
+---
+status: draft 
+flip: 223 (set to the issue number)
```
turbolent, line 3:
```suggestion
flip: 223
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,197 @@
...
+sponsor: Dieter Shirley (dete@dapperlabs.com)
+updated: 2023-12-04
+---
+
+# FLIP 223: EVM integration interface
+
+## Objective
+
+- Defining a Cadence interface for the EVM integrated into the FVM. 
+- Facilitates seamless interaction between Cadence and EVM enviornments.
```
turbolent, line 14: Maybe extend this a bit / change to prose

```suggestion
- Defining a Cadence interface for the EVM integrated into the FVM. 
- Facilitates seamless interaction between Cadence and EVM environments. 
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Objective
+
+- Defining a Cadence interface for the EVM integrated into the FVM. 
+- Facilitates seamless interaction between Cadence and EVM enviornments. 
+
+## Motivation
+
+This work would makes it easier for EVM-centric Dapps and Platforms to adopt Flow.
+Please see [this forum discussion](https://forum.flow.com/t/evm-on-flow-beyond-solidity/5260) for motivations.
```
turbolent, line 19: Maybe "copy" some of the main points into this section, to have this FLIP be a bit more self-standing. The link to the full forum post with more details around the motivation is still very useful 👍

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Design Proposal
+
+#### EVM as a standard smart contract
+
+An easy way to understand the approach proposed in this Flip is to consider Flow EVM as a virtual blockchain deployed to the Flow blockchain at a specific address (e.g., a service account). It behaves as if a complete EVM runtime has been implemented in Cadence. However, we leverage the extensibility property of the Cadence runtime and utilize the reference Geth implementation used by many Ethereum nodes. The Cadence runtime is designed to be extensible with standard Cadence smart contracts implemented in native languages, as long as the resource consumption is meter-able and bounded, and operations are deterministic.
```
turbolent, line 25: I would focus on the exposed interface, and remove implementation details, such as using geth, as that only applies to the Flow "reference implementation" (flow-go), but the FLIP should be implementation agnostic (e.g. imagine other implementations to be compatible with the Flow protocol); i.e. remove the third sentence

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,197 @@
...
+
+```
+import EVM from <ServiceAddress>
+```
+
+Within the flow transaction, if EVM interaction is successful 
+
+- it makes changes to the on-chain data
+- forms a new block if successful,
+- emits several FLOW event types (see [here](https://github.com/onflow/flow-go/blob/master/fvm/evm/types/events.go) ) that can be consumed to track the chain progress.
```
turbolent, line 37: Consistently use "Flow" to the protocol. It's a proper noun, so should be capitalized. Also, `FLOW` is the name of the "Flow fungible token", so avoid using that name as well to prevent confusion


```suggestion
Within the Flow transaction, if EVM interaction is successful 

- it makes changes to the on-chain data
- forms a new block if successful,
- emits several Flow event types (see [here](https://github.com/onflow/flow-go/blob/master/fvm/evm/types/events.go)) that can be consumed to track the chain progress.
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,197 @@
...
+And if unsuccessful, it reverts the transaction.
+
+As EVM interactions are encapsulated within Flow transactions, they leverage the security measures provided by Flow. These transactions undergo the same process of collection, execution, and verification as other Flow transactions, without any EVM intervention. Consequently, there is no requirement for intricate block formation logic (such as handling forks and reorganizations), mempools, or additional safeguards against malicious MEV (Miner Extractable Value) behaviours.
+
+On EVM environment, resource consumptions are metered as `gas usage`, when interacting with EVM environment, the total gas usage is translated back into computation usage and would be paid as part of FLOW transaction fees (weigh-adjusted conversion).
```
turbolent, line 42: Maybe stick with one tense instead of mixing them ("is" vs "would"), e.g.

```suggestion
In the EVM environment, resource consumption is metered as "gas usage". When interacting with the EVM environment, the total gas usage is translated back into Flow computation usage and is be paid as part of FLOW transaction fees (weigh-adjusted conversion).
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,197 @@
...
+    access(all) struct EVMAddress {
+
+        /// Bytes of the address
+        access(all) let bytes: [UInt8; 20]
+
+        /// Constructs a new EVM address from the given byte representation
+        init(bytes: [UInt8; 20])
+
+        /// Returns the balance of this address
+        access(all) fun balance(): @Balance
```
turbolent, line 64: Balance is a struct, not a resource

```suggestion
        access(all) fun balance(): Balance
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,197 @@
...
+
+```
+// Example of balance query
+import EVM from <ServiceAddress>
+
+  access(all)
+  fun main(bytes: [UInt8; 20]) {
+      let addr = EVM.EVMAddress(bytes: bytes)
+      let bal = addr.balance()
+  }
```
turbolent, line 103:
```suggestion
access(all)
fun main(bytes: [UInt8; 20]) {
    let addr = EVM.EVMAddress(bytes: bytes)
    let bal = addr.balance()
}
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,197 @@
...
+
+```
+// Example of tx wrapping
+import EVM from <ServiceAddress>
+
+  access(all)
+  fun main(rlpEncodedTransaction: [UInt8], coinbaseBytes: [UInt8; 20]) {
+      let coinbase = EVM.EVMAddress(bytes: coinbaseBytes)
+      EVM.run(tx: rlpEncodedTransaction, coinbase: coinbase)
+  }
```
turbolent, line 122:
```suggestion
access(all)
fun main(rlpEncodedTransaction: [UInt8], coinbaseBytes: [UInt8; 20]) {
    let coinbase = EVM.EVMAddress(bytes: coinbaseBytes)
    EVM.run(tx: rlpEncodedTransaction, coinbase: coinbase)
}
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,197 @@
...
+        fun call(to: EVMAddress, data: [UInt8], gasLimit: UInt64, value: Balance): [UInt8] 
+    }
+
+    /// Creates a new bridged account
+    access(all)
+    fun createBridgedAccount(): @BridgedAccount {
+        return <-create BridgedAccount(
+            addressBytes: InternalEVM.createBridgedAccount()
+        )
+    }
```
turbolent, line 169: Remove implementation details:


```suggestion
    fun createBridgedAccount(): @BridgedAccount
```

turbolent (commented):
👌

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,197 @@
...
+To facilitate the wrapping operation and refunding, the run interface also allows a `coinbase` address to be passed. The use of the `coinbase` address in this context indicates the EVM address which will receive the gas usage * gas price (set in transaction). Essentially, the transaction wrapper behaves similarly to a miner, receives the gas usage fees on an EVM address and pays for the transaction fees. 
+
+Any failure during the execution would revert the whole Flow transaction. 
+
+```
+// Example of tx wrapping
+import EVM from <ServiceAddress>
+
+  access(all)
+  fun main(rlpEncodedTransaction: [UInt8], coinbaseBytes: [UInt8; 20]) {
```
sisyphusSmiling, line 119: This is presented as a script, but I'm assuming the true implementation would be a transaction as mutation is required?

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,197 @@
...
+
+Bridged accounts also facilitate the withdrawal of Flow tokens back from the EVM balance environment into the Cadence environment through `withdraw`.
+
+**What about other fungible and non-tokens?**
+
+The term "bridged accounts" is used because their design facilitates the building of bridges. A Cadence smart contract can control a bridged account, enabling transactional operations between two environments. For instance, this smart contract can receive a Fungible on the Cadence side and send the equivalent on the EVM side to a specified address, all in a single Flow transaction. More details on this will be provided in subsequent updates.
+
+#### Safety and Reproducibility of the EVM state
+
+EVM state is not accessible by Cadence outside of the defined interfaces above, and Cadence state is also protected against raw access from the EVM.
```
sisyphusSmiling, line 183: It sounds like in this sense the "bridge" enabled by BridgedAccounts is unidirectional. On first read, I thought anyone might be able to codify their own tunnel between Flow X EVM contracts, but it seems that seamless, infrastructure-free bridging (as far as dapps are concerned) is only available via Flow -> EVM path, is that right?

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,197 @@
...
+# FLIP 223: EVM integration interface
+
+## Objective
+
+- Defining a Cadence interface for the EVM integrated into the FVM. 
+- Facilitates seamless interaction between Cadence and EVM enviornments. 
+
+## Motivation
+
+This work would makes it easier for EVM-centric Dapps and Platforms to adopt Flow.
```
gregsantos, line 18:
```suggestion
This work would make it easier for EVM-centric Dapps and Platforms to adopt Flow.
```

protocol/20231116-evm-support.md
turbolent: The FLIP only proposes FVM -> EVM interaction because that is the minimum required. Later, another FLIP could also propose EVM -> FVM interaction

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,193 @@
...
+Note that account nonce would protect against double execution of a transaction, similar to how other non-virtual blockchains prevent the minor from including a transaction multiple times. 
+
+#### Bridged accounts
+
+Another major goal for this work is seamless composability across environments. For this goal, we have introduced a new type of address a bridged account, to the EVM environment (besides EOA and Smart Contract accounts). This new address type is similar to EOA’s except instead of being tied to a public/private key it would be controlled by Cadence resources. Unlike EOAs which are created using the key presented by the wallet there is no corresponding EVM key present in a BridgedAccount. Any bridged account is interacted with through BridgedAccount resource and any Flow account or Cadence smart contract that holds this resource could interact with the EVM environment on behalf of the address that is stored in the resource. 
+
+```
+access(all)contract EVM {
+
+    access(all) resource BridgedAccount {
```
sisyphusSmiling, line 136: Apologies if this is out of scope for this FLIP - given that a BridgedAccount is a resource, what happens in the EVM environment if the resource is destroyed? Can it be reconstructed? Is there currently a similar case where an address exists and suddenly no longer does in EVM and would this present any unstated consequences?

turbolent: Destroying the `BridgedAccount` resource is basically equivalent to "throwing away the private key" for an account. Thought technically it could be reconstructed, how should that work – who/what proves ownership of the account?

>  where an address exists and suddenly no longer does in EVM
Can you elaborate? Address on which side?

sisyphusSmiling: > Thought technically reconstructed, how should that work – who/what proves ownership of the account?

While I take it as implied, I don't believe there are any stated guarantees that a `BridgedAccount` address is unique. From the creation method below, it looks address uniqueness would be handled by `InternalEVM`, is that right?

As far as reconstruction, I think the resource itself is sufficient to prove ownership. It make sense to me that destruction would be akin to losing the keys.

> Can you elaborate? Address on which side?

With the `BridgedAccount` destroyed, wouldn't the address no longer exist on either side? I think I overlayed an interface I saw earlier (not listed in this FLIP) which included a method `EVMAddress.exists(): Bool` which I had assume to return `false` once a `BridgedAccount` is destroyed.

Intuitively, I'd expect the address to persist in the EVM environment post-destruction and some artifact of the address in the Cadence environment - maybe in `InternalEVM` in whatever mechanism prevents duplication?

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,197 @@
...
+In the EVM world, there is no concept of accounts or a minimum balance requirement. Any sequence of bytes with a length of 20 is considered a valid address.
+
+Every EVM address has a balance of native tokens (e.g. ETH on Ethereum), a nonce (for deduplication) and a root hash of the state (if smart contract). 
+In this design, we use the FLOW token for this native token. The balance of an EVM address is stored as a smaller denomination of FLOW called `Atto-FLOW`, it works similarly to the way Wei is used to store ETH values on Ethereum.  
+
+```
+access(all) contract EVM {
+
+    /// EVMAddress is an EVM-compatible address
+    access(all) struct EVMAddress {
```
sisyphusSmiling, line 55: It would be helpful to have an easy comparator like a `toString()` method. Given the common pattern mapping addresses to values in EVM, I think it would be surprising to find that `EVMAddress` can't be used as a key in a mapping.

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,197 @@
...
+        access(all) let bytes: [UInt8; 20]
+
+        /// Constructs a new EVM address from the given byte representation
+        init(bytes: [UInt8; 20])
+
+        /// Returns the balance of this address
+        access(all) fun balance(): @Balance
+
+        /// Deposits the given vault into the EVM account with the given address
+        access(all) fun deposit(from: @FlowToken.Vault)
```
sisyphusSmiling, line 67: Since the EVMAddress struct can be publicly constructed, it doesn't necessarily have a corresponding `BridgedAccount` to withdraw deposited funds. This would allow for one to deposit a vault where funds would then be inaccessible. Would there be any problems moving `deposit()` to the `BridgedAccount` resource? I think it would prevent this problem.

turbolent: This is the equivalent of depositing in Flow/Cadence, deposits are public "by default". Likewise in EVM, transfering tokens is possible to any account. Why would we restrict it here?

protocol/20231116-evm-support.md
turbolent: Good idea to add a serialization!

What format would the function return? Hex-encoded? With `0x` prefix or without?
IMHO the function should make that clear in the name, so there is no confusion. `toString` is too general

protocol/20231116-evm-support.md
turbolent: > I don't believe there are any stated guarantees that a `BridgedAccount` address is unique.

Good point, that should be stated explicitly.

> it looks address uniqueness would be handled by `InternalEVM`, is that right?

"`InternalEVM`" is an implementation detail and shouldn't be part of the proposal.

I now think that I might have misread your question – Where you asking if there is a way to regain access to an EVM account for which the bridged account was destroyed in terms of suggesting that should be possible (how I read the question at first), or are you worried and are implying that shouldn't be possible?

> With the `BridgedAccount` destroyed, wouldn't the address no longer exist on either side? I think I overlayed an interface I saw earlier (not listed in this FLIP) which included a method `EVMAddress.exists(): Bool` which I had assume to return `false` once a `BridgedAccount` is destroyed.

Great question! @ramtinms how do you think this will/should work? This should be documented in the proposal.

protocol/20231116-evm-support.md
sisyphusSmiling: > What format would the function return? Hex-encoded? With 0x prefix or without?

Whichever format is returned should support deserialization into an `EVMAddress` as well. Since hex decoding doesn't support a prefix, I think we should exclude `0x`.

protocol/20231116-evm-support.md
sisyphusSmiling: I was asking out of concern, thinking it shouldn't be possible out of concern for implications in the EVM environment.

protocol/20231116-evm-support.md
sisyphusSmiling: > This is the equivalent of depositing in Flow/Cadence, deposits are public "by default". Likewise in EVM, transfering tokens is possible to any account. Why would we restrict it here?

Good point. Still wrapping my head around matching EVM patterns and expectations in Cadence.

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,193 @@
...
+
+        /// constructs a new bridged account for the address
+        init(addressBytes: [UInt8; 20]) 
+
+        /// The EVM address of the bridged account
+        access(all)
+        fun address(): EVMAddress 
+
+        /// Withdraws the balance from the bridged account's balance
+        access(all)
```
sisyphusSmiling, line 149: Not sure how EVM launch will line up with Cadence 1.0, but it would make sense for privileged methods to be entitled ensuring only the resource owner can withdraw, deploy and make calls. If planned deployment schedule is such that EVM contracts will be deployed pre-Cadence 1.0, I'm wondering if we should define and implement interfaces to refine access.

For example, I'd probably want `BridgedAccount.address()` to be accessible publicly on my Flow account. However, there's currently no way for me to expose that functionality without also allowing public withdrawal.

Entitlements that make sense might include `Withdrawable`, `Deployable` and `Callable` for each sensitive method.

franklywatson: The plan of record is that we will launch EVM on Cadence 1.0 barring any major upset to their current schedule. Right now emulator version being used if 0.4, but we are targeting being on 1.0 when launching on Crescendo network

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,192 @@
...
+        fun call(to: EVMAddress, data: [UInt8], gasLimit: UInt64, value: Balance): [UInt8] 
+    }
+
+    /// Creates a new bridged account
+    access(all)
+    fun createBridgedAccount(): @BridgedAccount
+}
+```
+
+Bridged account addresses are allocated by the FVM and stored inside the resource. Calls through bridged accounts form a new type of transaction for the EVM that doesn’t require signatures and doesn’t need nonce checking. Bridged accounts could deploy smart contracts or make calls to the ones that are already deployed on Flow EVM.
```
franklywatson, line 168:
```suggestion
Bridged account addresses are unique and allocated by the FVM and stored inside the resource. Calls through bridged accounts form a new type of transaction for the EVM that doesn’t require signatures and doesn’t need nonce checking. Bridged accounts could deploy smart contracts or make calls to the ones that are already deployed on Flow EVM. 
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,194 @@
...
+
+        /// Deploys a contract to the EVM environment.
+        /// Returns the address of the newly deployed contract
+        access(all)
+        fun deploy(code: [UInt8], gasLimit: UInt64, value: Balance): EVMAddress 
+
+        /// Calls a function with the given data.
+        /// The execution is limited by the given amount of gas
+        access(all)
+        fun call(to: EVMAddress, data: [UInt8], gasLimit: UInt64, value: Balance): [UInt8]
```
sisyphusSmiling, line 161: Does the existence of the method in the BridgedAccount imply that there is no other way for a BridgedAccount to transmit a call into EVM? I'm wondering if I'd be able to sequence a BridgedAccount EVM call within the function scope of a contract method.

More concretely, let's say I have a method:

```cadence
fun bridgeNFTFromEVM(
    caller: auth(Verifiable) BridgedAccount,
    nftID: UInt64,
    contractAddress: EVMAddress,
    transferCallData: [UInt8],
    value: Balance
): @NonfungibleToken.NFT {
	// Validate that EVMNFT.ownerOf[nftID] == caller.address()
	// Execute transfer to bridge account COA by passing throught transferCallData
		// ^^^ Is this possible from within a contract? ^^^
	// Validate that EVMNFT.ownerOf[nftID] == bridgeAccount.address()
	// Continue...
}
```
I'm imagining not since the data can't be signed by the BridgedAccount and the authentication mechanism here is that it's being executed from the resource itself. 

If we can't orchestrate calls in this way, what are thoughts about requiring an entitlement on `call` like `Callable`? That may be useful outside of this usecase anyway

turbolent: > Does [this] imply that there is no other way for a BridgedAccount to transmit a call into EVM?

What other mechanism are you thinking of? The `call` function closely resembles the EVM's `call` opcode.

> Is this possible from within a contract?

If I understand your questions, you'd simply `caller.call` there.

> data can't be signed by the BridgedAccount and the authentication mechanism here is that it's being executed from the resource itself.

Sorry, I don't quite understand the question. How are signatures involved here?

Maybe it is easier to discuss what you are trying to achieve, and how that would be done in Solidity / on the EVM, and then see how it could be translated to Cadence and the existing EVM API, or in case it is not possible, how it needs to be adjusted.

turbolent: > what are thoughts about requiring an entitlement on call like Callable? That may be useful outside of this usecase anyway

Yes, in general, the proposal is based on the currently released version of Cadence, which does not have entitlements. For Cadence 1.0, it might make sense to add coarse and fine-grained entitlements, so that parts of the functionality of bridged accounts can be exposed.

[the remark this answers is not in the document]
ramtinms: hmm, I think these ones are different, they don't move money from the Cadence land, they move money from the evm address balance to the call.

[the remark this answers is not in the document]
ramtinms: these params are basically the same param an evm transaction gets

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,194 @@
...
+Any failure during the execution would revert the whole Flow transaction. 
+
+```
+// Example of tx wrapping
+import EVM from <ServiceAddress>
+
+access(all)
+fun main(rlpEncodedTransaction: [UInt8], coinbaseBytes: [UInt8; 20]) {
+    let coinbase = EVM.EVMAddress(bytes: coinbaseBytes)
+    EVM.run(tx: rlpEncodedTransaction, coinbase: coinbase)
```
bluesign, line 122: not familiar with evm world, but I feel like `EVM.run` should return something to Cadence land to be useful somehow.

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,194 @@
...
+        access(self)
+        let addressBytes: [UInt8; 20]
+
+        /// constructs a new bridged account for the address
+        init(addressBytes: [UInt8; 20]) 
+
+        /// The EVM address of the bridged account
+        access(all)
+        fun address(): EVMAddress 
+
```
bluesign, line 148: maybe something like `fun balance()` here can be useful.

protocol/20231116-evm-support.md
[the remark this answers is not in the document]
turbolent: Ah, sorry, for a moment I got confused here – the balance is taken from the EVM account of course!
I'll update the comments to clarify that

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,194 @@
...
+        access(all)
+        fun withdraw(balance: Balance): @FlowToken.Vault
+
+        /// Deploys a contract to the EVM environment.
+        /// Returns the address of the newly deployed contract
+        access(all)
+        fun deploy(code: [UInt8], gasLimit: UInt64, value: Balance): EVMAddress 
+
+        /// Calls a function with the given data.
+        /// The execution is limited by the given amount of gas
```
turbolent, line 159:
```suggestion
        /// The execution is limited by the given amount of gas.
        /// The value (balance) is taken from the EVM account.
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,194 @@
...
+        /// The EVM address of the bridged account
+        access(all)
+        fun address(): EVMAddress 
+
+        /// Withdraws the balance from the bridged account's balance
+        access(all)
+        fun withdraw(balance: Balance): @FlowToken.Vault
+
+        /// Deploys a contract to the EVM environment.
+        /// Returns the address of the newly deployed contract
```
turbolent, line 154:
```suggestion
        /// Returns the address of the newly deployed contract.
        /// The value (balance) is taken from the EVM account.
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,196 @@
...
+    let coinbase = EVM.EVMAddress(bytes: coinbaseBytes)
+    EVM.run(tx: rlpEncodedTransaction, coinbase: coinbase)
+}
+```
+
+For example, a user might use Metamask to sign a transaction for Flow EVM and broadcast it to services that check the gas fee on the transaction and wrap the transaction to be executed. 
+
+Note that account nonce would protect against double execution of a transaction, similar to how other non-virtual blockchains prevent the minor from including a transaction multiple times. 
+
+#### Bridged accounts
```
SeanRobb, line 130: I suggest we consider calling this a Cadence Owned Account.

The impact of this account type is more a connective tissue between Cadence execution environment and EVM. Interacting and engaging with EVM composing with solidity smart contracts where needed.

This additionally fits nicely with the existing naming convention with "EOA" or "External Owned Account".

turbolent: +1 on this. Let's rename in the proposal text, add a variation of this explanation, and also adjust the API (i.e. rename `BridgedAccount` to `CadenceOwnedAccount`)

protocol/20231116-evm-support.md
sisyphusSmiling: I thought through it more carefully and believe my original question was unfounded. Entitlements will cover any other concerns I had!

protocol/20231116-evm-support.md
franklywatson: Name change to Cadence Owned Account merged

protocol/20231116-evm-support.md
ramtinms: EVMAddress it self exposes balance, but we can add sugar syntax for it.

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,198 @@
...
+access(all) contract EVM {
+
+    /// EVMAddress is an EVM-compatible address
+    access(all) struct EVMAddress {
+
+        /// Bytes of the address
+        access(all) let bytes: [UInt8; 20]
+
+        /// Constructs a new EVM address from the given byte representation
+        init(bytes: [UInt8; 20])
```
sisyphusSmiling, line 62: I'm finding it frustrating to convert between `[UInt8]` and ``[UInt8; 20]` when constructing EVMAddress and I think others will as well. For example, given some hex encoded address, I would want to construct an EVMAddress in a script to call a solidity contract. I would expect I could simply:

```cadence
let evmAddress = EVMAddress(bytes: hexAddress.decodeHex())
```

Instead, I have to do:
```cadence
let evmAddressBytes: [UInt8] = evmContractAddressHex.decodeHex()
    let evmAddress = EVM.EVMAddress(
            bytes: [
                evmAddressBytes[0], evmAddressBytes[1], evmAddressBytes[2], evmAddressBytes[3], evmAddressBytes[4],
                evmAddressBytes[5], evmAddressBytes[6], evmAddressBytes[7], evmAddressBytes[8], evmAddressBytes[9],
                evmAddressBytes[10], evmAddressBytes[11], evmAddressBytes[12], evmAddressBytes[13], evmAddressBytes[14],
                evmAddressBytes[15], evmAddressBytes[16], evmAddressBytes[17], evmAddressBytes[18], evmAddressBytes[19]
            ]
        )
```

I think it would be preferable to abstract away the complexities of dealing with the restricted byte array in the struct's initialization.

turbolent: It would be great to keep the type-safety, though I see and agree with the UX problems.

If the input type would be a variable-sized array, what would be the behaviour for input that is too short or too long? Developers might make assumptions, which could potentially be different from actual behaviour, and thus lead to costly mistakes.

For converting between constant-size and variable-size arrays, there is actually a Cadence feature request issue open: https://github.com/onflow/cadence/issues/2530

protocol/20231116-evm-support.md
turbolent: Good point!

I noticed that the FLIP was actually lacking the definition of `EVM.run`, so I added it in 6dbd64d32d5a9d78d04c958118a300958e4ea93f.

@ramtinms I noticed that the current implementation does not yet return the documented boolean return value. Also, I think at some point there was a discussion about if the function should:
- Not return a result value indicating success, but rather cause the outer Flow/Cadence transaction to abort (not proposing it, just documenting this was considered, and noting it should be part of the proposal;)
- Return a result value, but not just a boolean value, but e.g. a struct with more information (reason / error message? what else could be returned?)

protocol/20231116-evm-support.md
turbolent: Already part of the implementation, added below: https://github.com/onflow/flips/pull/225/files#diff-751f944fe9726f1b53ff7bc95a2093a616d0dbed1b10a8efc8507989aa0af9d1R176

protocol/20231116-evm-support.md
ramtinms: Unfortunatly it doesn't return anything, but the calls on the COA will return a value.

turbolent: Right, it makes sense that there is no value returned from executing an EVM transaction (just like executing a transaction on Flow does not return a value). 

However, it might still be useful to return a value from the function that runs the transaction which indicates the "result" of the transaction, e.g. if it succeeded or failed, e.g. as a boolean, or a struct which contains more information about the transaction execution. 

@ramtinms Could you maybe point me to where in the code (flow-go and geth) where/how an EVM transaction is run and what the result of it is?

protocol/20231116-evm-support.md
sisyphusSmiling: I think a pre-condition like `bytes.length == 20` in `EVMAddress.init` is behaviorally similar to the current implementation. Is that workable?

turbolent: A pre-condition is a dynamic (run-time) check, and not a static one (type-checking time). It would be nice not to have to express the requirement of the length statically so errors can be prevented as early as possible

protocol/20231116-evm-support.md
turbolent: Good point! Changed in aa6ca62f068744a475167355c921725f8aa429bd

protocol/20231116-evm-support.md
turbolent: The implementation of run currently aborts on any error: https://github.com/onflow/flow-go/blob/645923211a85641ade7d335c773d627390de4901/fvm/evm/handler/handler.go#L77.

Is that the intended behaviour? Then we should update the proposal and remove the current bool return value.

If it isn't: It looks like `BlockView.RunTransaction` returns a transaction result value (https://github.com/onflow/flow-go/blob/645923211a85641ade7d335c773d627390de4901/fvm/evm/emulator/emulator.go#L110-L112), `types.Result`, and we could expose (parts of) it to Cadence?

ramtinms:
Just a heads up we have new findings and received more feedback and I will update this Flip soon with more details.

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,238 @@
...
+
+        /// Bytes of the address
+        access(all)
+        let bytes: [UInt8; 20]
+
+        /// Constructs a new EVM address from the given byte representation
+        init(bytes: [UInt8; 20])
+
+        /// Returns the balance of this address
+        access(all)
```
sisyphusSmiling, line 68: This should be `view` so we can post-condition transactions on balances. Currently, the following post-condition
```cadence
post {
    self.preBalance + amount == self.bridgedAccount.balance().flow: "Error executing transfer!"
}
```
Results in the following
```cadence
Impure operation performed in view context
```

turbolent: As noted earlier (in regards to entitlements), the current API proposal is for the latest Cadence version (v0.42), which does not have `view` functions.

A future update should make it compatible and suitable for Cadence 1.0.

ramtinms: good point, we need to update when `view` become available.

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,479 @@
...
+
+- `call(to: EVMAddress, data: [UInt8], gasLimit: UInt64, value: Balance): [UInt8]` makes a call on behalf of the COA smart contract and the result of the call (the returned value) could be processed by the cadence side. The value (balance) is taken from the COA smart contract. Calling this method emits direct call transaction events. 
+
+
+#### COA’s EVM smart contract interface
+
+- `function supportsInterface(bytes4 interfaceId) external view returns (bool)` provides the functionality needed to satisfy EIP-165. If the given interfaceID is available it returns true. 
+
+- `receive() external payable` is the standard method needed to allow other EVM users to send ETH to the smart contract directly, if data is passed to this method it would revert. This method is not usually directly called and any call to smart contract it without the data is automatically rerouted to this function.
```
franklywatson, line 147: I assume this would make sense to an Ethereum builder because it doesn't to me 😀

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,479 @@
...
+
+- `function onERC721Received(address _operator, address _from, uint256 _tokenId, bytes calldata _data) external returns (bytes4)` is provided to support safe transfers from ERC721 asset contracts. It's called on the recipient after a transfer. Return of any value other than the magic value (`0x150b7a02`) must result in the transaction being reverted.
```
franklywatson, line 149: Is this magic value also a thing on other EVMs or this is specific to Flow EVM?

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,479 @@
...
+
+- `function tokensReceived(address operator, address from, address to, uint256 amount, bytes calldata data, bytes calldata operatorData) external` is provided to support safe transfers from ERC777 asset contracts. is called by the ERC777 token contract after a successful transfer or a minting operation. It doesn’t return anything.
+
+`function onERC1155Received(address _operator, address _from, uint256 _id, uint256 _value, bytes calldata _data) external returns (bytes4)` is provided to support safe transfers from ERC1155 asset contracts.  It should return 0xf23a6e61 if supporting the receiving of a single ERC1155 token type. 
+
+`function onERC1155BatchReceived(address _operator, address _from, uint256[] calldata _ids, uint256[] calldata _values, bytes calldata _data) external returns (bytes4)` is provided to support safe transfers from ERC1155 asset contracts (batch of assets). It returns `0xbc197c81` if supporting the receiving of a batch of ERC1155 token types.
```
franklywatson, line 155:
```suggestion
- `function onERC1155Received(address _operator, address _from, uint256 _id, uint256 _value, bytes calldata _data) external returns (bytes4)` is provided to support safe transfers from ERC1155 asset contracts.  It should return 0xf23a6e61 if supporting the receiving of a single ERC1155 token type. 

- `function onERC1155BatchReceived(address _operator, address _from, uint256[] calldata _ids, uint256[] calldata _values, bytes calldata _data) external returns (bytes4)` is provided to support safe transfers from ERC1155 asset contracts (batch of assets). It returns `0xbc197c81` if supporting the receiving of a batch of ERC1155 token types.
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,479 @@
...
+
+- `function isValidSignature(bytes32 _hash, bytes memory _signature) external view virtual returns (bytes4)` returns the bytes4 magic value `0x1626ba7e` when the signature is valid. This method is usually used to verify a personal sign for the smart contract wallets. (see EIP-1271 for more details). In the context of the COA smart contracts, we consider the signature as an aggregation of flow account address, key index, path to COA resource and a set of signatures (flow account). we return true if the signatures are valid, it provides enough weight for the account, and an account holds the resource at the given path. Under the hood, this method uses Cadence Arch contract for verification.
```
franklywatson, line 157:
```suggestion
- `function isValidSignature(bytes32 _hash, bytes memory _signature) external view virtual returns (bytes4)` returns the bytes4 magic value `0x1626ba7e` when the signature is valid. This method is usually used to verify a personal sign for the smart contract wallets (see EIP-1271 for more details). In the context of COA smart contracts, we consider the signature as an aggregation of Flow account address, key index, path to COA resource and a set of signatures (Flow account). we return true if the signatures are valid, it provides enough weight for the account, and an account holds the resource at the given path. Under the hood, this method uses Cadence Arch contract for verification.
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,479 @@
...
+
+
+## Appendix A - Embracing the EVM ecosystem.
+
+In this Flip, we described the foundation of the Flow EVM, yet there are other works built on top of this foundation to ensure an effortless onboarding experience for the existing EVM ecosystem product and tools. 
+
+**The Flow Gateway Software**
```
franklywatson, line 164:
```suggestion
**Flow EVM Gateway Software**
```

Keeping naming consistent

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,479 @@
...
+
+For example, a user might use MetaMask to sign a transaction for Flow EVM and broadcast it to services that check the gas fee on the transaction and wrap the transaction to be executed. Note that account nonce would protect against double execution of a transaction, similar to how other non-virtual blockchains prevent the minor from including a transaction multiple times.
+
+As EVM interactions are encapsulated within Flow transactions, they leverage the security measures provided by Flow. These transactions undergo the same process of collection, execution, and verification as other Flow transactions, without any EVM intervention. Consequently, there is no requirement for intricate block formation logic (such as handling forks and reorganizations), mempools, or additional safeguards against malicious MEV (Miner Extractable Value) behaviours. More information about Flow's consensus model is available [here](https://flow.com/core-protocol).
+
+You can read more about this work at [this Flip](https://github.com/onflow/flips/pull/235).
```
franklywatson, line 174:
```suggestion
You can read more about this work in the [Flow EVM Gateway](https://github.com/onflow/flips/pull/235) improvement proposal. 
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,479 @@
...
+
+
+## Appendix A - Embracing the EVM ecosystem.
+
+In this Flip, we described the foundation of the Flow EVM, yet there are other works built on top of this foundation to ensure an effortless onboarding experience for the existing EVM ecosystem product and tools. 
+
+**The Flow Gateway Software**
+
+A separate software package is provided by the Flow Foundation that can be run by anyone (including the Flow Foundation), that follows the Flow EVM chain, sends queries to the Flow access nodes, and provides the JSON-RPC endpoints to any 3rd party applications that want to interact directly with the Flow EVM. With the exception of account state proof endpoints, all other endpoints of the JSON-RPC is going to be provided at the release.
```
franklywatson, line 166:
```suggestion
A separate software package is provided by the Flow Foundation that can be run by anyone (including the Flow Foundation), that follows the Flow EVM chain, sends queries to the Flow access nodes, and provides the JSON-RPC endpoints to any 3rd party applications that want to interact directly with the Flow EVM. With the exception of account state proof endpoints, all other endpoints of the JSON-RPC will be provided at the release. 
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,479 @@
...
+
+**The Flow Gateway Software**
+
+A separate software package is provided by the Flow Foundation that can be run by anyone (including the Flow Foundation), that follows the Flow EVM chain, sends queries to the Flow access nodes, and provides the JSON-RPC endpoints to any 3rd party applications that want to interact directly with the Flow EVM. With the exception of account state proof endpoints, all other endpoints of the JSON-RPC is going to be provided at the release. 
+
+Any 3rd party running this software can accept EVM RLP encoded transactions from users and wrap them in a Flow transaction.  While they will pay for the fees, EVM.run provides a utility parameter (`coinbase`) to collect the gas fee from the user on the Flow EVM side while running the EVM transaction. The use of the coinbase address in this context indicates the EVM address which will receive the `gas usage * gas price` (set in transaction). Essentially, the transaction wrapper behaves similarly to a miner, receives the gas usage fees on an EVM address and pays for the transaction fees. The `gas price per unit of gas` creates a marketplace for these 3rd parties to compete over transactions.
```
franklywatson, line 168:
```suggestion
Any 3rd party running this software can accept EVM RLP encoded transactions from users and wrap them in a Flow transaction.  While they will pay for the fees, `EVM.run` provides a utility parameter (`coinbase`) to collect the gas fee from the user on the Flow EVM side while running the EVM transaction. The use of the `coinbase` address in this context indicates the EVM address which will receive the `gas usage * gas price` (set in transaction). Essentially, the transaction wrapper behaves similarly to a miner, receives the gas usage fees on an EVM address and pays for the transaction fees. The `gas price per unit of gas` creates a marketplace for these 3rd parties to compete over transactions. 
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,479 @@
...
+
+Any 3rd party running this software can accept EVM RLP encoded transactions from users and wrap them in a Flow transaction.  While they will pay for the fees, EVM.run provides a utility parameter (`coinbase`) to collect the gas fee from the user on the Flow EVM side while running the EVM transaction. The use of the coinbase address in this context indicates the EVM address which will receive the `gas usage * gas price` (set in transaction). Essentially, the transaction wrapper behaves similarly to a miner, receives the gas usage fees on an EVM address and pays for the transaction fees. The `gas price per unit of gas` creates a marketplace for these 3rd parties to compete over transactions. 
+
+For example, a user might use MetaMask to sign a transaction for Flow EVM and broadcast it to services that check the gas fee on the transaction and wrap the transaction to be executed. Note that account nonce would protect against double execution of a transaction, similar to how other non-virtual blockchains prevent the minor from including a transaction multiple times.
```
franklywatson, line 170:
```suggestion
For example, a user might use MetaMask to sign a transaction for Flow EVM and broadcast it to services that check the gas fee on the transaction and wrap the transaction to be executed. Note that account nonce would protect against double execution of a transaction, similar to how other non-virtual blockchains prevent the miner from including a transaction multiple times.
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,479 @@
...
+
+As EVM interactions are encapsulated within Flow transactions, they leverage the security measures provided by Flow. These transactions undergo the same process of collection, execution, and verification as other Flow transactions, without any EVM intervention. Consequently, there is no requirement for intricate block formation logic (such as handling forks and reorganizations), mempools, or additional safeguards against malicious MEV (Miner Extractable Value) behaviours. More information about Flow's consensus model is available [here](https://flow.com/core-protocol).
+
+You can read more about this work at [this Flip](https://github.com/onflow/flips/pull/235). 
+
+**Token Bridge (Cadence <> Flow EVM)**
```
franklywatson, line 176:
```suggestion
**Flow VM Bridge (Cadence <> Flow EVM)** 
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,479 @@
...
+
+- `function isValidSignature(bytes32 _hash, bytes memory _signature) external view virtual returns (bytes4)` returns the bytes4 magic value `0x1626ba7e` when the signature is valid. This method is usually used to verify a personal sign for the smart contract wallets. (see EIP-1271 for more details). In the context of the COA smart contracts, we consider the signature as an aggregation of flow account address, key index, path to COA resource and a set of signatures (flow account). we return true if the signatures are valid, it provides enough weight for the account, and an account holds the resource at the given path. Under the hood, this method uses Cadence Arch contract for verification.
+
+
+## Appendix A - Embracing the EVM ecosystem.
+
+In this Flip, we described the foundation of the Flow EVM, yet there are other works built on top of this foundation to ensure an effortless onboarding experience for the existing EVM ecosystem product and tools.
```
franklywatson, line 162:
```suggestion
In this FLIP, we described the foundation of the Flow EVM, yet there are other works built on top of this foundation to ensure an effortless onboarding experience for the existing EVM ecosystem product and tools. 
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,479 @@
...
+
+As EVM interactions are encapsulated within Flow transactions, they leverage the security measures provided by Flow. These transactions undergo the same process of collection, execution, and verification as other Flow transactions, without any EVM intervention. Consequently, there is no requirement for intricate block formation logic (such as handling forks and reorganizations), mempools, or additional safeguards against malicious MEV (Miner Extractable Value) behaviours. More information about Flow's consensus model is available [here](https://flow.com/core-protocol).
+
+You can read more about this work at [this Flip](https://github.com/onflow/flips/pull/235). 
+
+**Token Bridge (Cadence <> Flow EVM)** 
+Cadence Owned Accounts provide out of the box FLOW token bridging across environment, though they are powerful resources to build bridges that can bridge any fungible and non-fungible tokens between Cadence and Flow EVM. Checkout [this Flip](https://github.com/onflow/flips/pull/233) for more details.
```
franklywatson, line 177:
```suggestion
COAs provide out of the box $FLOW token bridging between environments. They are also powerful resources which integrate native cross-VM bridging capabilities through which applications can bridge arbitrary fungible and/or non-fungible tokens between Cadence and Flow EVM. Checkout the [Flow VM Bridge ](https://github.com/onflow/flips/pull/233) improvement proposal for more details. 
```

protocol/20231116-evm-support.md
turbolent:
```suggestion
- `function onERC721Received(address _operator, address _from, uint256 _tokenId, bytes calldata _data) external returns (bytes4)` is provided to support safe transfers from ERC721 asset contracts. It's called on the recipient after a transfer. Return of any value other than the magic value (`0x150b7a02`, as defined in the [ERC-721 standard](https://eips.ethereum.org/EIPS/eip-721)) must result in the transaction being reverted. 
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,479 @@
...
+
+For example, a user might use MetaMask to sign a transaction for Flow EVM and broadcast it to services that check the gas fee on the transaction and wrap the transaction to be executed. Note that account nonce would protect against double execution of a transaction, similar to how other non-virtual blockchains prevent the miner from including a transaction multiple times.
+
+As EVM interactions are encapsulated within Flow transactions, they leverage the security measures provided by Flow. These transactions undergo the same process of collection, execution, and verification as other Flow transactions, without any EVM intervention. Consequently, there is no requirement for intricate block formation logic (such as handling forks and reorganizations), mempools, or additional safeguards against malicious MEV (Miner Extractable Value) behaviours. More information about Flow's consensus model is available [here](https://flow.com/core-protocol).
+
+You can read more about this work in the [Flow EVM Gateway](https://github.com/onflow/flips/pull/235) improvement proposal. 
+
+**Flow VM Bridge (Cadence <> Flow EVM)**
```
franklywatson, line 176:
```suggestion
**Flow VM Bridge (Cadence <> Flow EVM)** 

```

turbolent:
@ramtinms 

- Is there still content that needs to be added before the status can be updated from draft to proposed? 
- Are there any open questions that still need to be figured out?

ramtinms:
@turbolent just changed the state to "proposed".

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,496 @@
...
+
+- `function onERC1155BatchReceived(address _operator, address _from, uint256[] calldata _ids, uint256[] calldata _values, bytes calldata _data) external returns (bytes4)` is provided to support safe transfers from ERC1155 asset contracts (batch of assets). It returns `0xbc197c81` if supporting the receiving of a batch of ERC1155 token types.
+
+- `function isValidSignature(bytes32 _hash, bytes memory _signature) external view virtual returns (bytes4)` returns the bytes4 magic value `0x1626ba7e` when the signature is valid. This method is usually used to verify a personal sign for the smart contract wallets (see EIP-1271 for more details). In the context of COA smart contracts, we consider the signature as an aggregation of Flow account address, key index, path to COA resource and a set of signatures (Flow account). we return true if the signatures are valid, it provides enough weight for the account, and an account holds the resource at the given path. Under the hood, this method uses Cadence Arch contract for verification.
```
turbolent, line 157:
```suggestion
- `function isValidSignature(bytes32 _hash, bytes memory _signature) external view virtual returns (bytes4)` returns the bytes4 magic value `0x1626ba7e` (as defined in [ERC-1271](https://eips.ethereum.org/EIPS/eip-1271)) when the signature is valid. This method is usually used to verify a personal sign for the smart contract wallets (see EIP-1271 for more details). In the context of COA smart contracts, we consider the signature as an aggregation of Flow account address, key index, path to COA resource and a set of signatures (Flow account). we return true if the signatures are valid, it provides enough weight for the account, and an account holds the resource at the given path. Under the hood, this method uses Cadence Arch contract for verification.
```

franklywatson (approved):

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,496 @@
...
+
+- `call(to: EVMAddress, data: [UInt8], gasLimit: UInt64, value: Balance): [UInt8]` makes a call on behalf of the COA smart contract and the result of the call (the returned value) could be processed by the cadence side. The value (balance) is taken from the COA smart contract. Calling this method emits direct call transaction events. 
+
+
+#### COA’s EVM smart contract interface
+
+- `function supportsInterface(bytes4 interfaceId) external view returns (bool)` provides the functionality needed to satisfy EIP-165. If the given interfaceID is available it returns true. 
+
+- `receive() external payable` is the standard method needed to allow other EVM users to send ETH to the smart contract directly, if data is passed to this method it would revert. This method is not usually directly called and any call to smart contract it without the data is automatically rerouted to this function.
```
sideninja, line 147: Shouldn't it be Flow
```suggestion
- `receive() external payable` is the standard method needed to allow other EVM users to send Flow to the smart contract directly, if data is passed to this method it would revert. This method is not usually directly called and any call to smart contract it without the data is automatically rerouted to this function.
```

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,496 @@
...
+
+**Flow EVM Gateway Software**
+
+A separate software package is provided by the Flow Foundation that can be run by anyone (including the Flow Foundation), that follows the Flow EVM chain, sends queries to the Flow access nodes, and provides the JSON-RPC endpoints to any 3rd party applications that want to interact directly with the Flow EVM. With the exception of account state proof endpoints, all other endpoints of the JSON-RPC will be provided at the release. 
+
+Any 3rd party running this software can accept EVM RLP encoded transactions from users and wrap them in a Flow transaction.  While they will pay for the fees, `EVM.run` provides a utility parameter (`coinbase`) to collect the gas fee from the user on the Flow EVM side while running the EVM transaction. The use of the `coinbase` address in this context indicates the EVM address which will receive the `gas usage * gas price` (set in transaction). Essentially, the transaction wrapper behaves similarly to a miner, receives the gas usage fees on an EVM address and pays for the transaction fees. The `gas price per unit of gas` creates a marketplace for these 3rd parties to compete over transactions.
```
JeffreyDoyle, line 168:
```suggestion
Any 3rd party running this software can accept EVM RLP encoded transactions from users and wrap them in a Cadence transaction.  While they will pay for the fees, `EVM.run` provides a utility parameter (`coinbase`) to collect the gas fee from the user on the Flow EVM side while running the EVM transaction. The use of the `coinbase` address in this context indicates the EVM address which will receive the `gas usage * gas price` (set in transaction). Essentially, the transaction wrapper behaves similarly to a miner, receives the gas usage fees on an EVM address and pays for the transaction fees. The `gas price per unit of gas` creates a marketplace for these 3rd parties to compete over transactions. 
```

turbolent:
Does the current state of the proposal reflect the changes that were made in the implementation?

sideninja:
@turbolent I've updated the contract and I believe it should be up to date now.

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,877 @@
+---
+status: proposed
+flip: 223
+authors: Ramtin Seraj (ramtin.seraj@flowfoundation.org), Bastian Müller (bastian@dapperlabs.com)
```
turbolent, line 4:
```suggestion
authors: Ramtin Seraj (ramtin.seraj@flowfoundation.org), Bastian Müller (bastian.mueller@flowfoundation.org)
```

protocol/20231116-evm-support.md
```diff
@@ -389,16 +389,24 @@ contract EVM {
         access(all)
         let data: [UInt8]
 
+        /// returns the newly deployed contract address
```
turbolent, line 392: The comment for `data` above says 
>For coa.deploy calls it returns the address bytes of the newly deployed contract.

Why is `deployedContractAddress` added, if it is already in `data`?

sideninja: because there's another PR to fix that, actually that's already in that branch, I just need to merge. Sorry for confusion. https://github.com/onflow/flips/pull/257

sideninja: Actually no, wait :D got myself confused now. You are right. Updated here

protocol/20231116-evm-support.md
```diff
@@ -0,0 +1,882 @@
+---
+status: proposed
```
franklywatson, line 2:
```suggestion
status: implemented
```

turbolent:
@franklywatson [13cc05f](https://github.com/onflow/flips/pull/225/commits/13cc05fcf5950cc9d55f9086a67641c7f50f092e) changed the status to implemented, but the FLIP hasn't even been accepted yet? See https://github.com/onflow/flips?tab=readme-ov-file#proposal-states

ramtinms:
@KshitijChaudhary666 what is the process for moving this forward? is there a meeting to present it?

turbolent:
@ramtinms This proposal probably lies in between the [Cadence Language and Execution Working Group](https://github.com/onflow/Flow-Working-Groups/tree/main/cadence_language_and_execution_working_group) and [Flow Core-Protocol Working Group](https://github.com/onflow/Flow-Working-Groups/tree/main/core_protocol_working_group). 

Maybe indicate that this proposal is ready for a final round of feedback, and will get approved unless there are any further concerns brought forward. For example, post in Discord and in the forum.

We can maybe schedule a working group meeting soon to have a last discussion, and then get it approved (so far sentiment has been positive).

turbolent:
@ramtinms Is the FLIP up-to-date with the latest changes made in flow-go?

KshitijChaudhary666:
> @KshitijChaudhary666 what is the process for moving this forward? is there a meeting to present it?

Hey - simply talking from a FLIP-lifecycle-perspective - presenting the FLIP in the next WG meeting for a final feedback like @turbolent proposed does make sense to me. Once the author is confident of the community sentiment, the status of the FLIP could be changed to "accepted" and thereafter go to implementation
@vishalchangrani

ramtinms:
@turbolent I believe it has all the recent changes. @sideninja any thing we might be missing?

AlexHentschel:
Discord conversation on getting this FLIP over the finish line and a Working Group Meeting for a final discussion and getting the FLIP formally approved: https://discord.com/channels/613813861610684416/1167476806333513800/1237414321445797939

AlexHentschel:
## 👉 Announcement 👈 

Thank you for the insightful discussions on this FLIP. We feel there is an overwhelming support from the active developer community and the broader ecosystem. 

On **May 15th**, 2024 **10-11am PT (17-18:00 UTC)** we'll hold a joint **meeting** of the [Core Protocol Working Group](https://www.google.com/url?q=https://github.com/onflow/Flow-Working-Groups/tree/main/core_protocol_working_group&sa=D&source=calendar&ust=1715626269987743&usg=AOvVaw3xNADeHWNuPnhG2Dc4JdkJ) and the [Cadence Language and Execution Working Group](https://www.google.com/url?q=https://github.com/onflow/Flow-Working-Groups/tree/main/cadence_language_and_execution_working_group&sa=D&source=calendar&ust=1715626269987743&usg=AOvVaw0cAlXS0qgEpWxMydqQNwmY), where we would like to briefly go over it together and approve this FLIP unless there are any major concerns remaining. 

You can add the meeting to your calendar via [this link](https://calendar.google.com/calendar/event?action=TEMPLATE&tmeid=MmoyN2Q0bGx0Y3Q1ZnFib2dsNWIyNnM1bnRfMjAyNDA1MjNUMTcwMDAwWiBjXzQ3OTc4ZjVjZDlkYTYzNmNhZGM2Yjg0NzMxMDJiNTA5MmMxYTg2NWRkMDEwNTU4MzkzZWNiN2Y5ZmQwYzlhZDBAZw&tmsrc=c_47978f5cd9da636cadc6b8473102b5092c1a865dd010558393ecb7f9fd0c9ad0%40group.calendar.google.com&scp=ALL) or find it on our 📆 [Flow Events & Working Groups](https://bit.ly/flow-events-calendar) calendar along with all the other working group meetings. Everyone is welcome 💚

turbolent:
Minor, but would be great to resolve: There are still a few differences between the EVM contract in the implementation and in the FLIP:
- FLIP has view annotations, missing in implementation
- Implementation has protectedAddress function, missing in FLIP
- Implementation passes new required domainSeparationTag argument, missing in FLIP
- Implementation has `timestamp` field in `EVMBlock`, missing in FLIP

@ramtinms @sideninja Could you please update FLIP (more important) and implementation to bring them in sync?

sideninja:
@turbolent I plan to, after I also complete this one https://github.com/onflow/flow-go/issues/5911

AlexHentschel:
## Summary of joint [Cadence & Execution](https://www.google.com/url?q=https://github.com/onflow/Flow-Working-Groups/tree/main/cadence_language_and_execution_working_group) and [Core Protocol](https://www.google.com/url?q=https://github.com/onflow/Flow-Working-Groups/tree/main/core_protocol_working_group) Working Groups on May 15, 2024
* Dieter recapped the motivation for EVM and its anticipated impact for Flow's ecosystem and broader adoption
* Ramtin summarized central technical decisions for Flow's EVM compatibility and integration into the Flow platform
* There were no concerns brought up. 

links to the video recording and meeting transcript [here](https://github.com/onflow/Flow-Working-Groups/blob/main/cadence_language_and_execution_working_group/meetings/2024-05-15.md#links-and-further-reading) (at the bottom)

### Next Steps

As there were no concerns in the meeting, we intend to close this **FLIP** as **approved/accepted** on **Wednesday May 22nd**, unless significant concerns are brought up here. Thank you.

sideninja:
@turbolent I updated all the recent changes in the FLIP. As far as the content goes I believe it should be up to date.

PR #229, opened 2023-12-15
FLIP proposal to expand Access Streaming API

Guitarheroua:
Draft proposal: Expanding the current Access Streaming API by introducing new subscription endpoints.


protocol/20231215-accessnode-streaming-api-expansion.md
```diff
@@ -0,0 +1,280 @@
...
+
+### SubscribeBlocksHeaders
+
+This endpoint enables users to subscribe to the streaming of block headers, commencing from a provided block ID or height. Additionally, users are required to specify the desired status of the streamed blocks to receive. The response for each block header should include the block's `Header`. This is a lighter version of `SubscribeBlocks` as it does not include the heavier `Payload` field.
+
+- **BlockHeight** (Block height of the streamed block)
+- **BlockId** (Block ID of the streamed block)
+- **BlockStatus** (`BlockStatusSealed` or `BlockStatusFinalized`)
+
+Usage example:
```
peterargue, line 101: it would be helpful to include the proposed response format for each new endpoint, as well as details about expected behavior. 

e.g. how do the block height/ids work in different situations (both included, neither)? When does `BlockStatus` not equal the status from the request?

protocol/20231215-accessnode-streaming-api-expansion.md
```diff
@@ -0,0 +1,280 @@
...
+
+- SubscribeBlocks
+- SubscribeBlocksHeaders
+- SubscribeBlocksLightweight
+- SendAndSubscribeTransactionStatuses
+- SubscribeAccountStatuses
+
+Additionally, it proposes an improved version of the REST WebSocket Subscription with a single connection.
+
+### WebSocket Subscriptions with a Single Connection
```
peterargue, line 33: I think this section needs some more detail. Alternatively, you could leave it out of this flip and include it in a more comprehensive flip with the new WebSockets design.

protocol/20231215-accessnode-streaming-api-expansion.md
```diff
@@ -0,0 +1,280 @@
...
+}
+```
+
+### SendAndSubscribeTransactionStatuses
+
+This endpoint enables users to send a transaction and immediately subscribe to its status changes. The status is streamed back until the block containing the transaction becomes sealed. Each response for the transaction status should include the `ID`, `Status` and `Error` fields.
+
+- **TransactionMsg** (transaction submitted to stream status)
+
+Usage example:
```
peterargue, line 186: This endpoint will need a sequence number in the response so the client can ensure they received all of the messages.

protocol/20231215-accessnode-streaming-api-expansion.md
```diff
@@ -0,0 +1,280 @@
...
+```
+
+### SubscribeAccountStatuses
+
+This endpoint enables users to subscribe to the streaming of account status changes. Each response for the account status should include the `Address` and the `Status` fields ([built-in account event types](https://developers.flow.com/build/basics/events#core-events)). In the future, this endpoint could potentially incorporate additional fields to provide supplementary data essential for tracking alongside the status.
+
+- **Address** (address of the account to stream status changes)
+- **Filter** (array of statuses matching the client’s filter. Any statuses that match at least one of the conditions are returned.)
+
+Usage example:
```
peterargue, line 230: this endpoint will also need a sequence number, as well as a heartbeat interval

protocol/20231215-accessnode-streaming-api-expansion.md
```diff
@@ -0,0 +1,280 @@
...
+        txResult.Error,
+    )
+}
+```
+
+### SubscribeAccountStatuses
+
+This endpoint enables users to subscribe to the streaming of account status changes. Each response for the account status should include the `Address` and the `Status` fields ([built-in account event types](https://developers.flow.com/build/basics/events#core-events)). In the future, this endpoint could potentially incorporate additional fields to provide supplementary data essential for tracking alongside the status.
+
+- **Address** (address of the account to stream status changes)
```
peterargue, line 227: I think this should also accept a start block

protocol/20231215-accessnode-streaming-api-expansion.md
```diff
@@ -0,0 +1,280 @@
...
+        resp.Address.String(),
+        resp.Status.String(),
+    )
+}
+
+```
+
+### Future Subscriptions
+
+The `SubscribeResourcesMovement` would enable users to subscribe to the streaming of account resource movement, while `SubscribeResourcesChanges` would allow users to subscribe to the streaming of account resource changes. As these functionalities necessitate additional research and discussions, they will be described and developed in the future.
```
peterargue, line 265: `SubscribeResourcesMovement` and `SubscribeResourcesChanges` could probably be combined into a single endpoint where movement is a type of change.

protocol/20231215-accessnode-streaming-api-expansion.md
```diff
@@ -0,0 +1,280 @@
...
+- An unsubscribe mechanism will be introduced. With multiple subscriptions on a single connection, simply closing this connection to halt updates becomes insufficient. Therefore, a dedicated endpoint for unsubscribing from undesired updates is necessary.
+- Each subscription response will now include a message type. For instance, the `SubscribeEvents` endpoint will result in a message with `type: "Events"`, while `SubscribeBlocksHeaders` will yield `type: "BlockHeader"` in messages, and so forth. This is essential to distinguish between various subscription results that will be delivered through a single connection.
+
+### SubscribeBlocks
+
+This endpoint enables users to subscribe to the streaming of blocks, commencing from a provided block ID or height. Additionally, users are required to specify the desired status of the streamed blocks to receive. Each block's response should include the `Block` containing `Header` and `Payload`.
+
+- **BlockHeight** (Block height of the streamed block)
+- **BlockId** (Block ID of the streamed block)
+- **BlockStatus** (`BlockStatusSealed` or `BlockStatusFinalized`)
```
peterargue, line 56: this should also support `full_block_response` similar to the `GetBlock` endpoint

protocol/20231215-accessnode-streaming-api-expansion.md
```diff
@@ -0,0 +1,280 @@
...
+
+Usage example:
+
+```go
+req := &access.SubscribeBlocksLightweightRequest{
+    // If no start block height or Id is provided, the latest block is used
+    StartBlockHeight: 1234,
+    // or StartBlockId: startBlockID[:],
+    BlockStatus: BlockStatus.BlockStatusFinalized // Use BlockStatusSealed to receive only finalized or only sealed blocks information
```
peterargue, line 152:
```suggestion
    BlockStatus: BlockStatus.BlockStatusFinalized // Returns only finalized blocks. Use BlockStatusSealed to receive only sealed blocks
```

peterargue (commented):
Thanks for the FLIP @Guitarheroua! Overall it's well outlined. I added a few comments where some more information would help.

bjartek:
Is it possible to have a streaming api that could give me sealed blocks with the following information:

 - Header: normal block header
 - Tx: the amount of tx in the block
 - HasSystemRveents: true/false


This would make the indexer in flowdiver a lot easier and more robust.

bjartek:
The flow of the "engine" i flowdiver atm is as follows:

-  GetTheBlock 
-  GetTransactionsByBlockID (we call flowkit GetTransactionByBlockID)
-  GetTransactionsResultByBlockID

Then i have to massage the return data to extract out systemChunkTransactionEvents. If I have any of those i enqueue a job to index them separately since we don't want this job to slow down the main queue. 

I am also very conscious to try to minimize the data that goes over the wire since it can be a lot, and with EVM stuff this will probably just increase.

peterargue:
I agree that we should keep this endpoint generalized. I think adding counts like txs and events is OK, but I don't think it would be practical to include them in a streaming block endpoints.

The block data only includes a list of collectionIDs, so the AN needs to first sync all of the collections from the network to count the number of tx. Generally this is done before the block is sealed, but not always. Event data is synced from execution nodes after the block is sealed. So they are never available until some time after the block is sealed (usually <1s but depends on the node's resources and location).

I think this type of data would work in an endpoint that notifies when data for a block is indexed. At that point, all of this data would be available. I haven't heard anyone else requesting it these, so it probably would come in a second iteration. We will need to add some endpoint(s) to expose this data anyway.

bjartek:
Knowing if it is empty or not is also useful.

bjartek:
In order to be 100% sure you got everything from a sub would it be possible to include some information in response about what the last message sent was? And have the ability to later get the missing data?

How would this architecture work on a flaky network for instance?

bjartek:
Ideally all. Just some way of knowing that you have gotten it all and can resume for where you left off safely.

peterargue:
> Ideally all. Just some way of knowing that you have gotten it all and can resume for where you left off safely.

this is what the "sequence number" comments I made were about. I think it's only relevant for endpoints that don't have data in every block (e.g. events, tx, accounts). With this included, a client could using this identifier to detect when they missed a message, and restart from the last good height if one was missed.

for example. if we included a messageNum field in the events stream. If you were streaming events and got a message with a messageNum that was greater than expected, you could restart the stream using the height from the last successful message as the start height.

protocol/20231215-accessnode-streaming-api-expansion.md
peterargue: Let's use `MessageIndex`. that aligns well with other fields like `EventIndex` and `TransactionIndex`

protocol/20231215-accessnode-streaming-api-expansion.md
```diff
@@ -0,0 +1,406 @@
+---
+status: draft 
+flip: 229
```
franklywatson, line 3:
```suggestion
flip: 244
```

protocol/20231215-accessnode-streaming-api-expansion.md
```diff
@@ -0,0 +1,406 @@
+---
+status: draft 
+flip: 244
+authors: Andrii Slisarchuk (andriyslisarchuk@gmail.com)
+sponsor: AN Expert (core-contributor@example.org)
```
franklywatson, line 5:
```suggestion
sponsor: Jerome Pimmel (jerome.pimmel@flowfoundation.org), Peter Argue (peter.argue@flowfoundation.org)
```

franklywatson:
Hey everyone. Will be closing this FLIP as accepted on Friday. I think we're down to the last of the feedback by now, but heads up in case of stragglers

bjartek:
Hey. Feel free to close if you are ok with current scope. 

It is still missing pieces that I need in order to use it. Ref streaming of events emitted in serviceGhunk transaction.

PR #235, opened 2024-01-03
Add initial draft of Flow EVM Gateway FLIP

m-Peter:


application/20240103-evm-gateway.md
```diff
@@ -0,0 +1,311 @@
...
+![Flow EVM Events](./2024-01-03-evm-gateway-resources/flow-evm-events.png)
+
+### Storage Integrity
+
+While consuming events, it is very critical to make sure that they are processed
+in a sequential manner, and no events response is missed for a single block.
+If this were to happen, the storage index could end up corrupted, and we would
+return incorrect data for things such as account balances, nonces etc.
+
+In the code snippet below, we specify a heartbeat interval of `1` and check
```
sideninja, line 235: I would say this would be good to change in the future as a performance improvement, ideally we would get multiple events and still preserve the order and integrity, however this currently might not be possible yet due to API not providing the heights, but it might be worth linking to an issue and keeping a promise to improve it.

application/20240103-evm-gateway.md
```diff
@@ -0,0 +1,311 @@
...
+start using it for development and testing.
+
+## Prior Art
+
+As explained above, the [Geth](https://github.com/ethereum/go-ethereum) client
+was the main source of inspiration for the work on the JSON-RPC server.
+
+## Questions and Discussion
+
+- Do we expect to have one official EVM Gateway run by the Flow team and
```
sideninja, line 309: We expect the community to maintain the EVM Gateways the same way as currently are for AN.

application/20240103-evm-gateway.md
```diff
@@ -0,0 +1,311 @@
...
+
+The main events from the `EVM` smart contract are two:
+- `flow.evm.BlockExecuted`,
+- `flow.evm.TransactionExecuted`
+
+To see the indexer & the events' payloads in action:
+
+![Flow EVM Events](./2024-01-03-evm-gateway-resources/flow-evm-events.png)
+
+### Storage Integrity
```
franklywatson, line 228: The integrity mechanism outlined below is a good start though we may need to build in more logic to recover missed events. While the goal is to not miss events we should plan for a reality where it will eventually happen.

application/20240103-evm-gateway.md
```diff
@@ -0,0 +1,311 @@
...
+
+### Tutorials and Examples
+
+A reference implementation addressing parts of this proposal can be found in
+this [PR](https://github.com/onflow/flow-evm-gateway/pull/11#issue-2006802251).
+
+### User Impact
+
+The Flow EVM Gateway is a standalone service that complements the FlowEVM
+functionality. It should be released at around the same time as FlowEVM,
```
franklywatson, line 292:
```suggestion
The Flow EVM Gateway is a standalone service that complements the Flow EVM
functionality. It should be released at around the same time as Flow EVM,
```

application/20240103-evm-gateway.md
m-Peter: Just to clarify, even with a heartbeat interval of `1`, we still get all the events emitted within the transactions of a single block. You are right though, I guess it will improve performance if we use a larger heartbeat interval when the backend introduces the sequence number field. Quoting the original message from Peter Argue in https://github.com/onflow/flow-evm-gateway/pull/11#discussion_r1431733617

> 
> The SubscribeEvents endpoint has a heartbeatInterval request parameter:
> https://github.com/onflow/flow/blob/b7ed2c3e452b5e44561399afc6b5e868e8957117/protobuf/flow/executiondata/executiondata.proto#L140-L148
> 
> This is used to set how often the backend will return a response message. If you set it to 1, it will return a response for every block. If the block has no events, the list is empty. If you check the block height from each message increases sequentially, you can verify that you received all messages, and thus all blocks were searched.
> 
> The API is currently missing an important sequence number field, which would let you get the same guarantee with larger heartbeatIntervals without receiving a response for every block.
>
> Given missing any events may corrupt your index, I'd recommend using the 1 block heartbeat for now.

application/20240103-evm-gateway.md
```diff
@@ -0,0 +1,311 @@
...
+	}
+}
+```
+
+### Performance Implications
+
+We should have some end-to-end tests, perhaps with the usage of the Flow Emulator.
+As well as some benchmarks, to assess the service degradation with regards to the
+increase of the storage, the incoming requests and possibly the emitted events.
+
```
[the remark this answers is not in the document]
m-Peter: Very good suggestion :clap:

application/20240103-evm-gateway.md
m-Peter: Sounds good, I guess I should introduce more flexible config options then.

application/20240103-evm-gateway.md
m-Peter: Yes, certainly. Back-filling of past events is something we should account for.

application/20240103-evm-gateway.md
```diff
@@ -0,0 +1,311 @@
+---
+status: proposed
+flip: [235](https://github.com/onflow/flips/pull/235)
```
franklywatson, line 3:
```suggestion
flip: [243]
```

franklywatson:
Marked this FLIP as proposed

application/20240103-evm-gateway.md
```diff
@@ -0,0 +1,311 @@
+---
+status: proposed
```
franklywatson, line 2:
```suggestion
status: implemented
```

franklywatson (approved):

application/20240103-evm-gateway.md
```diff
@@ -0,0 +1,311 @@
+---
+status: implemented
+flip: [243]
```
turbolent, line 3:
```suggestion
flip: 243
```

application/20240103-evm-gateway.md
```diff
@@ -0,0 +1,311 @@
...
+authors: Ardit Marku (markoupetr@gmail.com), Gregor Gololicic (gregor.gololicic@flowfoundation.org)
+sponsor: Jerome Pimmel (jerome.pimmel@dapperlabs.com)
+updated: 2024-01-03
+---
+
+# FLIP 235: Flow EVM Gateway
```
turbolent, line 9:
```suggestion
# FLIP 243: Flow EVM Gateway
```

turbolent:
This proposal has not been approved yet, so should have a proposed status, not implemented

PR #245, opened 2024-01-29
FLIP 245: Cadence Date and time

darkdrag00nv2:


darkdrag00nv2:
cc: @turbolent @SupunS @dsainati1

bluesign:
> It is assumed that there are 3600*24 seconds in every day and there are no leap years.

I think ignoring leap years is not a good idea.

turbolent:
> > It is assumed that there are 3600*24 seconds in every day and there are no leap years.
> 
> I think ignoring leap years is not a good idea.

Agreed, we can probably at least support leap years, like e.g. Java's `java.time.LocalDate` does

darkdrag00nv2:
> In general, we should not reinvent the wheel here and base the API and implementation of an established solution, like e.g. Java's JSR310

Yep, the current proposal is more based on the [python datetime module](https://docs.python.org/3/library/datetime.html#datetime-objects).

I wrongly made the assumption that they don't support leap years :slightly_frowning_face:. I was just giving the wrong input with the year 2011 :facepalm: 

```python
>>> datetime.datetime(2011, 2, 29)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: day is out of range for month
```

Trying with an actual leap year works fine.
```python
>>> datetime.datetime(2024, 2, 29)
```

KshitijChaudhary666:
Hi @darkdrag00nv2 - this FLIP is not reflected on FLIP project tracker. Did you follow the process outlined in https://github.com/onflow/flips? Specifically please remember to do the following without which the FLIP won't get visibility on the project tracker-

Create an [issue](https://github.com/onflow/flips/issues/new/choose) by using one of the FLIP issue templates based on the type of the FLIP - application, governance, cadence or protocol. The title of the issue should be the title of your FLIP, e.g., "Dynamic Inclusion fees". Submit the issue. Note the issue number that gets assigned. 
Then, create your FLIP as a pull request to this repository ([onflow/flips](https://github.com/onflow/flips)). Use the issue number generated in step 2 as the FLIP number. And mention the FLIP issue by copying the GitHub URL or the issue in the comment section.

Thank you!

cadence/20240115-cadence-date-time.md
```diff
@@ -0,0 +1,245 @@
+---
+status: draft
+flip: 245
+authors: darkdrag00n (darkdrag00n@proton.me)
+sponsor: Bastian Müller (bastian@dapperlabs.com)
```
turbolent, line 5:
```suggestion
sponsor: Bastian Müller (bastian.mueller@flowfoundation.org)
```

cadence/20240115-cadence-date-time.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+#### Constructor Functions
+Constructor functions will be defined to instantiate variables of type `LocalDateTime`. They are defined below.
+
+- `LocalDateTime.now()`: Creates a `LocalDateTime` with value based on the current block timestamp.
+
+- `LocalDateTime.fromTimestamp(timestamp : UFix64)`: Creates a `LocalDateTime` with value based on the passed timestamp argument.
+
+- `LocalDateTime.of(year: UInt64, month: Month, day: UInt8, hour: UInt8, minute: UInt8, second: UInt8)`: Creates a `LocalDateTime` with value based on the passed arguments assuming that there are 3600*24 seconds in every day.
```
turbolent, line 83: This can maybe just be a constructor function. It might be good to improve the naming of `day` and make it explicit that it is the day of the month, e.g. `dayOfMonth`:

```suggestion
- `LocalDateTime(year: UInt64, month: Month, dayOfMonth: UInt8, hour: UInt8, minute: UInt8, second: UInt8)`: Creates a `LocalDateTime` with value based on the passed arguments assuming that there are 3600*24 seconds in every day.
```

Examples below also need to be updated

cadence/20240115-cadence-date-time.md
```diff
@@ -0,0 +1,245 @@
...
+
+#### Public Members
+The `LocalDateTime` will provide the following public members:
+
+1. `getYear(): UInt32`: Get the year of the date-time.
+2. `getMonth(): Month`: Get the month of the date-time.
+3. `getDate(): UInt8`: Get the date of the date-time.
```
turbolent, line 90: Maybe improve the naming and clarify it is the day of the month, i.e.

```suggestion
3. `getDayOfMonth(): UInt8`: Get the day-of-month of the date-time.
```

Examples below also need to be updated

cadence/20240115-cadence-date-time.md
```diff
@@ -0,0 +1,245 @@
...
+## Design Proposal
+
+New types `LocalDateTime` & `Duration` will be added to Cadence. Both of these data types assume that there are 3600*24 seconds in every day and there are no leap years.
+
+### LocalDateTime
+
+The `LocalDateTime` type will be defined as follows:
+
+```cadence
+pub struct LocalDateTime {
```
turbolent, line 34: The proposal needs to be updated to Cadence 1.0. For example:
- `pub` -> `access(all)`
- Add `view` function modifier to getter functions

cadence/20240115-cadence-date-time.md
```diff
@@ -0,0 +1,245 @@
...
+
+Doing them manually is both hard and error-prone. As a result, standard library support similar to other modern languages would enhance the usability of the language.
+
+## User Benefit
+
+Manipulating date and time using raw timestamps is difficult and error-prone. Providing rich suite of standard library types and functions would increase the usability and user-experience.
+
+## Design Proposal
+
+New types `LocalDateTime` & `Duration` will be added to Cadence. Both of these data types assume that there are 3600*24 seconds in every day and there are no leap years.
```
turbolent, line 27: I wonder if we should add this functionality as a standard library contract, i.e. namespace it and not add them to the global namespace.

cadence/20240115-cadence-date-time.md
```diff
@@ -0,0 +1,245 @@
...
+3. 1 minute is converted to 60 seconds
+4. 1 milisecond is converted to 1000 microseconds
+
+#### Public Members
+The `Duration` will provide the following public members:
+
+- `getDays(): Int32`: Get the days of the `Duration`.
+- `getSeconds(): UInt32`: Get the second of the `Duration`.
+- `getMicroseconds(): UInt32`: Get the microseconds of the `Duration`.
+- `addYears(years : UInt32): Duration`: Return a new `Duration` after adding the provided number of years to it.
```
turbolent, line 166: Here and for the other functions below: The function name already clarifies what the parameter is, so the argument label can be removed:

```suggestion
- `addYears(_ years: UInt32): Duration`: Return a new `Duration` after adding the provided number of years to it.
```

turbolent, line 166: The `add` name sounds like it is mutating, but it seems like that isn't the case right?

Maybe we can use the `plus` and `minus` naming like in Java?

cadence/20240115-cadence-date-time.md
```diff
@@ -0,0 +1,245 @@
...
+let dateTimeFromTs = LocalDateTime.fromTimestamp(1415829132)
+dateTimeFromTs.getYear()      // 2014
+dateTimeFromTs.getMonth()     // november
+dateTimeFromTs.getDate()      // 12
+dateTimeFromTs.getHour()      // 21
+dateTimeFromTs.getMinute()    // 52
+dateTimeFromTs.getSecond()    // 12
+dateTimeFromTs.getDayOfWeek() // wednesday
+
+let dateTimeFromComponents = LocalDateTime.of(2019, april, 29, 19, 49, 31)
```
turbolent, line 119: Need to add argument labels, as required by the function declaration

turbolent (commented):
Looks good!

As we're finishing work on Cadence 1.0, we can finally come back to proposals like this. Thank you for your patience!

I left some comments and suggestions, and am happy to apply suggested edits and add remaining details. Does that sound good to you @darkdrag00nv2 ?

Once we addressed the outstanding issues, we can schedule to vote on it in the next WG call

turbolent:
We had a working group call yesterday and discussed this FLIP, see the notes in https://github.com/onflow/Flow-Working-Groups/blob/main/cadence_language_and_execution_working_group/meetings/2024-10-29.md#flips.

Consensus is that the proposed library would be useful, but is low priority. Luckily it can be implemented purely in Cadence, and does not need any knowledge of or required any contributions to the Cadence implementation.

@onflow/flow-develop-experience-tools @gregsantos: @dete proposed this could be a good item for the community and could e.g. be implemented through a grant, at a hackathon, etc.

PR #258, opened 2024-04-11
FLIP 258: NFT Storage Requirement

joshuahannan:
Adds a FLIP that proposes adding `access(contract) var ownedNFTs: @{UInt64: {NonFungibleToken.NFT}}` to the NFT standard `Collection`

PR: https://github.com/onflow/flow-nft/pull/211
[Discord Discussion](https://discord.com/channels/613813861610684416/621847426201944074/1227401046926692443)

### DIFF
--- a/application/20240411-nft-storage.md
+++ b/application/20240411-nft-storage.md
@@ -1,8 +1,8 @@
 ---
 status: Proposed 
-flip: 55
+flip: 258
 authors: Joshua Hannan (joshua.hannan@flowfoundation.org)
-         Cody Evaluate 
+         Cody Bouche (cody@evaluate.market)
 sponsor: Joshua Hannan (joshua.hannan@flowfoundation.org) 
 updated: 2024-03-11 
 ---
--- a/application/20240411-nft-storage.md
+++ b/application/20240411-nft-storage.md
@@ -1,5 +1,5 @@
 ---
-status: Proposed 
+status: Implemented
 flip: 258
 authors: Joshua Hannan (joshua.hannan@flowfoundation.org)
          Cody Bouche (cody@evaluate.market)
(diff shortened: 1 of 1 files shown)


application/20240411-nft-storage.md
```diff
@@ -0,0 +1,127 @@
+---
+status: Proposed 
+flip: 55
+authors: Joshua Hannan (joshua.hannan@flowfoundation.org)
+         Cody Evaluate
```
turbolent, line 5: Maybe use the real name here? Maybe also add an email

```suggestion
authors: Joshua Hannan (joshua.hannan@flowfoundation.org), Cody Bouche
```

joshuahannan:
After looking at the [currently staged contracts on testnet](https://testnet.flowview.app/account/0x2ceae959ed1a7e7a/storage), it looks like there are a good amount that still have `ownedNFTs` as `access(all)`. Not sure how much of a problem this is. We would need to tell them all to re-stage their contracts

cody-evaluate:
> After looking at the [currently staged contracts on testnet](https://testnet.flowview.app/account/0x2ceae959ed1a7e7a/storage), it looks like there are a good amount that still have `ownedNFTs` as `access(all)`. Not sure how much of a problem this is. We would need to tell them all to re-stage their contracts

fwiw, it dont think it will break anything because `access(all)` is a more derived/loose access level than the interface signature but Im not sure if the `access(contract)` on the `Collection` interface definition would gate access regardless of the `access(all)` concrete implementation or not (i would think it will but im not 100% sure)

joshuahannan:
> fwiw, it dont think it will break anything because access(all) is a more derived/loose access level than the interface signature but Im not sure if the access(contract) on the Collection interface definition would gate access regardless of the access(all) concrete implementation or not (maybe someone could just force cast to the concrete and withdraw but im not 100% sure).

Why is that true? I just changed the access modifier in `ExampleNFT` to `access(all)` but kept the one in `NonFungibleToken` as `access(contract)` and it looked like it worked. It seems like an interface specifying an access level should need to be followed and having a looser access level should not be allowed.

@cody-evaluate

joshuahannan:
I think you're right actually. So this won't be a breaking change. That is great news! I'll update the FLIP

application/20240411-nft-storage.md
```diff
@@ -0,0 +1,127 @@
...
+
+This also allows the standard to define more sophisticated functionality that
+interacts with stored NFTs, like we have done with adding the `forEachID()` function.
+In the future, there may be other functionality that is enabled
+by having standardized storage.
+
+## Drawbacks
+
+This change could be breaking for some contracts who
+have already done their Cadence 1.0 upgrades.
```
sisyphusSmiling, line 77: Is this still true?

sisyphusSmiling (approved):
Amazing work getting this out so quickly!

application/20240411-nft-storage.md
joshuahannan: no, I updated it

PR #266, opened 2024-05-03
Update 20230323-transaction-fee.md

KshitijChaudhary666:
Updating data

### DIFF
--- a/governance/20230323-transaction-fee.md
+++ b/governance/20230323-transaction-fee.md
@@ -17,23 +17,23 @@ Despite the trifurcation however, transaction fees on Flow continue to remain in
 
 **Low transaction fee poses threats of attack, necessitates continued token-issuance by the protocol (inflation), and makes FLOW use-cases seem peripheral vis-a-vis other networks.**
 
-In early days of the Flow network, when the number of transactions were fewer, transaction fees were intentionally kept low to encourage network usage. Today however, with Flow’s purpose-built compute platform and web3 utility for the masses driving a growing user-base (18M+ accounts) on the network consistently sending 6M+ transactions in an epoch (week), there is a need to revisit the fees.
+In early days of the Flow network, when the number of transactions were fewer, transaction fees were intentionally kept low to encourage network usage. Today however, with Flow’s purpose-built compute platform and web3 utility for the masses driving a growing user-base (30M+ accounts) on the network consistently sending 3-4M transactions in an epoch (week), there is a need to revisit the fees.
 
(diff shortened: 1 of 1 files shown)


governance/20230323-transaction-fee.md
```diff
@@ -71,44 +71,24 @@ Second, while this proposal by itself will only marginally reduce inflation (fro
...
+| Solana | $ 5.71E-02 | 4,891 times | $ 5.71E-02 | 49 times |
+| Flow | $ 1.17E-05 | N/A | $ 1.17E-03 | N/A |
 
 ## **Final Word**
 
 Transaction fees offer many benefits in Flow network’s economic design, including compensating the node operators for the resources deployed to process transactions and secure the network, protecting the network against spamming attacks, and generating long-term stability in the Flow economy via reduced inflation. In a mature state, transaction fees should allow the Flow blockchain to self-regulate transaction throughput in a way where it would always approach optimal throughput. This proposal however is meant to be a starting point for community discussions on how an increase in inclusion fees and unit cost of execution effort could lead to network-wide benefits including economically deterring network attacks and strengthening the overall FLOW economy. The community is invited to share feedback on this post and help shape the economics of FLOW.
 
-## **Appendix 1**
-
-**Comparison** **of** **Flow v/s Solana transaction fees (current scenario) for specific transaction types**
```
vishalchangrani, line 93: why remove?

KshitijChaudhary666: Removed this because first, we need to wait for Crescendo launch for real data from mainnet transactions broken down by type, and second, Solana's prioritization fee wasn't considered in the analysis as it's hard to average it out by transaction types. I can explain this more in our conversations with the community but including this table in the FLIP no longer makes sense as the data wouldn't be correct

governance/20230323-transaction-fee.md
```diff
@@ -71,44 +71,24 @@ Second, while this proposal by itself will only marginally reduce inflation (fro
...
+Third, this change would make FLOW’s use-cases more perceptible and established, further accentuating its stature as a token of value. The table below shows the average transaction fees of various networks vis-a-vis Flow in the current and proposed scenarios. Simply put, for every dollar that users would pay to transact on the Flow network, they currently pay $300 on NEAR, $4890 on Solana, and $270,000 on ETH - for the same transaction(s). This differential would reduce 100x with the approval and implementation of this proposal, yet even with the implementation of the new fee structure, Flow would continue to remain an economically attractive network to transact on. Note that given FLOW market price’s high correlation with the listed tokens’ prices (”R” averaging 0.8+), even a change in FLOW token price or stochastic market volatility would not substantially alter the competitiveness of the proposed Flow fees compared to other networks.
```
AlexHentschel, line 74: > Note that given FLOW market price’s high correlation with the listed tokens’ prices (”R” averaging 0.8+)

In scientific literature, it is generally discouraged to use symbols (such as `R`) without precise definition. While there are often commonly used symbols for some specific quantity, these conventions may change over time. Furthermore, given only the symbol, it is often highly non-trivial to figure out what specific quantity is meant. 

I assume with `R` you mean the Pearson's correlation coefficient? Furthermore, not sure what you mean with "averaging", because Pearson's correlation coefficient is usually defined for a sample. Do you mean you average over different tokens? I would just remove the "average". 

Suggestion:
```suggestion
Third, this change would make FLOW’s use-cases more perceptible and established, further accentuating its stature as a token of value. The table below shows the average transaction fees of various networks vis-a-vis Flow in the current and proposed scenarios. Simply put, for every dollar that users would pay to transact on the Flow network, they currently pay $300 on NEAR, $4890 on Solana, and $270,000 on ETH - for the same transaction(s). This differential would reduce 100x with the approval and implementation of this proposal, yet even with the implementation of the new fee structure, Flow would continue to remain an economically attractive network to transact on. Note that given FLOW market price’s high correlation with the listed tokens’ prices (Pearson's correlation coefficient of 0.8+), even a change in FLOW token price or stochastic market volatility would not substantially alter the competitiveness of the proposed Flow fees compared to other networks.
```

Issue #275, opened 2024-06-04
Type Removal In Contract Updates


dsainati1:

KshitijChaudhary666:
Hi @j1010001 - could you assign the DRI for this and update the status if this has moved?

SupunS:
I can take over this one

PR #276, opened 2024-06-04
FLIP 275: Removal of Types in Contract Updates

dsainati1:
Closes https://github.com/onflow/flips/issues/275 and https://github.com/onflow/cadence/issues/3210

This is a FLIP for a new pragma that would allow types to be removed in contract updates. 

As also mentioned in the FLIP itself, a prototype implementation was added in https://github.com/onflow/cadence/pull/3376 and https://github.com/onflow/cadence/pull/3380. This is currently disabled by default, and has been added only as a proof of concept and to be able to test in local environments.

### DIFF
--- a/cadence/20240604-cadence-type-removal.md
+++ b/cadence/20240604-cadence-type-removal.md
@@ -1,9 +1,9 @@
 ---
-status: proposed
+status: accepted
 flip: 275 
 authors: Daniel Sainati (daniel.sainati@flowfoundation.org)
 sponsor: Daniel Sainati (daniel.sainati@flowfoundation.org)
-updated: 2024-10-22
+updated: 2024-10-30
 ---
 
 # FLIP 275: Removal of Types in Contract Updates
--- a/cadence/20240604-cadence-type-removal.md
+++ b/cadence/20240604-cadence-type-removal.md
@@ -21,6 +21,10 @@ the ability to remove types from their contracts as they become outdated or othe
 The Cadence 1.0 upgrade in particular had exacerbated this need, as the changes to access control have
 rendered a number of types obsolete.
 
+However, [removing a type definition hasn't been possible/allowed](https://cadence-lang.org/docs/language/contract-updatability#invalid-changes-2)
+so far, since it can create type safety issues, by being able to re-add a different type definition with the same name later.
+So this motivated the necessity of having a way to tombstone a type definition.
+
 ## User Benefit
 
 This will allow users to remove deprecated or unnecessary type definitions from their contracts.
--- a/cadence/20240604-cadence-type-removal.md
+++ b/cadence/20240604-cadence-type-removal.md
@@ -1,0 +1,93 @@
+---
+status: draft 
+flip: 275 
+authors: Daniel Sainati (daniel.sainati@flowfoundation.org)
+sponsor: Daniel Sainati (daniel.sainati@flowfoundation.org)
+updated: 2024-06-04
+---
+
+# FLIP 275: Removal of Types in Contract Updates
+
+## Objective
+
+This adds a new pragma to Cadence that when processed by the contract update validator will
+allow a type to be removed from a contract, and also prevent a type with the same name from being
+added to that contract later. 
+
+## Motivation
+
+As originally outlined in https://github.com/onflow/cadence/issues/3210, users have been asking for
+the ability to remove types from their contracts as they become outdated or otherwise unnecessary. 
+The Cadence 1.0 upgrade in particular has exacerbated this need, as the changes to access control have
+rendered a number of types obselete. Additionally as all contracts on chain need to be updated for Crescendo, 
+the ability to remove dependencies on contracts that are difficult or otherwise challenging to update has become more important.
+
+## User Benefit
+
+This will allow users to remove deprecated or unnecessary type definitions from their contracts. In particular, 
+it unblocks the upgrade of the `FiatToken` contract on chain, a contract that numerous others depend upon. 
+
+## Design Proposal
+
+As implemented in https://github.com/onflow/cadence/pull/3376 and https://github.com/onflow/cadence/pull/3380, this feature introduces a new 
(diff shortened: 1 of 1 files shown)


turbolent (approved):

turbolent:
I guess this proposal does not prevent a `#replacedType` pragma being added in the future?

cadence/20240604-cadence-type-removal.md
```diff
@@ -0,0 +1,93 @@
...
+```cadence
+access(all) contract C {
+
+   #removedType(R)
+   
+    // ... contract body
+}
+```
+
+Placing the pragma at the top level, or nested inside another definition in `C` has no effect, and does not permit the removal of `R`.
```
SupunS, line 60: Could we make this an error/warning?

SupunS (approved):
👍

bluesign:
I see too many edge cases to be honest, maybe at least we can limit this to migration. I always suggested composeability to people, instead of copying to reuse ( not that it worked out much so far ), I feel this has opposite effect.

austinkline:
My main concern here is that permission to remove types effectively means a contract cannot revoke its keys if it has any dependency because of the risk that it could permanently break.

I'd like to more clearly understand the reason that this is needed beyond just "unblocking the FiatToken contract", what has to change that cannot be without this? Is there another path to doing it? 

Ideally, Circle should chime in here, maybe some of us can help them find another way.

bluesign:
> My main concern here is that permission to remove types effectively means a contract cannot revoke its keys if it has any dependency because of the risk that it could permanently break.

It is not only that, imagine having a field that using a struct with type that is removed. 

On Circle side, only explanation makes sense is they lost the keys.

austinkline:
> It is not only that, imagine having a field that using a struct with type that is removed.

Or an interface type definition that is the bedrock for other contracts downstream

bluesign:
copied from Discord to keep discussion in one place:

I think the main issue is you it feels like you are committed to this ( at least in FiatToken case ), unless some security issues comes out of this, I don't feel that it would be rejected.

austinkline:
Is this FLIP restricted only to concrete types or interface types as well?

If interface types as well, what is the difference between this FLIP and just allowing contract upgrades to remove conformance types?

dsainati1:
> Is this FLIP restricted only to concrete types or interface types as well?

In the design section of the FLIP it says that the current design proposal only allows removing concrete types, not interfaces.

bluesign:
btw if we stay out of Circle situation; and focus on the FLIP:

> The Cadence 1.0 upgrade in particular has exacerbated this need, as the changes to access control have
 rendered a number of types obselete.

I think would be nice to have some examples on this. 

> Additionally as all contracts on chain need to be updated for Crescendo, the ability to remove dependencies on contracts that are difficult or otherwise challenging to update has become more important.

This part I totally agree but maybe this can be relaxed with one time contract update checking for Crescendo. ( But also it is also tricky )

SupunS:
>I think would be nice to have some examples on this.

I remember running into some cases when updating nft/ft code bases to match C1.0, e.g: https://github.com/onflow/flow-nft/blob/f741cb5352f1889984e60437e0805b8371dcd7fe/contracts/ExampleNFT.cdc#L157C34-L157C41 where it would have been cleaner to be able to remove certain type definitions. In the above case it was an interface, and would have of-course needed some addition relaxations (e.g: being able to remove a conformance of a removed type) to make it really useful for that particular case. Perhaps `replacedType(T, R)` would have been a better option for that. But just wanted to highlight that there are cases that could benefit from a solution along these lines.
Similar case with the [NonFungibleToken](https://github.com/onflow/flow-nft/blob/f741cb5352f1889984e60437e0805b8371dcd7fe/contracts/NonFungibleToken.cdc#L167) as well.

austinkline:
> Similar case with the [NonFungibleToken](https://github.com/onflow/flow-nft/blob/f741cb5352f1889984e60437e0805b8371dcd7fe/contracts/NonFungibleToken.cdc#L167) as well.

This case is quite different fwiw. All existing marketplace contracts (including Flow's own NFTStandards) use `NFT.CollectionPublic` capabilities in their types, removal of this isn't an option like the ExampleNFT interface might be

bluesign:
thanks @SupunS , yeah the problem is it makes sense in the first look, but then it becomes tricky when someone else relies on it somewhere. for example like this: 

https://github.com/bluesign/mainnet-contracts/blob/cfcb76228edd5711d9892ce5cb99a178d5caaae5/contracts/0x4eded0de73020ca5/TopShotEscrowV3.cdc#L90

janezpodhostnik (approved):

bluesign:
> Concretely, we've thought of different "levels" for updatability: 

This is a great idea actually. I agree lack of macro view creates problems, to be fair, so far we always had micro level FLIPs. It is really hard to see bigger picture like this. I think this FLIP doesn't sound that bad when we read with that prespective. 

> A merged implementation does not mean that a decision has been made!

I don't know any project ( with a process to similar to FLIP process ) can claim that. But to be fair on this, I am not involved in much of anything else other than Cadence on this level.

turbolent:
To make it clearer that the feature has not been approved, the implementation of it is now behind a feature flag and disables by default: https://github.com/onflow/cadence/pull/3410

turbolent:
Thank you for updating the FLIP to the current, post-Crescendo, status @SupunS. The changes look great 👍

SupunS:
Thank you again everyone for taking time to review and provide feedback. As we discussed in the [last working group session](https://github.com/onflow/Flow-Working-Groups/blob/main/cadence_language_and_execution_working_group/meetings/2024-10-29.md), I'm going to consider this FLIP as accepted 🎉 

The concerns regarding breaking of downstream dependencies (from type-removals, and any other changes such as removing functions, removing fields, etc.) would be addressed separately as part of the discussion for "making contracts immutable", and a separate FLIP would be proposed in future.

PR #277, opened 2024-06-12
Add flip for changing import statement semantics

SupunS:
Work towards https://github.com/onflow/flips/issues/278

### DIFF
--- a/cadence/20240612-import-contract-as-reference.md
+++ b/cadence/20240612-import-contract-as-reference.md
@@ -1,0 +1,158 @@
+---
+status: proposed
+flip: 277
+authors: Supun Setunga (supun.setunga@flowfoundation.org)
+sponsor: Supun Setunga (supun.setunga@flowfoundation.org)
+updated: 2024-06-11
+---
+
+# FLIP 134: Import contracts as references
+
+## Objective
+
+Change the import statement semantics to import a reference to the contract, instead of importing the concrete value.
+Thus, accessing the fields of a contract externally would return references to fields, preventing any unintended external mutations.
+
+## Motivation
+
+A huge thanks goes to Deniz Mert Edincik (@bluesign) for pointing out the below security foot-gun in the existing design.
+
+In the [external mutability FLIP](https://github.com/onflow/flips/pull/89), the behaviour for accessing fields of a
+composite value was changed, to prevent external-mutation foot gun.
+You can read more about the overall vision [here](https://github.com/onflow/flips/blob/main/cadence/vision/mutability-restrictions.md).
+The essence of the change is that, if someone owns the composite value, they would have the full access (read/write/mutate) to its fields.
+Otherwise, if someone only got a reference to the composite value, then they can access the fields also as references,
+and the entitlements would kick-in to limit what they can do with those fields.
+
+Unfortunately, one edge-case we didn't realize at the time is that, contracts only have one value per network (it's a shared state),
+and by importing a contract, one could get the full access to that contract value.
+The imported contract value behaves like a reference, but is not represented using a reference in the language semantics. 
+Because of that, if the contract had an array/dictionary/composite -typed field defined as read-only (i.e: `access(all)`),
+then anyone could modify the content of that field, such as inserting/removing elements, etc.
+
+```cadence
+access(all) contract Foo {
+    access(all) var array : [Int] // Can't set a new value to the array, but can "update" the content (insert/remove/etc) 
+}
+```
+
+Here, the author of the contract may have intended it to be "read-only", however, it is not truly read only.
+
+```cadence
+import Foo from 0x1
+
+access(all) fun main() {
+    Foo.array[0] = 3
+    Foo.array.append(42)
+    Foo.array.remove(at: 0)
+}
+```
+
+This is exactly the type of foot-gun we tried to prevent using the external mutability improvements ([FLIP 89](https://github.com/onflow/flips/pull/89)).
+But unfortunately, given the contract import results in a non-reference value, the suggested changes in that FLIP
+does not get apply here.
+
+## User Benefit
+
+Prevent the foot-gun of unintended external mutations to read-only contracts fields.
+
+## Design Proposal
+
+The proposal is to change the import semantics to import contracts as a reference, instead of the concrete value.
+
+```cadence
+import Foo from 0x1
+
+access(all) fun main() {
+    var foo: &Foo = Foo  // The type of the imported contract value `Foo` will be `&Foo`
+}
+```
+
+Then, accessing a container typed field will also result in a reference (because of the changes added in [FLIP 89](https://github.com/onflow/flips/pull/89)),
+limiting the operation that can be performed on them.
+For example, mutating the `Foo.array` would result in an error.
+
+```cadence
+import Foo from 0x1
+
+access(all) fun main() {
+    var array: &[Int] = Foo.array  // Accessing field would also return a reference.
+
+    // All of the below operation would be statically rejected,
+    // since `Foo.array` would be an unauithorized reference.
+    Foo.array[0] = 3
+    Foo.array.append(42)
+    Foo.array.remove(at: 0)
+}
+```
+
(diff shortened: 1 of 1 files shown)


cadence/20240612-import-contract-as-reference.md
```diff
@@ -0,0 +1,156 @@
...
+
+### User Impact
+
+This would impact the Cadence 1.0 staged contracts. I ran an analyzer for the updated core contracts and the
+staged-contracts, and as of 12th June 2024, only three (3) contracts out of 456 are impacted by this change.
+The changes needed for those contracts are fairly simple, and would only require changing a couple of lines of code.
+
+## Related Issues
+
+None
```
dsainati1, line 147: This is obviously beyond the scope of this current FLIP (and would be a new feature that can be added after Cadence 1.0), but I think that changing imports to create an unentitled reference to the contract value does suggest that it may be useful to also be able to import an entitled reference to the contract as well. There would need to be some kind of way to guarantee that only certain people can do this, and syntax for it as well.

bluesign: I think easiest solution is to disallow entitlements on contract functions.

cadence/20240612-import-contract-as-reference.md
```diff
@@ -0,0 +1,158 @@
...
+
+In the [external mutability FLIP](https://github.com/onflow/flips/pull/89), the behaviour for accessing fields of a
+composite value was changed, to prevent external-mutation foot gun.
+You can read more about the overall vision [here](https://github.com/onflow/flips/blob/main/cadence/vision/mutability-restrictions.md).
+The essence of the change is that, if someone owns the composite value, they would have the full access (read/write/mutate) to its fields.
+Otherwise, if someone only got a reference to the composite value, then they can access the fields also as references,
+and the entitlements would kick-in to limit what they can do with those fields.
+
+Unfortunately, one edge-case we didn't realize at the time is that, contracts only have one value per network (it's a shared state),
+and by importing a contract, one could get the full access to that contract value.
```
turbolent, line 28:
```suggestion
Unfortunately, one edge-case that was not considered during the evaluation of the [external mutability improvements FLIP](https://github.com/onflow/flips/pull/89) is that contracts are singletons, and importing a contract provides owned access to that contract value.
```

cadence/20240612-import-contract-as-reference.md
```diff
@@ -0,0 +1,158 @@
...
+The imported contract value behaves like a reference, but is not represented using a reference in the language semantics. 
+Because of that, if the contract had an array/dictionary/composite -typed field defined as read-only (i.e: `access(all)`),
```
turbolent, line 30:
```suggestion
Because of that, if the contract has a container-typed field (e.g., array, dictionary, composite)  defined with public access (i.e., `access(all)`),
```

cadence/20240612-import-contract-as-reference.md
```diff
@@ -0,0 +1,158 @@
...
+import Foo from 0x1
+
+access(all) fun main() {
+    Foo.array[0] = 3
+    Foo.array.append(42)
+    Foo.array.remove(at: 0)
+}
+```
+
+This is exactly the type of foot-gun we tried to prevent using the external mutability improvements ([FLIP 89](https://github.com/onflow/flips/pull/89)).
```
turbolent, line 51:
```suggestion
This is exactly the type foot-gun that was intended to get prevented by the changes proposed in the [external mutability improvements FLIP](https://github.com/onflow/flips/pull/89).
```

cadence/20240612-import-contract-as-reference.md
```diff
@@ -0,0 +1,158 @@
...
+But unfortunately, given the contract import results in a non-reference value, the suggested changes in that FLIP
+does not get apply here.
```
turbolent, line 53:
```suggestion
do not apply here.
```

cadence/20240612-import-contract-as-reference.md
```diff
@@ -0,0 +1,158 @@
...
+
+## User Benefit
+
+Prevent the foot-gun of unintended external mutations to read-only contracts fields.
```
turbolent, line 57:
```suggestion
Prevent the foot-gun of unintended external mutations to read-only contract fields.
```

turbolent (commented):
LGTM! Just a few minor suggestions for phrasing

cadence/20240612-import-contract-as-reference.md
turbolent: I double checked this, and entitled contract members are already forbidden: https://github.com/onflow/cadence/blob/a1199d656c222a1253fdb362c20bd05f7c310287/runtime/tests/checker/entitlements_test.go#L1673-L1686

cadence/20240612-import-contract-as-reference.md
```diff
@@ -0,0 +1,157 @@
...
+        // Modifying fields within the contract is allowed.
+        Foo.array[0] = 3
+        Foo.array.append(42)
+        Foo.array.remove(at: 0)
+    }
+}            
+```
+### Drawbacks
+
+This is a breaking change. Certain users may have to update their code.
```
bjartek, line 114: So code that would need to be changed here is if you directly use a imported contracts fields and mutate them right?

cadence/20240612-import-contract-as-reference.md
```diff
@@ -0,0 +1,157 @@
...
+
+### Alternatives Considered
+
+A non-breaking alternative solution is to keep current language semantics as-is, and warn about contract fields that
```
bjartek, line 118: Personally I think this is very dangerous. Cadence-1.0 is already breaking so this is just one more thing that we need to do.

cadence/20240612-import-contract-as-reference.md
SupunS: yes, but in addition, if you use a field of a imported inside a function (say for example return as is), then the type annotation would change.
e.g:

```cadence
fun myFunction(): T {
    return ImportedContract.foo
}
```

then this would change to
```cadence
fun myFunction(): &T {  // <-- reference type
    return ImportedContract.foo
}
```
(This is also the change needed for those three contracts that breaks)

SupunS:
@bluesign Like mentioned in the impact section, there were three contracts affected by this, and all of them were returning a imported-contract field (a dictionary) as-is, so the type annotation of the function would needed to be changed. e.g: similar to https://github.com/onflow/flips/pull/277#discussion_r1643013008. I don't quite know what the use-case of those functions are; They were not used anywhere AFAIU (did a string search in all staged contracts, but came empty).

cadence/20240612-import-contract-as-reference.md
SupunS: Agree!

turbolent:
> I can think of `access(account) var` cases as most critical ( will require setter I suppose in the current model ) 

Yeah, that's one of the downsides, but IMHO probably not a big deal, given the workaround is adding a setter. Adding special rules to the language for treating `access(account)` differently is not worth it (more rules increase the burden for users and implementors, and there is little benefit over an explicit setter function)

bluesign:
> Yeah, that's one of the downsides, but IMHO probably not a big deal, given the workaround is adding a setter. 

Yeah I checked mainnet contracts now, it seems it is rarely used to modify, often used to just read.

bjartek:
LGTM, this is a no brainer.

austinkline:
Proposal is pretty straight-forward, a bit surprised it didn't work this way already!

:+1:

bluesign:
LGTM

SupunS:
For additional information, these are the fixes needed for the three breaking contracts: https://gist.github.com/SupunS/c732360df3fa969e09a161776a1af0ab

turbolent (approved):

SupunS:
Thanks everyone for the feedback! This FLIP has been open for a while, and had multiple breakout sessions for discussing the details. There has been no concerns and the feedback has been very positive.
Thus, I'll consider this FLIP as approved! 🎉 

You can follow the implementation progress at https://github.com/onflow/cadence/pull/3417

PR #283, opened 2024-07-27
FLIP 282: Import of pre-Cadence 1.0 Programs

turbolent:
FLIP #282

### DIFF
--- a/cadence/20240726-pre-cadence-1-import.md
+++ b/cadence/20240726-pre-cadence-1-import.md
@@ -1,0 +1,485 @@
+---
+status: proposed
+flip: 282
+authors: bastian.mueller@flowfoundation.org
+updated: 2024-07-26
+---
+
+# FLIP 282: Import of pre-Cadence 1.0 Programs
+
+## Objective
+
+The overall goal of this proposal is to allow Cadence 1.0 to import pre-1.0 programs.
+
+In particular, the goals of the proposed feature are to
+- Allow access to data (fields), not functionality (functions)
+- Allow programs to (keep) directly importing / depending on pre-1.0 programs
+- Allow programs indirectly working with values that have types defined in pre-1.0 programs
+  to (keep) functioning
+
+The proposal suggests a general feature of Cadence and does not specify the exact details
+of how a concrete implementation of this feature will achieve these goals.
+It is likely that achieving these goals in general is not possible,
+and an implementation will only support the import of certain pre-1.0 programs,
+and only certain parts of those programs.
+
+This proposal does NOT suggest adding a general means of importing any pre-1.0 program, fully.
+
+The proposed functionality can for example be used to keep some well-known programs,
+such as programs conforming to standards like the `FungibleToken` and `NonFungibleToken`,
+working to the degree where at least the fields required by the standards stay readable.
+
+## Motivation
+
+Cadence 1.0 contains many breaking changes.
+Pre-1.0 programs deployed to networks need to be updated to 1.0 to stay functional.
+However, might not get updated which renders them unusable:
+Both directly importing a pre-1.0 program
+and indirectly importing it by working with a value that has a type defined in a pre-1.0 program
+will very likely result in a parsing or type checking error,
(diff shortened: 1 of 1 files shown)


bluesign:
I strongly support this FLIP. 

One side question, I was gonna bring up yesterday on the meeting but time ran out: As we will have pre 1.0 parser in the codebase, does that mean that contracts can still update to 1.0 if they want later?

turbolent:
> As we will have pre 1.0 parser in the codebase, does that mean that contracts can still update to 1.0 if they want later?

At the moment this is not possible, because the state migration for Cadence 1.0/Crescendo is implemented as "all-at once", and does not support individual contracts to be updated later. It might be feasible, but we have not investigated how individual contracts could be updated later. Especially the migration of links to capability controller and the entitlements migration makes this problematic. 

We would definitely need the pre-1.0 parser in the codebase for the pre-1.0 to 1.0 contract updatability checking (currently done during the migration).

cadence/20240726-pre-cadence-1-import.md
```diff
@@ -0,0 +1,485 @@
...
+
+The proposed functionality can for example be used to keep some well-known programs,
+such as programs conforming to standards like the `FungibleToken` and `NonFungibleToken`,
+working to the degree where at least the fields required by the standards stay readable.
+
+## Motivation
+
+Cadence 1.0 contains many breaking changes.
+Pre-1.0 programs deployed to networks need to be updated to 1.0 to stay functional.
+However, might not get updated which renders them unusable:
```
joshuahannan, line 36:
```suggestion
However, they might not get updated which renders them unusable:
```

joshuahannan (commented):
I also strongly support this.

I think there is an argument to be made for still supporting deposit and withdraw. I need to consider it more, but it seems like if a project is willing to let their contract break with Cadence 1.0, the concern about potentially overwriting any custom withdraw/deposit functionality might be overblown. IMO, losing access to that functionality completely is worse than losing access to some small custom logic in those functions, which probably only applies to <1% of contracts anyway. 

Maybe if we are worried about overwriting custom withdraw/deposit functionality, we could do some sort of parsing of those functions to see if they just use the typical implementation and only overwrite them if they match the generic implementation.

SupunS (approved):
Nice write-up! I like the proposal, and particularly like the idea of this being a run-time recovery, rather than a one-off recovery during the migration. So the recovery can be improved later, and could support more features, and more contracts in future.

turbolent:
I've opened https://github.com/onflow/cadence/pull/3505, which implements one of the suggested solutions to how programs could determine if a type (of a value) is defined in a recovered program, e.g. to avoid function calls (which will always fail): Add a new field `isRecovered: Bool` to the Cadence type `Type` (run-time types).

If this seems like a good idea, I can add it to the proposal. cc @austinkline

j1010001:
Thanks everyone for the feedback! This FLIP has not received new comments in over 2 weeks, and had no updates since last breakout session. There has been no concerns and the feedback has been very positive, so we can consider this FLIP as approved! 🎉

turbolent:
@joshuahannan From a technical perspective that is definitely possible. In general, the proposal suggests to panic in all functions, as it is impossible to correctly convert pre-1.0 function implementations to 1.0.
However, we currently recover specific cases, i.e. FT and NFT implementations, so can probably implement certain functions, like `getIDs()` with a "sensible default". For `getIDs()` it looks like this would be `return self.ownedNFTs.keys`.

The current implementation of the recovered NFT contract can be found here:
https://github.com/onflow/flow-go/blob/661a4198def2f102f5b0aec8874cc539d5e0af46/fvm/environment/program_recovery.go#L132. We could just add `getIDs()` there.

Could you please open an issue in flow-go? PRs are of course also very welcome

joshuahannan:
Thank you! I'm glad that that should be possible. I created [an issue](https://github.com/onflow/flow-go/issues/6450) and I can try to work on a PR soon once I get done with my current priorities.

Issue #281, opened 2024-07-15
Integrating with the Lost and Found contract


nvdtf:
This FLIP will address better product integration with the Lost-and-Found contract to reduce the number of errors users might receive if the recipient is uninitialized or out of storage.

PR #286, opened 2024-08-07
FLIP 281: Integrating `LostAndFound` with Flow Port and Flow Wallet for Token Transfers

nvdtf:
FLIP #281

### DIFF
--- a/application/20240807-lost-and-found.md
+++ b/application/20240807-lost-and-found.md
@@ -1,0 +1,131 @@
+---
+status: draft
+flip: 281
+authors: Navid TehraniFar (navid.tehranifar@flowfoundation.org)
+sponsor: Greg Santos (greg.santos@flowfoundation.org)
+updated: 2024-08-07
+---
+
+# FLIP 281: Integrating `LostAndFound` with Flow Port and Flow Wallet for Token Transfers
+
+## Objective
+
+When a Flow user wants to send a token to another account, the recipient's account must be initialized to receive that type and have sufficient storage. If these conditions are not met, the token transfer transaction will fail.
+
+To address this issue, an intermediate contract can store the tokens until the recipient is ready to claim them. Austin Kline (@austinkline) from Flowty has developed a solution based on this concept called [LostAndFound](https://github.com/Flowtyio/lost-and-found). The sender can transfer tokens to the `LostAndFound` contract, and the recipient can claim them when they are online.
+
+This FLIP proposes integrating the `LostAndFound` contract into the Flow Port and Flow Wallet products to mitigate account initialization and storage errors without altering the Flow protocol or Cadence.
+
+## Motivation
+
+Token transfers fail if the recipient's account is not initialized or lacks sufficient storage. This situation requires the recipient to initialize their account before the transaction can complete, negatively impacting the user experience. Our goal is to enhance the token transfer process by providing a solution that allows tokens to be stored until the recipient is ready to claim.
+
+## User Benefit
+
+Integrating `LostAndFound` with Flow Port and Flow Wallet ensures that token transfers will not fail due to the recipient's account not being initialized. Senders can transfer tokens to the `LostAndFound` contract, and recipients can claim them when they are online.
+
+## Design Proposal
+
+We propose integrating the `LostAndFound` contract, already [deployed](https://contractbrowser.com/A.473d6a2c37eab5be.LostAndFound) on mainnet, with Flow Port and Flow Wallet. The contract uses a type-based ticketing system with support for fee estimation. More details are available in the project's [repository](https://github.com/Flowtyio/lost-and-found).
+
+#### Sending Tokens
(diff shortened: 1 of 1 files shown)


application/20240807-lost-and-found.md
```diff
@@ -0,0 +1,131 @@
...
+
+    The transaction proceeds as usual.
+
+- **Initialized but lacks storage**
+
+    The amount of extra storage and required FLOW tokens will be calculated and displayed to the sender. If the user confirms the transaction, the required FLOW tokens are also sent to the receiver. In this case the `FeeEstimator` contract will be used, but the tokens can be sent directly to the receiver.
+
+- **Not initialized**
+
+    The amount of needed storage and required FLOW tokens are calculated and displayed to the sender. Upon confirmation, the tokens and required FLOW for storage will be deposited into the `LostAndFound` contract. The receiver can choose to claim the tokens and FLOW when they are online.
```
austinkline, line 45: It is possible to also use a resource we call the `Depositor` for this in case you weren't aware:
https://github.com/Flowtyio/lost-and-found/blob/main/contracts/LostAndFound.cdc#L506

You can find sample transactions for its use here:
https://github.com/Flowtyio/lost-and-found/tree/main/transactions/depositor

The issue I've found with `FeeEstimator` is that every account on Flow is different. What works the account that does fee estimation might be too little to for another account that intends to use that estimation

application/20240807-lost-and-found.md
```diff
@@ -0,0 +1,131 @@
...
+
+When a user logs in to Flow Port or Flow Wallet the following will happen:
+
+1. A script will be executed to check if the user has any unclaimed tokens in the `LostAndFound` contract.
+
+2. If the user has unclaimed tokens, they will be shown in the new "inbox" section. In this screen, the user can:
+
+    - View a list of unclaimed tokens and the associated FLOW stored for each.
+
+    - Redeem one or multiple number of tokens. The user will be prompted to approve the transaction to claim the token(s) and FLOW.
```
austinkline, line 57: Not sure I follow here, the account that sent an item to the LostAndFound is who gets the FLOW used to store it once the item is redeemed

application/20240807-lost-and-found.md
```diff
@@ -0,0 +1,131 @@
...
+
+- This solution addresses token transfer issues at the application level and does not resolve the problem at the protocol level.
+
+- Different platforms can choose to deploy their own `LostAndFound` contract, which may lead to fragmentation.
+
+### Alternatives Considered
+
+#### Using child accounts
+
+Instead of a contract, the sender could create a child account for the recipient and transfer the tokens there. The recipient can later claim the account. However, the minimum storage balance cannot be claimed by the recipient as accounts cannot be destroyed on Flow, making this solution more costly for the sender and recipient.
```
austinkline, line 71: Probably worth mentioning here that the minimum cost for a new account is likely to increase (I think?) Which would drastically increase the minimum needed to send things using this method.

I recall doing benchmarks a very long time ago and found that you could send a few thousand NFTs (very basic ones) for 1 FLOW of storage, it should be the cheapest option by far

application/20240807-lost-and-found.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Performance Implications
+
+All the tokens are stored in a single universal `LostAndFound` contract. This can cause performance issues if the contract is used by a large number of users.
+
+### Dependencies
+
+This FLIP introduces the `LostAndFound` contract as a dependancy for the Flow Port and Flow Wallet products.
+
+It's recommended that the Flow Foundation co-own the `LostAndFound` contract, ensuring that changes are confirmed by multiple parties.
```
austinkline, line 85: Please bring this up with us offline.

application/20240807-lost-and-found.md
nvdtf: > The issue I've found with FeeEstimator is that every account on Flow is different. What works the account that does fee estimation might be too little to for another account that intends to use that estimation

Can you elaborate on whats the best way to estimate the total cost of interaction (deposit)? Can we use https://github.com/Flowtyio/lost-and-found/blob/main/scripts/lost-and-found/estimate_deposit_nft.cdc

application/20240807-lost-and-found.md
nvdtf: My bad, fixed it here https://github.com/onflow/flips/pull/286/commits/755a0cc826eb79292b4dc7961be6e79b5aba8fec

application/20240807-lost-and-found.md
nvdtf: Yeah, the storage balance is the deal breaker for this approach. Added a comment https://github.com/onflow/flips/pull/286/commits/af2cddcd8bc159d9e97977575df73c136227faaf

nvdtf:
We have a basic integration working live on port.flow.com now

Issue #288, opened 2024-09-23
Simple String Interpolation


RZhang05:
Provide support for the use of identifiers in string interpolation.

PR #289, opened 2024-09-23
FLIP 288: Simple String Interpolation

RZhang05:
Flip #288 addresses https://github.com/onflow/cadence/issues/3579.

### DIFF
--- a/cadence/20240923-simple-string-interpolation.md
+++ b/cadence/20240923-simple-string-interpolation.md
@@ -20,21 +20,23 @@ In general many languages support string interpolation for readability and ease-
 
 ## User Benefit
 
-Developers can avoid repeatedly using `concat` to generate strings which will simplify code and readability.
+Developers can avoid repeatedly using `concat` to generate strings which will simplify code and readability. Repeated concatenation is inefficient, as each concatenation results in an intermediate/temporary result, leading to an allocation and associated metering (computation and memory usage).
+
+In comparison, string interpolation can be implemented efficiently, and will thus also will lead to the user benefit of being less costly in terms of computation and memory usage.
 
 ## Design Proposal
 
-The main constraint is backward compatibility, existing Cadence 1.0 code cannot be affected. The proposed solution follows the syntax of swift as below:
+The proposed syntax for the string-literal with interpolation looks like follows:
 
 ```swift
 "Variable = \(someVar)"
 ```
 
-This change is backwards compatible because `\(` is not currently a valid escape character. 
+The main constraint is backward compatibility, existing Cadence 1.0 code cannot be affected. This change is backwards compatible because `\(` is not currently a valid escape character. 
 
 For this initial proposal there will be several limitations on `someVar`:
 - `someVar` will only be variable references with ability to be extended to expressions in the future
-- `someVar` must support the built-in function `toString()` meaning it must be either a String, Number, Address or Character.  
+- `someVar` must support the built-in function `toString()` meaning it must be either a `String`, `Number`, `Address`, `Character`, `Bool` or `Path`.  
 
 This is still useful for the first iteration since there are easy workarounds for these limitations such as extracting expressions into local variables.
 
@@ -44,14 +46,27 @@ None.
 
 ### Alternatives Considered
 
-String interpolation is preferred over a formatting function (such as `printf`) for the following reasons:
+String interpolation is preferred over a formatting function (such as `printf`) for two main reasons (1) string formatting is error prone and (2) there are many languages which moved from traditional formatting to string interpolation (e.g. Python).
 
-- Formatting is error prone (e.g. passing a `String` to `%d`)
-- There are many languages which moved from traditional formatting to string interpolation (e.g. Python)
+To talk more about (1):
+- The type of the argument passed to the formatting facility may not match the format specifier (e.g. value of type `String` is passed to format specifier for type Integer like `%d`)
+- Number of arguments passed to the formatter might not match the number of placeholders in the format specifier (e.g. too many or too few arguments)
+- When there are many placeholders in the format string, the order of arguments can get mixed up easily
+In most implementations of string formatting facilities, like when it is implemented as a normal function, these errors can not be detected statically, but only at run-time.
+
+However, some languages have formatting facilities that avoid some of the problems noted above, by statically checking the format specifier and the argument list. For example, Rust's `format!` macro does that.
+
+There are also many other implementations for string interpolation itself. For example:
+- Python has "f-strings" where an `f` prefix of a string-literal indicates interpolation with curly brace placeholders (e.g. `f"{someVar}"`)
+- JavaScript has backticks to indiciate string interpolation with curly brace placeholders (e.g. `` `{someVar}` ``)
+- Kotlin has string interpolation in normal string literals with optional curly brace placeholders with a leading `$` (e.g. `"$someVar and ${someVar2}"`)
+
+There exists no clear "standard". The proposed design leverages the fact that escaping is already supported `\` and parentheses naturally surround values in expressions. Therefore, no new syntax is introduced (e.g. no literal prefix, no placeholder indicator, no new expression surround characters).
+
 
 ### Performance Implications
 
-None, non-interpolated strings should not be affected.
+None, non-interpolated strings should not be affected. The proposed design of string interpolation has much better performance compared to repeated string concatenation.
 
 ### Dependencies
 
@@ -78,9 +93,11 @@ This is a feature addition, no impact.
 https://github.com/onflow/cadence/issues/3579
 
 ## Prior Art
-
+Existing instances in other languages
 - https://docs.swift.org/swift-book/documentation/the-swift-programming-language/stringsandcharacters/#String-Interpolation
 - https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals
+Some rationale and useful comparisons of string interpolation versus string formatting
+- https://peps.python.org/pep-0498/
 
 ## Questions and Discussion
 
--- a/cadence/20240923-simple-string-interpolation.md
+++ b/cadence/20240923-simple-string-interpolation.md
@@ -24,21 +24,19 @@ Developers can avoid repeatedly using `concat` to generate strings which will si
 
 ## Design Proposal
 
-There are many possible implementations two of which are highlighted below. The main constraint is backward compatibility, existing Cadence 1.0 code cannot be affected.
+The main constraint is backward compatibility, existing Cadence 1.0 code cannot be affected. The proposed solution follows the syntax of swift as below:
 
-### Python f-string syntax
-Support the following:
-```python
-f"value is {value}"
-```
-This change is backwards compatible as it introduces a new class of strings.
-
-### Swift \\()
-Support the following:
 ```swift
-"value is \(value)"
+"Variable = \(someVar)"
 ```
+
 This change is backwards compatible because `\(` is not currently a valid escape character. 
+
+For this initial proposal there will be several limitations on `someVar`:
+- `someVar` will only be variable references with ability to be extended to expressions in the future
+- `someVar` must support the built-in function `toString()` meaning it must be either a String, Number, Address or Character.  
+
+This is still useful for the first iteration since there are easy workarounds for these limitations such as extracting expressions into local variables.
 
 ### Drawbacks
 
@@ -77,7 +75,7 @@ This is a feature addition, no impact.
 
 ## Related Issues
 
-Extension to support expressions as opposed to just identifiers. Support custom `toString()` functions as well.
+https://github.com/onflow/cadence/issues/3579
 
 ## Prior Art
 
@@ -88,6 +86,6 @@ Extension to support expressions as opposed to just identifiers. Support custom 
 
 Feel free to discuss on the PR.
 
-Questions:
-- Preferences or concerns on proposed syntax
-- It may be less work to allow expressions versus enforcing identifier only, how much extra testing would be involved?
+## Implementation
+
+A POC is available at https://github.com/onflow/cadence/pull/3585.
--- a/cadence/20240923-simple-string-interpolation.md
+++ b/cadence/20240923-simple-string-interpolation.md
@@ -1,0 +1,93 @@
+---
+status: draft 
+flip: 288
+authors: Raymond Zhang (raymond.zhang@flowfoundation.org)
+sponsor: Supun Setunga (supun.setunga@flowfoundation.org) 
+updated: 2024-09-23
+---
+
+# FLIP 288: Simple String Interpolation
+
+## Objective
+
+This FLIP proposes adding support for simple string interpolation limited to identifiers only (no support for full expressions).
+
+## Motivation
+
+Currently Cadence has no support for string interpolation. It is convenient for developers to be able to inline variables in strings as opposed to the current solution of applying `concat` repeatedly. 
+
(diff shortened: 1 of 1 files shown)


cadence/20240923-simple-string-interpolation.md
```diff
@@ -0,0 +1,93 @@
...
+f"value is {value}"
+```
+This change is backwards compatible as it introduces a new class of strings.
+
+### Swift \\()
+Support the following:
+```swift
+"value is \(value)"
+```
+This change is backwards compatible because `\(` is not currently a valid escape character.
```
SupunS, line 41: I think this part can go to "prior art" section. Here in the "Design Proposal" section, we need to present the design that "we" are proposing. For example, which syntax are we going to use in cadence, and what expressions would be supported (only variable references for now, but mention that this is extensible for other expressions in future),  types of the expressions (i.e: only string-typed expressions for now, again, can/will be extended in future), etc.

cadence/20240923-simple-string-interpolation.md
```diff
@@ -0,0 +1,93 @@
...
+### Swift \\()
+Support the following:
+```swift
+"value is \(value)"
+```
+This change is backwards compatible because `\(` is not currently a valid escape character. 
+
+### Drawbacks
+
+None.
```
SupunS, line 45: One drawback is if someone already has the interpolation notation already (`${}`, `{}`, or which ever we pick) in one of their existing string literals, then they would have to update their code. This is highly unlikely, but in theory that makes it not completely backward compatible in that sense.

However, we can pre-analyze all the existing deployed contracts, both mainnet, and testnet, to see whether there are any such cases though. And if not, then all good.

cadence/20240923-simple-string-interpolation.md
RZhang05: Updated to choose a preferred design with more information on possible restrictions.

cadence/20240923-simple-string-interpolation.md
RZhang05: I believe this should not be possible with a properly chosen interpolation notation so it was not mentioned, but it can be added if the notation changes.

cadence/20240923-simple-string-interpolation.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Motivation
+
+Currently Cadence has no support for string interpolation. It is convenient for developers to be able to inline variables in strings as opposed to the current solution of applying `concat` repeatedly. 
+
+In general many languages support string interpolation for readability and ease-of-use. 
+
+## User Benefit
+
+Developers can avoid repeatedly using `concat` to generate strings which will simplify code and readability.
```
turbolent, line 23: Maybe also mention the performance improvement here: 

Repeated concatenation is inefficient, as each concatenation results in an intermediate/temporary result, leading to an allocation and associated metering (computation and memory usage).

In comparison, string interpolation can be implemented efficiently, and will thus also will lead to the user benefit of being less costly in terms of computation and memory usage.

cadence/20240923-simple-string-interpolation.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Design Proposal
+
+The main constraint is backward compatibility, existing Cadence 1.0 code cannot be affected. The proposed solution follows the syntax of swift as below:
```
turbolent, line 27: Maybe swap the order: First, describe what is being proposed, and then explain why the proposed design is a good idea (e.g. it satisfies backward-compatibility, and how)

cadence/20240923-simple-string-interpolation.md
SupunS: that makes sense

cadence/20240923-simple-string-interpolation.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Drawbacks
+
+None.
+
+### Alternatives Considered
+
+String interpolation is preferred over a formatting function (such as `printf`) for the following reasons:
+
+- Formatting is error prone (e.g. passing a `String` to `%d`)
```
turbolent, line 49: Maybe expand a bit on how string formatting is error prone.

For example: 
- Type of argument passed to formatting facility might not match format specifier (e.g. value of type `String` is passed to a format specifier for type `Integer`, like `%d`)
- Number of arguments passed to formatting might not match number of placeholders in format specifier (e.g. passed too many or too few more arguments)
- When there are many placeholders in the format string, the order of the arguments might get mixed up easily

In most implementations of string formatting facilities, like when it is implemented as a normal function, these errors can not be detected statically, but only at run-time.

However, some languages have formatting facilities that avoid some of the problems noted above, by statically checking the format specifier and the argument list. For example, Rust's `format!` macro does that.

Maybe also discuss alternatives considered for string interpolation.

For example:
- In some languages string interpolation is only performed in certain kinds of string literals. For example:
  - Python has "f-strings", where an `f` prefix of the string literal indicates interpolation (e.g.  `f"Hello, {name}"`). Placeholders are surrounded by curly braces
  - JavaScript has template strings/literals, where backticks indicate string interpolation (e.g. `` `Hello, {name}` ``). Placeholders are surrounded by curly braces
  - Scala has "s-strings", where an `s` prefix of the string literal indicates interpolation (e.g. `s"Hello, $name"`). Placeholders are indicated with a `$` and do not need to be surrounded
  - Ruby has string interpolation in normal string literals (e.g. `"Hello, #{name}"`). Placeholders are indicated by `#` and surrounded by curly braces

As can be seen, there's no clear "standard". The proposed design leverages the fact that escaping already exists `\`, and parentheses naturally surround values in expressions. Therefore, no new syntax is introduced (e.g. no literal prefix, no placeholder indicator, no new expression surrounding characters)

cadence/20240923-simple-string-interpolation.md
```diff
@@ -0,0 +1,91 @@
...
+- There are many languages which moved from traditional formatting to string interpolation (e.g. Python)
+
+### Performance Implications
+
+None, non-interpolated strings should not be affected.
```
turbolent, line 54: In addition to no negative performance implications on existing strings, maybe mention that the proposed design of string interpolation has a much better performance compared to repeated string concatenation.

cadence/20240923-simple-string-interpolation.md
SupunS, line 27: Don't need to say we took Swift's syntax as-is. It's more like we got inspiration from the syntaxes of existing languages, including Swift, and picked a one that is more suitable for Cadence.
```suggestion
The main constraint is backward compatibility, existing Cadence 1.0 code cannot be affected. The proposed syntax for the string-literal with interpolation looks like follows:
```

cadence/20240923-simple-string-interpolation.md
```diff
@@ -0,0 +1,91 @@
...
+This is a feature addition, no impact. 
+
+## Related Issues
+
+https://github.com/onflow/cadence/issues/3579
+
+## Prior Art
+
+- https://docs.swift.org/swift-book/documentation/the-swift-programming-language/stringsandcharacters/#String-Interpolation
+- https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals
```
turbolent, line 83: Maybe mention https://peps.python.org/pep-0498/ here, which has a rational and useful comparison to existing string formatting facilities.

cadence/20240923-simple-string-interpolation.md
```diff
@@ -0,0 +1,91 @@
...
+
+```swift
+"Variable = \(someVar)"
+```
+
+This change is backwards compatible because `\(` is not currently a valid escape character. 
+
+For this initial proposal there will be several limitations on `someVar`:
+- `someVar` will only be variable references with ability to be extended to expressions in the future
+- `someVar` must support the built-in function `toString()` meaning it must be either a String, Number, Address or Character.
```
turbolent, line 37: Maybe also add support for `Bool` and `Path`:

```suggestion
- `someVar` must support the built-in function `toString()` meaning it must be either a `String`, `Number`, `Address`, `Character`, `Bool`, or `Path`.
```

cadence/20240923-simple-string-interpolation.md
```diff
@@ -0,0 +1,91 @@
...
+The main constraint is backward compatibility, existing Cadence 1.0 code cannot be affected. The proposed solution follows the syntax of swift as below:
+
+```swift
+"Variable = \(someVar)"
+```
+
+This change is backwards compatible because `\(` is not currently a valid escape character. 
+
+For this initial proposal there will be several limitations on `someVar`:
+- `someVar` will only be variable references with ability to be extended to expressions in the future
```
SupunS, line 36: If we are going to extend this for other expression in future, would it be better to pick a notation and would less likely to conflict with other expressions. For example parenthesis are more commonly used in expressions (e.g: function calls, parenthesized expressions, etc.), than curly-braces (only in dictionary literal and inlined functions - both are rarely used). So if we put curly-braces as to mark the start/end of the interpolation (i.e: `"hello, \{name()}`), there's less chance it will collide with other expressions.

Regardless of which one is picked, the amount of work for the implementation would be the same. But most-common use-cases would be efficient, and less error-prone.

SupunS (approved):
Nice! This is a really convenient feature to have.

turbolent (commented):
Nice! This is going to be really useful

cadence/20240923-simple-string-interpolation.md
turbolent: How would the proposed notation conflict for more complex expressions? For example, an invocation expression like in your example would just be `"hello, \(name())"`. I don't quite see how a different set of delimiting characters, like e.g. curly braces, would help.

SupunS: So yeah, I don't quite know what the implementation would look like eventually, but in an expression like `hello, \(name())` or `\{{"name": "Supun"}}`, I assume that it may need to keep track of parenthesis/braces count or some parser-mode that recursively can distinguish between the closing parenthesis/brace of an expression vs the closing parenthesis/brace of the interpolation, and properly switch back to Lexing the remainder of the string. I'm not sure how accurately/efficiently this can be done. That's why I suggested to use parentheses instead.
Like I said, it doesn't solve the problem for nested parenthesis (only shift it to a separate place), but, If it turned out to be impossible to accurately distinguish the expression's curly-brace and the end of the interpolation, then we have the option to not allow curly-braces inside the interpolation at all. But that would still allow the most common use-cases, like function calls.

cadence/20240923-simple-string-interpolation.md
RZhang05: Done.

cadence/20240923-simple-string-interpolation.md
RZhang05: Added.

cadence/20240923-simple-string-interpolation.md
turbolent: Yeah, the lexer probably needs to keep track of the nesting level of whatever delimiter that is chosen (e.g. `(`) 👍

turbolent: Just as an aside: Here's how Swift lexes it: https://github.com/swiftlang/swift/blob/main/lib/Parse/Lexer.cpp#L1455

bjartek:
I really like the concept if this.

bluesign:
maybe little off topic but quick note on string interpolated strings vs printf; they are used for different needs; it is not one or the other. ( interpolation is better DX, but limited comparing to printf ( you cannot have dynamic templates etc ) )  So having both is a bit benefit. I think thinking with printf and building over it can be better option.

I am more into prefix or backtick to indicate interpolation personally, I think it is also easier to analyze statically. ( also by using a new prefix or backtick we can have multiline strings as additional benefit) 

Something like this:

```
var s = `hello, ${getUsername(userId)};
you have ${getUserBalance(userId)} tokens.
`

```

turbolent:
> maybe little off topic but quick note on string interpolated strings vs printf; they are used for different needs; it is not one or the other. ( interpolation is better DX, but limited comparing to printf ( you cannot have dynamic templates etc ) )

Regarding the limitation of not having dynamic templates: when do you ever _need_ to dynamically construct the template? From my experience, that is almost never the case.

We might still want to also provide a string formatting facility, but we should promote the safer alternative (string interpolation). Given the benefits of string interpolation, static checking, we should prioritize it over string formatting.

bluesign:
> Regarding the limitation of not having dynamic templates: when do you ever need to dynamically construct the template? From my experience, that is almost never the case.

Sorry it is a bit lack of my English; when I said dynamic templates, I meant format string stored somewhere and can be changed ( vs storing in code )  ( something like https://github.com/onflow/flow-evm-bridge/blob/main/cadence/contracts/templates/mainnet/EVMBridgedNFTTemplate.cdc )

sisyphusSmiling:
This would be a great feature!

Though I do agree with @bluesign that being able to use this in the context of stored strings would also be useful. As another example, the recent FlowRewards contracts ([`FlowRewardsMetadataViews#L106`](https://contractbrowser.com/A.a45ead1cf1ca9eda.FlowRewardsMetadataViews)) uses an SVG template stored within a struct, rendered by joining values with the contained string chunks.

It would be much easier to simply store the whole template (rather than splitting before storing) and render using native interpolation.

e.g. In practice, we currently need to so something like

```cadence
access(all)
struct Template {
  access(all) let chunks: [String]
  init(_ template: [String]) { self.chunks = template }
  
  access(all)
  fun render(param1: Int, param2: Int): String {
    return self.chunks[0].concat(param1.toString())
      .concat(self.chunks[2]).concat(param2.toString()
  }
}
```

Where it would be much nicer to do something like

```cadence
access(all)
struct Template {
  access(all) let template: String
  init(_ template: String) { self.template = template }
  
  access(all)
  fun render(param1: Int, param2: Int): String {
    // where template == "First parameter: \(param1) | Second parameter: \(param2)"
    return printf(`self.template`, param1, param2) // or some variation
  }
}
```

Though perhaps the above is a separate issue entirely or is at least not mutually exclusive with this FLIP as proposed.

cadence/20240923-simple-string-interpolation.md
```diff
@@ -0,0 +1,108 @@
...
+- `someVar` must support the built-in function `toString()` meaning it must be either a `String`, `Number`, `Address`, `Character`, `Bool` or `Path`.  
+
+This is still useful for the first iteration since there are easy workarounds for these limitations such as extracting expressions into local variables.
+
+### Drawbacks
+
+None.
```
bartolomej, line 45: Nit: One little drawback I see is that `\(` is a bit of an uncommon escape sequence for string interpolation. I haven't seen a language that uses that. I don't think this really matters, but I do feel like it's a bit easier to guess the correct syntax or scan the code if using a common syntax (which I'm not sure is possible given backward compatibility constraints as mentioned).

SupunS:
That's a good point on dynamic string formatting @bluesign and @sisyphusSmiling! Totally agree that it could be useful to have that as well. But at the same time, I think, as you pointed out as well, it could be another separate feature (can propose it as a new FLIP), and shouldn't really block us from having the static string templates/interpolations.

joshuahannan (approved):
This looks great! I'm excited to use this instead of concat!

RZhang05:
As discussed in the [last working group session](https://github.com/onflow/Flow-Working-Groups/blob/main/cadence_language_and_execution_working_group/meetings/2024-10-29.md), this FLIP has been accepted 🎉

joshuahannan:
Checking in on this again. Is this in the latest Cadence release and available on the latest emulator?

Issue #293, opened 2024-10-21
StructStringer interface


RZhang05:
Add a public interface which all string-convertible struct types will implement.

PR #294, opened 2024-10-21
FLIP 293: StructStringer interface

RZhang05:
Flip #293 working towards https://github.com/onflow/cadence/issues/3579.

### DIFF
--- a/cadence/20241010-stringer-interface.md
+++ b/cadence/20241010-stringer-interface.md
@@ -1,0 +1,104 @@
+---
+status: draft 
+flip: NNN (set to the issue number)
+authors: Raymond Zhang (raymond.zhang@flowfoundation.org)
+sponsor: AN Expert (core-contributor@example.org) 
+updated: 2024-10-10
+---
+
+# FLIP NNN: Stringer Interface
+
+## Objective
+
+This flip proposes the addition of a struct interface `StructStringer` which all string-convertible structs will implement. The goal of this FLIP is to simplify the code for converting `AnyStruct` to `String` while allowing for customizability. 
+
+## Motivation
+
+Currently if a developer wants to accept a value as `AnyStruct` and convert it to a string they have to identify which subtype the value belongs to and call the associated `toString` function. This leads to code such as the following
+```cadence
+access(all) fun toString(_ value: AnyStruct): String? {
+    if let stringValue = value as? String {
+        return stringValue
+    } else if let boolValue = value as? Bool {
+        return boolValue ? "true" : "false"
+    } else if let characterValue = value as? Character {
+        return characterValue.toString()
+    } else if let addressValue = value as? Address {
+        return addressValue.toString()
+    } else if let pathValue = value as? Path {
+        return pathValue.toString()
+    } else if let intValue = value as? Int {
+        return intValue.toString()
+    }
+    ...
+}
+```
+With the proposed addition, the code could be simplified to 
+```cadence
+access(all) fun toString(_ value: AnyStruct): String? {
+    if let stringerValue = value as ? {StructStringer} {
+        return stringerValue.toString()
+    }
+}
+```
+The same code above additionally allows for conforming `Struct`s to be converted to `String`s as well which was not previously generalizable.
+
+## User Benefit
+
+This will significantly streamline the process of converting `AnyStruct` values to `String` which is useful for developers. It also allows for much cleaner and simpler descriptions of `Struct`s for both developers and end users.
+
+## Design Proposal
+
+The proposed interface is as follows
+
+```cadence
+access(all) 
+struct interface StructStringer {
+    access(all)
+    view fun toString(): String
+}
+```
+
+### Drawbacks
+
+As with all user-defined types there are risks associated with calling someone else's `toString` function such as panic or gas usage concerns that developers need to be aware of. This means that unless it is a primitive type you will have to perform the necessary defensive checks. 
+
+### Alternatives Considered
+
+An alternative is to have `AnyStruct` itself implement a `toString` function. As a native function there is no longer any risk of malicious code and it still solves the problem of unnecessarily long `AnyStruct` to `String` conversions. The downside of this approach is in limiting customizability for developers. 
+
+### Performance Implications
+
+None.
+
+### Dependencies
+
+None.
+
+### Engineering Impact
+
+This change should be simple to implement.
+
+### Best Practices
+
+This should become the preferred way of converting `AnyStruct` to `String`.
+
+### Compatibility
+
+Proposed changes are backwards compatible.
+
+### User Impact
+
+Feature addition, no impact.
+
+## Related Issues
+
+string formatting
+
+## Prior Art
+
+Many languages have an interface for formatting types as `String` such as
+
+- [Display trait](https://doc.rust-lang.org/std/fmt/trait.Display.html) in Rust
+- `Stringer` in Go
+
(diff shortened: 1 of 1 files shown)


SupunS:
[Just cross-referencing] There were some concerns raised before, over having a user-defined `toString` functions for composites:
- https://github.com/onflow/cadence/issues/3579#issuecomment-2390004295
- https://github.com/onflow/cadence/issues/3579#issuecomment-2390880567

cadence/20241010-stringer-interface.md
```diff
@@ -0,0 +1,104 @@
...
+authors: Raymond Zhang (raymond.zhang@flowfoundation.org)
+sponsor: Supun Setunga (supun.setunga@flowfoundation.org) 
+updated: 2024-10-21
+---
+
+# FLIP 293: Stringer Interface
+
+## Objective
+
+This flip proposes the addition of a struct interface `StructStringer` which all string-convertible structs will implement. One goal of this FLIP is to simplify the process for representing `AnyStruct` as a `String`. Secondly, a customizable `toString` will be useful for future string formatting such as string interpolation with this interface.
```
turbolent, line 13:
```suggestion
This FLIP proposes the addition of a struct interface `StructStringer` to Cadence, which all string-convertible structs will implement. One goal of this FLIP is to simplify the process for representing `AnyStruct` as a `String`. Secondly, a customizable `toString` will be useful for future string formatting such as string interpolation with this interface.
```

cadence/20241010-stringer-interface.md
```diff
@@ -0,0 +1,104 @@
...
+    } else if let intValue = value as? Int {
+        return intValue.toString()
+    }
+    ...
+}
+```
+With the proposed addition, the code could be simplified to 
+```cadence
+access(all) fun toString(_ value: AnyStruct): String? {
+    if let stringerValue = value as ? {StructStringer} {
```
turbolent, line 39:
```suggestion
    if let stringerValue = value as? {StructStringer} {
```

cadence/20241010-stringer-interface.md
```diff
@@ -0,0 +1,104 @@
...
+        return stringerValue.toString()
+    }
+}
+```
+Additionally a conforming `Struct` can be converted to a `String` in the same function which was not previously possible. This can be useful for various string formatting functions.
```
turbolent, line 44:
```suggestion
Additionally a conforming struct can be converted to a `String` in the same function which was not previously possible. This can be useful for various string formatting functions.
```

cadence/20241010-stringer-interface.md
```diff
@@ -0,0 +1,104 @@
...
+}
+```
+
+### Drawbacks
+
+As with all user-defined types there are risks associated with calling someone else's `toString` function such as panic or gas usage concerns that developers need to be aware of.
+
+### Alternatives Considered
+
+An alternative is to have `AnyStruct` itself implement a `toString` function. As a native function there is no longer any risk of malicious code and it still solves the first goal. Additionally, it still allows for more powerful string formatting. The downside of this approach is in limiting customizability for developers.
```
turbolent, line 68: This might be infeasible, given that not all values of `AnyStruct` are stringable.

cadence/20241010-stringer-interface.md
```diff
@@ -0,0 +1,104 @@
...
+The proposed interface is as follows
+
+```cadence
+access(all) 
+struct interface StructStringer {
+    access(all)
+    view fun toString(): String
+}
+```
+
```
turbolent, line 61: Maybe add a note regarding naming, i.e. why the name of the interface has the prefix `Struct`

cadence/20241010-stringer-interface.md
```diff
@@ -0,0 +1,104 @@
...
+## Related Issues
+
+[string formatting](https://github.com/onflow/cadence/issues/3579)
+
+## Prior Art
+
+Many languages have an interface for formatting types as `String` such as
+
+- [Display trait](https://doc.rust-lang.org/std/fmt/trait.Display.html) in Rust
+- `Stringer` in Go
```
turbolent, line 103:
```suggestion
- [`Display` trait](https://doc.rust-lang.org/std/fmt/trait.Display.html) in Rust
- [`Stringer` interface](https://pkg.go.dev/fmt#Stringer) in Go
- [`CustomStringConvertible`](https://developer.apple.com/documentation/swift/customstringconvertible)  in Swift
```

turbolent (approved):
Nice! Adding a built-in interface, implementing it for built-in types that support it, and allowing Cadence developers to implement the interface for their types, makes a lot of sense 👍

SupunS (approved):
👍 from me

bluesign:
I had some concerns in the beginning but after thinking a bit more on it, this is great 👍 

- Is 'string' also 'Stringer'?
- should some standard library functions change 'string' parameters to 'Stringer'?

cadence/20241010-stringer-interface.md
RZhang05: Added in [85e9c20](https://github.com/onflow/flips/pull/294/commits/85e9c20c9c7627ad356bff8b5a9d803ae6a48105)

RZhang05:
> Is 'string' also 'Stringer'?

Yes it would be, this is the case in Swift with their [CustomStringConvertible](https://developer.apple.com/documentation/swift/customstringconvertible#conforming-types).

turbolent:
> should some standard library functions change 'string' parameters to 'Stringer'?

That should not be necessary and would just complicate and slow down the functions, because they would all have to first call `toString` on the argument, even though in most cases it would just be a string.

This is also uncommon in the standard library or in "user code" of other languages. For example, Go's standard library functions work with `string`s as inputs, not `fmt.Stringer`.

turbolent:
One thing I realized in the discussion above is that we should probably clarify what the result of the `toString` function should be. Not all, but quite a few other languages make the distinction between a "human-readable representation", and a "debug representation". The former is used for display purposes to show to the user using the application, while the latter is meant to be helpful for the developer of the application.

For example
- In Python there are the `__str__` and `__repr__` functions
- In Swift there are the `CustomStringConvertible` and `CustomDebugStringConvertible` interfaces, with `description` and `debugDescription` properties
- In Rust there are the `Display` and `Debug` traits

It might be good to clarify this in the proposal and note that the purpose of the proposed stringer interface is the former, providing a "human-readable representation".

I'm noting this, because currently we actually have the latter, the "debug representation", which makes sense for the `log` debugging function. For example, `log("foo")` logs `"foo"` (note the quotes), not `foo` (similar to how Swift `"foo".debugDescription` produces `"foo"`). 

When we're implementing this proposal we need to make sure to implement a new set of string functions and use it for the built-in types. Concretely, e.g. `interpreter.StringValue.String()` currently quotes. We might want to rename this e.g. `DebugString`, `Representation`, etc.

cadence/20241010-stringer-interface.md
```diff
@@ -0,0 +1,109 @@
+---
+status: draft 
+flip: 293
+authors: Raymond Zhang (raymond.zhang@flowfoundation.org)
+sponsor: Supun Setunga (supun.setunga@flowfoundation.org) 
+updated: 2024-10-21
+---
+
+# FLIP 293: Stringer Interface
```
turbolent, line 9: Maybe update the title to reflect the name change
```suggestion
# FLIP 293: StructStringer Interface
```

turbolent:
We discussed this FLIP in yesterdays [working group call](https://github.com/onflow/Flow-Working-Groups/blob/main/cadence_language_and_execution_working_group/meetings/2024-10-29.md#flips). Attendees liked the proposal and there were no concerns, so this FLIP is accepted 🎉

As always, thank you everyone for reviewing proposals and providing valuable feedback 🙏

PR #295, opened 2024-10-22
FLIP: Allow adding new fields to existing types

turbolent:
Moved from https://github.com/onflow/flow/pull/1097 

Please see the initial discussion there


cadence/20220817-add-fields-to-existing-definitions.md
SupunS, line 1: While at it, it's probably good to add a section, something like "Open Questions", mentioning that we still need to find a way to avoid old removed fields from getting re-added (a way to tombstone).

SupunS (commented):
👌

cadence/20220817-add-fields-to-existing-definitions.md
turbolent: Good idea! Added in [7012713](https://github.com/onflow/flips/pull/295/commits/7012713b04bbcdc6fe3a859f53bc72f9d58f57a4)

turbolent:
cc @austinkline @bluesign @bjartek

turbolent:
@austinkline We had a working group discussion about this FLIP yesterday, see https://github.com/onflow/Flow-Working-Groups/blob/main/cadence_language_and_execution_working_group/meetings/2024-10-29.md#flips. 

Feedback has been that it is a low priority feature, and given the complicated situation (existing stored state with "removed" fields) and the dependency on first having to prevent field removal (for both, see https://github.com/onflow/flips/pull/295/files#diff-6243eb63ad4bcc7c53d3b54c65552ad22149247ebef272dbb0413115b04e989aR104-R116), there is unfortunately quite a lot of effort required to implement this FLIP. 

Consensus has been that there are more important proposals to get in, and this proposal will be on hold, unless there is a strong reason to justify the effort. What do you think about this?

PR #296, opened 2024-10-31
Flip 298: Utilize Dynamic Protocol State for Version Beacon (coordinating upgrades of the Execution Stack) 

AlexHentschel:
This flip is tracked by issue https://github.com/onflow/flips/issues/298.

## Motivation

We want to move away from existing Version Beacon for the following reasons: 

1. Current Version Beacon requires that nodes have (potentially long) history (have seen version beacon service event, which is not guaranteed for nodes joining at epoch boundaries)
    
    Better: each block specifies which component version is to be used for processing it
    
2. Current Version Beacon based on height and hence not usable for upgrading most consensus-related aspects (any many other protocol aspects).
    
    Better: using View instead of height for triggering behaviour changes is generally applicable and more robust (view monotonously increases over time, while height might also decrease).

3. The Current Version Beacon does not sufficiently differentiate between a software version and specified behaviour of how the software should behave and makes very strong assumptions about downwards compatibility, which are too easily broken.

## Context
This flip is a draft resulting from the Core Protocol Working Group meeting on Oct 31, 2024. 
The original version was drafted in Notion: https://www.notion.so/flowfoundation/Core-Protocol-WG-Meeting-07-October-31-2024-1271aee1232480d0850eced4fafc1596?pvs=4


AlexHentschel:
P.S. As part of this PR, I have added the markdown [.md] extension to filenames that were previously missing it

protocol/20241031-execution-stack-versioning.md
```diff
@@ -0,0 +1,172 @@
...
+
+**Software Version** - The version identifier of a binary distribution of Flow Node software.
+By convention, we use semver-ish tag in Git and Docker releases.
+
+Software has bugs. and is frequently incomplete (e.g. API returning ‘not yet implemented’).
+The software version is a meaningful reference to describe what the software does in the real world. 
+
+However, we also desire a compact identifier [which we will call the ‘Component Version’] of how a Flow node *should* behave. 
+
+![VersioningConcepts.png](20241031-execution-stack-versioning/VersioningConcepts.png)
```
AlexHentschel, line 32: VersioningConcepts

protocol/20241031-execution-stack-versioning.md
```diff
@@ -0,0 +1,171 @@
...
+E.g. AN supporting script execution across HCU boundaries
+    
+    ❗Don’t couple the software version to the component version! We know there will be scenarios where we want one software to implement multiple
+    Component Versions and at that point, any one-to-one coupling of Software and Component Version will necessarily break. Instead, for each software
+    version, we conceptually have a _list_ of Component Versions that this software supports (even if that list only contains a single element most of the time).
+    
+
+### Reasons we want to move away from existing Version Beacon:
+
+Current Version Beacon:
```
turbolent, line 52: This proposal assumes knowledge of what the "version beacon" is. Maybe link to a description of it, or give a short introduction. It seems essential, because the essence of the proposal is to use an alternative solution.

protocol/20241031-execution-stack-versioning.md
```diff
@@ -0,0 +1,171 @@
...
+**We work towards transitioning the _existing_ Version Beacon to use the Protocol State.**
+
+- Decide now what convention we use for:
+    - Component Version format of Execution Stack? (e.g. $\texttt{major}.\texttt{minor}$ or single integer? Currently, the Version Beacon uses semver,
+      but Bastian thinks this might be unnecessary complex for Cadence)
+    - One Component Version for Cadence only?
+        
+        or one Component Version for Cadence+FVM combined?
+        
+        or two separate Component Versions (one for Cadence and one for FVM)
```
turbolent, line 147: We should definitely have a "Cadence Language" component that defines e.g. the behaviour of a Cadence program in terms of available functionality (like language features) and their behaviour.

In addition, we might also want additional components, like "storage format" (how are objects stored in account storage encoded?), "CCF format" and "JSON format" (how are "external" objects encoded?), and more, but IDK if we need those to begin with or can rather add them once needed.

@onflow/cadence What do you think about how Cadence as a whole should be "compartmenatilized" for versioning?

protocol/20241031-execution-stack-versioning.md
```diff
@@ -0,0 +1,171 @@
...
+
+![Illustration of Process](20241031-execution-stack-versioning/Execution_Stack_Versioning_(2).png)
+
+Example:
+
+![Overview](20241031-execution-stack-versioning/Execution_Stack_Versioning_(5).png)
+
+
+
+## Semantic versioning
```
turbolent, line 89: Does the proposal require the use of a semantic version? In the recent discussion it was mentioned that a component author may choose to use a semantic version for their component, but could also use e.g. a single version number. Sentiment was also that semantic versioning should not be used, but a single version number is sufficient and simpler.

Maybe make it clear exactly what is proposed as a recommendation and/or is prescribed

protocol/20241031-execution-stack-versioning.md
```diff
@@ -0,0 +1,171 @@
...
+ - mechanism for [scheduling version upgrades (at future view)](https://github.com/onflow/flow-go/blob/a6b157ce2770be9356e1cf35d1b0fff63f5e4a76/state/protocol/protocol_state/kvstore/upgrade_statemachine.go#L78-L142) exists
+ - mechanism [enforcing that node supports and uses correct](https://github.com/onflow/flow-go/blob/a6b157ce2770be9356e1cf35d1b0fff63f5e4a76/state/protocol/protocol_state/state/protocol_state.go#L235-L248) version as specified by the protocol
+
+# Roadmap: Dynamic Protocol State for coordinating Execution Stack upgrades (including Cadence changes)
+
+Biggest change (and possibly only significant change):
+
+- Dynamic Protocol State should ingest Version Beacon Service Event and track’s the Execution Stack’s Component Version
+
+![Illustration of Process](20241031-execution-stack-versioning/Execution_Stack_Versioning_(2).png)
```
turbolent, line 81: There is a *lot* going on in this diagram. Maybe add a description of what is important to this proposal.

Regarding this interface: maybe make this more abstract and explain it. I had a hard time understanding this core piece of the proposal. What is a "KVStoreReader"? What is a "ViewBasedActivator"? Maybe improve that naming and make it less based on current implementation details (Go/flow-go).

Maybe also make it clearer that the idea is that for each component, there will be two functions:
- A function returning the current required version for this component
- A function returning future/upcoming versioning for this component

It is not very clear that the example shows the versioning of the component named "ExectionStack".

turbolent, line 81: Not sure where this belongs, but: Like @zhangchiqing asked about/pointed out, the proposal only provides a mechanism to _retrieve_ the current component version. If the software version does not support this component version, the proposed mechanism does _not_ provide any means to let the component indicate it does not support it. In such a case, it is up to the component to decide how to proceed (e.g. fail execution, shut down, etc.)

protocol/20241031-execution-stack-versioning.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Do we want to track a Component Version for _every_ component of Flow? 
+
+**Answer**: For most components, we do _not_ track their version explicitly. Necessary updates are infrequent and not time sensitive, so that we
+can just bundle all changes across many components and ship them all together as part of a major upgrade (aka Spork).
+
+However, for very few components, upgrades are frequent and time sensitive (e.g. security fixes in Cadence), so that we cannot
+wait for the major upgrade (aka Spork). In that case, we want to deploy the upgrades into the running network and need to specify
+what should happen (i.e. the component version) and when that is going to change. Only for those components we want to track their component version
+via the protocol state.
```
turbolent, line 171: Maybe I missed it, but it would be great to describe how components are added to the system. For example, we might want to define just one or a couple components for the execution stack to begin with. How can/are additional components added?

protocol/20241031-execution-stack-versioning.md
```diff
@@ -0,0 +1,171 @@
...
+        
+        or one Component Version for Cadence+FVM combined?
+        
+        or two separate Component Versions (one for Cadence and one for FVM)  
+        
+- Start by including Component Version for Execution Stack into Dynamic Protocol State
+  
+  Breakdown of steps:
+  ![Execution Stack Versioning (3).png](20241031-execution-stack-versioning/Execution_Stack_Versioning_(3).png)
+
```
turbolent, line 153: Like in the [FLIP template](https://github.com/onflow/flips/blob/main/yyyymmdd-flip-template.md#best-practices), it would be great to have a "best practices" section on how to use this mechanism. I understand that such practices will be determined empirically, but some recommendations for the following topics might prevent mistakes, which could potentially lead to execution forks.
Recommending practices also ensures some consistency across components.

- During the discussion, the sentiment was that component versions should be decoupled from software versions
- With the introduction of new version, it is important to keep track and document them. Maybe recommend a light-weight process, like a changelog. This is not only important for contributors, but also important for node operators: As @zhangchiqing mentioned, a node might want to catch up and must ensure that it runs the appropriate software versions to catch up from the last block it processed to the latest block.
- Maybe recommend avoiding feature flags-based behaviour in code, and if needed, preferring version-based behaviour. It was also mentioned that this should ideally be only done temporarly, like e.g. for security fixes.

turbolent (commented):
Great proposal! +100 on adopting this for the execution stack and Cadence.

protocol/20241031-execution-stack-versioning.md
```diff
@@ -0,0 +1,171 @@
+---
+status: draft 
+flip: 296
+authors: Alex Hentschel (alex.hentschel@flowfoundation.org)
+sponsor: Jordan Schalm (jordan.schalm@flowfoundation.org), Yurii Oleksyshyn (yurii.oleksyshyn@flowfoundation.org)
+updated: 2024-10-31
+---
+
+
+# [FLIP 296] Utilize Dynamic Protocol State for Version Beacon (coordinating upgrades of the Execution Stack)
```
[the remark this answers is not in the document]
AlexHentschel: Sorry for my confusion. Appreciate your help with figuring out the convention (see also my suggestion for refining the readme about how to write a flip: https://github.com/onflow/flips/pull/297). 

Fixed in https://github.com/onflow/flips/pull/296/commits/b5f8620a59e472d9113d80edb1b466cdf4a6ca75

protocol/20241031-execution-stack-versioning.md
```diff
@@ -0,0 +1,171 @@
+---
+status: draft 
+flip: 296
```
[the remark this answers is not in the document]
AlexHentschel: Sorry for my confusion. Appreciate your help with figuring out the convention (see also my suggestion for refining the readme about how to write a flip: https://github.com/onflow/flips/pull/297). 

Fixed in https://github.com/onflow/flips/pull/296/commits/fd9c54dcd7345d9df20afdbfb9eba37edd011058

protocol/20241031-execution-stack-versioning.md
```diff
@@ -0,0 +1,171 @@
...
+authors: Alex Hentschel (alex.hentschel@flowfoundation.org)
+sponsor: Jordan Schalm (jordan.schalm@flowfoundation.org), Yurii Oleksyshyn (yurii.oleksyshyn@flowfoundation.org)
+updated: 2024-10-31
+---
+
+
+# [FLIP 296] Utilize Dynamic Protocol State for Version Beacon (coordinating upgrades of the Execution Stack) 
+
```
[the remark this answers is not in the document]
AlexHentschel: • added sections `Objectives` and `Motivation` and `User Benefit` according to Flip template

> Maybe mention that the existing solution (version beacon) has shortcomings, and the explicitly mention that the proposal is to use the new dynamic protocol state to coordinate the execution stack (the title already says that, but still)

• Clarified right at the beginning in the newly created `Objectives` section:
   **Direct goal [focus of this flip]**: transition the existing mechanism for upgrading Flow's execution stack to using Dynamic Protocol State**.

> Currently the section headers (like "Reasons we want to move away from existing Version Beacon", "dynamic Protocol State already implements better mechanism") already do so, but instead, make that the content (e.g. title: "benefits", content: limitations of current solution, and advantages of proposed solution)

* thanks for the suggestion. I put the limitations of current solution, and advantages of proposed solution into the `Motivation` section
* for user benefit, I briefly explained the benefits of having a robust and sufficiently general upgrade mechanism, which I hope clarifies why it makes sense to make the existing upgrade progress more robust and more generally applicable 


👉 commit https://github.com/onflow/flips/pull/296/commits/fd9c54dcd7345d9df20afdbfb9eba37edd011058 for details

protocol/20241031-execution-stack-versioning.md
AlexHentschel: Great suggestion. Thanks. 

In the newly- added `Motivation` section (specifically subsection `Status Quo`), I provide a short overview of the existing upgrade mechanism and explain the terminology `Version Beacon` and `HCU`

bluesign:
hmm I think semver is not a good idea here, actually I think version alone is bad idea tbh. I think this should be version +  an array of feature flags instead.

protocol/20241031-execution-stack-versioning.md
AlexHentschel: summarized your thoughts in section [`Outlook`](https://github.com/onflow/flips/blob/1a424d283061bb6377f774423f5dc369d65552f4/protocol/20241031-execution-stack-versioning.md#outlook). For now I would like to focus on the working of the _existing_ Version Beacon. Of course, if we have a clear understanding that we should update some versioning convention, we should do that ... but in lack of an alignment what to change (my impression at the moment), I think it is better to move ahead with the existing versioning convention rather than get blocked on missing alignment on how to do the versioning better (my impression of where we are at the moment).

turbolent:
@bluesign the idea is to keep the solution/mechanism simple. you could imagine that a particular component could translate a version to feature flags internally, so there isn't really any reason to put features into the "component version" that is tracked/coordinated by the protocol.

protocol/20241031-execution-stack-versioning.md
AlexHentschel: Thanks for the suggestion. I have restructured and significantly extended the versioning discussion.  Please see the new [section `Discussion of Possible Versioning Schemes`](https://github.com/onflow/flips/blob/3bf9ad02c4d52bdf226fbe2205719601c84d8544/protocol/20241031-execution-stack-versioning.md#discussion-of-possible-versioning-schemes-not-exhaustive). In specifically regarding your suggestion, I included the following at the beginning of this new section:

> At the moment, it is not sufficiently evident that there exists a _single_ versioning convention that is _easy and intuitive_ for _all_ the different components in Flow. Therefore, we recommend that **maintainers** for a specific component **decide** what component **versioning convention** works well for their component's particular upgrade pattern.

bluesign:
>  I would agree that certain features we would want to repeatedly turn off and on at runtime and in those cases feature flag make sense.

I was more of thinking of rollback scenarios without building new version. Also it can be helpful for stuff that deprecates. ( without releasing a new version ) Also it can make backwards guarantees easier, you can say that you guarantee component to be compatible with 1 previous version, and next one is previous + feature flags.

protocol/20241031-execution-stack-versioning.md
AlexHentschel: I have updated the flip to focus on Flow's execution stack
> * Direct goal [focus of this flip]: transition the existing mechanism for upgrading Flow's execution stack to using Dynamic Protocol State. 
> * Midterm goal: upgrades of the execution stack are by far the most frequent. By utilizing the Dynamic Protocol State for those upgrades, we want to generate learnings on where the new framework should be improved.

At the current state of the framework, I'd like to focus on gaining more experience with first applications before discussing details of adding new versions at scale. Nevertheless, I provide a short outlook on how new components are added: 

> :exclamation: For the time being, the Dynamic Protocol State must be updated each time we want to introduce a version for a previously unversioned protocol component. This is reasonably straight forward and contained engineering work, but not entirely free either. Furthermore, all versioning information also needs to pass through a smart contract and a specialized service event (this is already the case for the Execution Node's old Version Beacon), which at the moment are largely strictly typed (for high-assurance purposes). The scope of this flip is to gain more experience with first applications before discussing details of adding new versions in large numbers.

protocol/20241031-execution-stack-versioning.md
```diff
@@ -0,0 +1,222 @@
...
+\textnormal{Software Version :}\quad  \underbrace{\texttt{major}\,.\,\texttt{minor}\,}_{\textnormal{Component Version}}.\,\texttt{patch}
+$$
+
+- Software with identical  $\texttt{major}.\texttt{minor}$ is cross-compatible irrespective of patch version,
+  because they all implement the same specification (represented by the Component Version). Hence, the $\texttt{patch}$ version is not
+  part of the Component Version, because it does not influence the conceptual behaviour. Though, it represents implementation details,
+  so it is part of the Software Versio.
+- In addition, we introduce compatibility requirement from semantic versioning:
+    
+    $\textnormal{Component Version :} \quad \texttt{major}\,.\,\texttt{minor}$
```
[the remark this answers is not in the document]
AlexHentschel: > We need to coordinate a component version upgrade only when that component is upgraded in a backward-incompatible manner (major version increment in semver). Otherwise a rolling upgrade suffices.

Maybe I missunderstand - added new [section `Discussion of Possible Versioning Schemes`](https://github.com/onflow/flips/blob/3bf9ad02c4d52bdf226fbe2205719601c84d8544/protocol/20241031-execution-stack-versioning.md#discussion-of-possible-versioning-schemes-not-exhaustive), which in part discusses your point if I understood it correctly.

In a nutshell, a pure feature addition is still something that needs to be coordinated, because we need to agree _when_ the new feature becomes usable.

protocol/20241031-execution-stack-versioning.md
```diff
@@ -0,0 +1,393 @@
...
+introduce cutoffs, e.g. saying that the new Cadence implementation only supports Component Spec Versions `[1.7, 1.8, 1.9]`, then there is little to no difference to
+just using integer versions, e.g. the new Cadence implementation specifying that it Component Versions `[7, 8, 9]`.
+
+The only tangible benefit for SemVer I can see is for the Access Node. Here, a client could say that it wants to execute a script on the block with view 999.
+Transactions at this view were not allowed to use trigonometric functions. But the client may explicitly request a newer cadence version for its script that already
+supports trigonometric functions.  Nevertheless, this is a niche scenario with limited practical relevance. It is not clear whether this scenario warrants the
+additional complexity of SemVer with the associated correctness risks (SemVer assumes downwards compatability by default, while maintaining downwards compatability 
+in the implementation is generally additional work, so the default assumption of compatability induces additional risks for the happy path of block execution - no a good tradeoff in my opinion).
+
+### Versioning Scheme based on Feature Vectors
```
AlexHentschel, line 274: @bluesign I wanted to consolidate the conversation around feature flags into one thread here, which we can possibly resolve after we reach alignment and transferred the conclusions into the flip. This also allows other contributors to comment on the ideas without different conversations interleaving.

---
[@**bluesign**'s initial comment:](https://github.com/onflow/flips/pull/296#issuecomment-2455265692) hmm I think semver is not a good idea here, actually I think version alone is bad idea tbh. I think this should be version +  an array of feature flags instead. 

---
[@**AlexHentschel** reply](https://github.com/onflow/flips/pull/296#issuecomment-2456270283): I have extended [section `Discussion of Possible Versioning Schemes`](https://github.com/onflow/flips/blob/3bf9ad02c4d52bdf226fbe2205719601c84d8544/protocol/20241031-execution-stack-versioning.md#discussion-of-possible-versioning-schemes-not-exhaustive). It discusses Feature Vectors for versioning in detail. I would agree that certain features we would want to repeatedly turn off and on at runtime and in those cases feature flag make sense. 

In most cases we batch updates and for those I think an integer version would just be fine. So I agree with your general assessment of version + an array of feature flags instead. Though, at the moment I am not sure of any feature that we want to repeatedly turn on and off on mainnet. Hence, I'd start with the integer version alone. 

---

[@**bluesign**'s reply:](https://github.com/onflow/flips/pull/296#issuecomment-2456279680) I was more of thinking of rollback scenarios without building new version. Also it can be helpful for stuff that deprecates. ( without releasing a new version ) Also it can make backwards guarantees easier, you can say that you guarantee component to be compatible with 1 previous version, and next one is previous + feature flags. 

something like this: 
```
1.0 
1.0 + [feature_A, feature_B]
1.1 ( includes feature_A, feature_B) + [feature_C] 
```

AlexHentschel: Thanks for the clarification @bluesign. It makes sense to me to introduce a feature flag to potentially roll back one specific feature that we are worried about. However, we could also just roll back to the prior version if the software supports that:
```
• version 6 
• version 7 (extends version 1.0 by adding features A and B) 
  ⋮
• version 6: we learn feature A is bugged and roll back to version 6
• version 7: extends version 1.0 by adding features A and a fixed version of feature B) 
```
I would guess we share the opinion that many different could potentially be useful for different scenarios. Though, I have been asked repeatedly for "best practise guidelines" on what versioning scheme should be picked when -- which I am struggling with because I also don't think I have the necessary experience to give properly educated suggestions. 
That's mainly the reason why I tend to come back to the questions: 
1. Can we not express the same evolutionary path of the protocol by using solely integer versions? 
2. And what is the strong benefit of using a more sophisticated versioning scheme (here SemVer + feature flags)? Sure, it can be useful in specific cases, though a more sophisticated versioning schemes also introduces more intellectual and implementation complexity.

I agree that integer versioning is a comparatively simplistic tool. Though it is also extremely general as it makes effectively no assumptions about usage patterns.

turbolent: Interesting, I had not realized that this mechanism would support rollbacks 🤔 For whatever reason I assumed that versions would have to increment, even if an higher version effectively might just have the same behaviour as an earlier one, i.e. I had assumed that "version 6: we learn feature A is bugged and roll back to version 6" would rather be a "version 7: we learn feature A is bugged and roll back to the _behaviour_ of version 6, features A and B are disabled"

turbolent: > Can we not express the same evolutionary path of the protocol by using solely integer versions?
> [...] what is the strong benefit of using a more sophisticated versioning scheme 

That was mostly the realization in prior discussions: A simple integer is sufficient and any benefits (which?) are likely not worth the complexity

protocol/20241031-execution-stack-versioning.md
```diff
@@ -0,0 +1,222 @@
...
+For Flow, controlling complexity and codifying compatibility is essential.
+The **compatibility expectations, associated risks and additional complexity** of semantic versioning are not beneficial in *some* areas
+of the core protocol. For example the **Component Version** of the Protocol State is specified by a **single integer** (👉[code](https://github.com/onflow/flow-go/blob/c1a1cc0e05f0d323ab2f83dd5d74d8ad486d451e/state/protocol/kvstore.go#L30-L43)). 
+
+
+**Component Version** guidelines
+
+- Use semantic versioning for areas where
+  - benefit from backwards compatibility outweigh the implementation and complexity cost
+  - we want to maintain backwards compatability over a longer period of time and across multiple upgrades
```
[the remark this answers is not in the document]
AlexHentschel: thanks. Appreciate the comment. 

I tried to include your example and assessment into the [`Guidelines and Best Practices for Choosing a Versioning Scheme` section](https://github.com/onflow/flips/blob/cba3277b638d3fdc79a8ab29cd6132a8c0a2c680/protocol/20241031-execution-stack-versioning.md#guidelines-and-best-practices-for-choosing-a-versioning-scheme) (see commit https://github.com/onflow/flips/pull/296/commits/cba3277b638d3fdc79a8ab29cd6132a8c0a2c680). 

If you like the result, please resolve this comment. thanks

protocol/20241031-execution-stack-versioning.md
AlexHentschel: I have extended the [`Guidelines and Best Practices for Choosing a Versioning Scheme` section](https://github.com/onflow/flips/blob/240b378cd2bc2f36b98adf9c0fb6de82a5ea36ae/protocol/20241031-execution-stack-versioning.md#guidelines-and-best-practices-for-choosing-a-versioning-scheme) and explicitly covered this aspect:

> Note that the proposal only provides a mechanism to retrieve the current component version. If the software version does not support this component version, the proposed mechanism does not provide any work around. This is on purpose, because for the protocol to make progress, software needs to implement the respective protocol requirements. Being incompatible is a failure scenario. In such a case, it is up to the component to decide how to proceed (e.g. return an exception, crash, etc.)

Please resolve this conversation if you are happy with the result.

protocol/20241031-execution-stack-versioning.md
AlexHentschel: I have extended the [`Guidelines and Best Practices for Choosing a Versioning Scheme` section](https://github.com/onflow/flips/blob/cba3277b638d3fdc79a8ab29cd6132a8c0a2c680/protocol/20241031-execution-stack-versioning.md#guidelines-and-best-practices-for-choosing-a-versioning-scheme)

--- 
> • During the discussion, the sentiment was that component versions should be decoupled from software versions

I am not sure how to improve in the flip, because the flip already discusses Software version and Component versions as independent concepts and the relationship between both (section [`Terminology`](https://github.com/onflow/flips/blob/3a19aaa261aca0082b09fba5d902c2e519883d50/protocol/20241031-execution-stack-versioning.md#terminology)). Specifically, the following text is very clear:

> ❗Don’t couple the software version to the component version! We know there will be scenarios where we want one software to implement multiple Component Versions and at that point, any one-to-one coupling of Software and Component Version will necessarily break. Instead, for each software version, we conceptually have a list of Component Versions that this software supports (even if that list only contains a single element most of the time).

I would appreciate more specific advice on what to change, because at the moment, I am not sure. Nevertheless, I have re-iterated this point in the [`Guidelines and Best Practices for Choosing a Versioning Scheme` section](https://github.com/onflow/flips/blob/cba3277b638d3fdc79a8ab29cd6132a8c0a2c680/protocol/20241031-execution-stack-versioning.md#guidelines-and-best-practices-for-choosing-a-versioning-scheme)

> This FLIP makes no suggestions how to version software (on purpose). However, it is critically important to:
> * decouple component version from software versions 
> * A single component software can implement multiple component versions. Hence, each software must itself know which component versions it implements.

AlexHentschel: > With the introduction of new version, it is important to keep track and document them. Maybe recommend a light-weight process, like a changelog. This is not only important for contributors, but also important for node operators: As @zhangchiqing mentioned, a node might want to catch up and must ensure that it runs the appropriate software versions to catch up from the last block it processed to the latest block.

Included the following in the [Guidelines and Best Practices for Choosing a Versioning Scheme section](https://github.com/onflow/flips/blob/cba3277b638d3fdc79a8ab29cd6132a8c0a2c680/protocol/20241031-execution-stack-versioning.md#guidelines-and-best-practices-for-choosing-a-versioning-scheme):

> * With the introduction of new version, it is important to keep track and document them. For human readability, we recommend a light-weight process, like a changelog. However, the software must also provide algorithmic means to determine whether it can process a specific task according to specified component version.
>  - For a software, human-readable information about what component versions it support is important for node operators: The node operator needs to know which software version they have to bootstrap their node with (depending on the component versions specified by the blocks in the Sealing Segment).  If they bring up a node starting from a sealing segment far into the past, the node will catch up and might quickly learn that it requires yet another upgrade and then automatically stop/crash. It is helpful for the node operator to know this upfront to plan further software upgrades.
>  - Note that the human-readable versioning information is helpful for the node operator to do the right thing (use compatible software). But human-readable versioning information in insufficient to guarantee that node operators will do the right thing. Hence, the software must additionally automatically check that is supports the specified component version for every block it processes and all tasks associated with hat block.

AlexHentschel: > Maybe recommend avoiding feature flags-based behaviour in code, and if needed, preferring version-based behaviour. It was also mentioned that this should ideally be only done temporarly, like e.g. for security fixes.

The Flip now contains a detailed discussion of Feature Flags, discusses how the same could be achieved with Integer versioning and explicitly discourages use of Feature Flag vectors. For details, please see the [`Versioning Scheme based on Feature Vectors` section](https://github.com/onflow/flips/blob/cba3277b638d3fdc79a8ab29cd6132a8c0a2c680/protocol/20241031-execution-stack-versioning.md#versioning-scheme-based-on-feature-vectors).

protocol/20241031-execution-stack-versioning.md
bluesign: thanks @AlexHentschel , unfortunately I lack the experience in distributed protocols area, but from my experience in mobile game publishing, where deploying a new version is a bit delayed where many users are effected, it is always best to think with "bad version release" in mind. 

There are few parameters to consider I guess 

- do we need rollback vs new version release ? (btw I was thinking like @turbolent when I commented, component rollback will be a new version ) 

I don't know the cost of releasing new version to network, I assume ( from execution context only ) it involves updating and restarting the nodes. This can be covered with version ranges for sure, but I think there would need some guarantees on at least supporting 1 previous version in case things go wrong.

- what is the component support range should be? 

I think one version before is a must, just in case of a failure. version N supporting N-1 is probably enough. ( at least for security patches ) 

For bigger batches ( features bundled in a version ) it was the origin of the version + flags idea, as all those are parallel to previous version, linear rollback does not let us to remove one feature without removing another. ( not sure how important is this one also to the business case here ) 

It is really tricky subject, I don't think there is single best practice to follow, it all depends on targets and trade offs.

AlexHentschel: > always best to think with "bad version release" in mind

💯 agree. Especially for high-assurance systems such as flow, it's super important to have a fallback plan, because every software change requires time-intensive coordination and if mainnet were to break it would take on the order of hours to coordinate with a supermajority of node operators to deploy a fix. We thoroughly test, but of course that is not a guarantee that the software behaves as intended. 

I really appreciate this discussion, because it helps find a good balance: we want to make our toolset better, but it doesn't have to be perfect. Nevertheless, the improvement of our tooling need to be significant enough to warrant the respective investment of engineering resources. In the context of this Flip, the goal is to _extend_ our toolset for deploying software upgrades and addressing problems with upgrades:
* I personally think that already Integer Version via the Dynamic Protocol State for some (few) software components would be a significant step forward. We are going to learn a lot about for which upgrade scenarios our new approach is working well and for which scenarios there are challenges. Furthermore, integer versioning is simple and the most general, so it makes sense to me to start with that. 
* Nevertheless, @bluesign is raising very good points: What mechanisms do we need _in addition_, for integer versioning to be a _significant_ improvement? Paraphrasing (my understanding) of @bluesign's previous points:
  > (i) Should we recommend that software supports version N plus N-1, to keep the option for quick rollback without requiring time-intensive software upgrades on supermajority of nodes?
  > (ii) How should be represent rollback of new version in the Component Versioning Scheme?
  > (iii)  In case of releasing a bigger batch of features, do we want the ability to roll back select features without needing to reverse the entire upgrade as a whole?

   (Hope I understood correctly 😅, don't hesitate to jump in @bluesign)

For the points (i)-(iii), the software implementation as well as the component versioning scheme have to support it. 

_regarding (i):_
Whether we can make the implementation work (one software supporting component version [N-1, N]) will strongly depend on the feature we want to upgrade/add and the component it lives in. We should recommend support for component versions [N-1, N] but _not depend_ on it always being possible. I feel in many cases, it will be possible for the implementation to at least support component versions [N-1, N]. So for this flip, I think we should consider this scenarios explicitly. 


_regarding (ii):_ (rollback of new version as a whole)
I think for this scenario, I think integer versioning is sufficient. Lets walk over the two important sub-cases:

* The software only supports version `N`, but some time after deployment we discover there are problems with the new behaviour and we need to roll back. We need to deploy a new software to a majority of nodes, because the only behaviour the software supports is the bricked behaviour. Then, it doesn't really matter, whether we increment or decrement the version number for the following reason: 

   Example: Lets say we just deployed component version `N=6` and now find out there are severe problems. If we switch back to _exactly_ the same behaviour of version `5`, we could decrement the component version from 6 to 5 in the protocol state. The nodes with bricked behaviour will halt at the switchover point, the node operators would have to deploy the version-5-software and restart the nodes, which then proceed from the switchover point. Nevertheless, there is also no strict reason why two component specifications couldn't behave identically. In other words, we could specify version 7 to behave exactly as 5 and then we could increase the version number from the bricked version 6 to 7. The deployment process for the node software would be exactly the same.
* The other scenario is that our software supports component versions [N-1, N]; we just switched to version `N=6` and now find out about the problems. But differently to the prior case, our software still supports the old component version `5`.  Then we could quickly switch back to version 5 via a governance transaction without needing to touch the software on the nodes. In contrast, if we hypothetically chose to increase the version to 7 to remove the bricked behaviour, we would need new software. So decrementing the version number is clearly advantageous in my opinion. 

Hence, we conclude: decrementing the component version is generally suitable to rollback an upgrade as a whole. There are exceptions, but I would recommend this approach as the default. 

_regarding (iii):_ (rollback of new selective features of a new version)
So we have seen that the Integer Versioning Scheme works well for rolling back to a prior version (abandoning the new version as a whole). However, it doesn't allow us to roll back _selective features_ (arbitrary combinations) that were all deployed as one new version. I am not sure how prominent this scenario will be - my gut feeling is that we will badly need this only in a minority of cases. Hence, I am inclined to stick with Integer Versioning Scheme and think about _optional_ extensions for this edge-case scenario. 

### Here is my proposal: "Integer Versioning with Variable Feature Vector Extension"

PR #317, opened 2024-12-24
FLIP 316: FCL Ethereum Provider for Cross-VM Apps

### LINKED ISSUE #316: FCL Ethereum Provider for Cross-VM Apps
Issue for https://github.com/onflow/flips/pull/317

jribbink:
Closes #316

### DIFF
--- a/application/20241223-fcl-ethereum-provider.md
+++ b/application/20241223-fcl-ethereum-provider.md
@@ -1,9 +1,9 @@
 ---
-status: accepted
-flip: 316 (set to the issue number)
+status: released
+flip: 316
 authors: Jordan Ribbink (jordan.ribbink@flowfoundation.org), Chase Fleming (chase.fleming@flowfoundation.org)
 sponsor: Jordan Ribbink (jordan.ribbink@flowfoundation.org)
-updated: 2024-12-23
+updated: 2024-04-28
 ---
 
 # FCL Ethereum Provider for Cross-VM Apps


application/20241223-fcl-ethereum-provider.md
```diff
@@ -0,0 +1,281 @@
...
+
+// Will authenticate both FCL currentUser & COA Ethereum provider
+const evmAddress = await provider.request({ method: "eth_requestAccounts" })[0]
+
+// Single session shared between Ethereum JSON-RPC & FCL
+console.log("EVM Address:", evmAddress)
+console.log("Cadence Address:", (await fcl.currentUser().snapshot()).addr)
+
+// Send EVM TX from COA
+await signer.sendTransaction(...)
```
bluesign, line 113: s/signer/provider ?

application/20241223-fcl-ethereum-provider.md
```diff
@@ -0,0 +1,281 @@
...
+            execute {
+                let valueBalance = EVM.Balance(attoflow: 0)
+                valueBalance.setFLOW(flow: value)
+                let callResult = self.coa.call(
+                    to: self.evmAddress,
+                    data: calldata.decodeHex(),
+                    gasLimit: gasLimit,
+                    value: valueBalance
+                )
+                assert(callResult.status == EVM.Status.successful, message: "Call failed")
```
bluesign, line 169: I think here also fail case should be ok

application/20241223-fcl-ethereum-provider.md
```diff
@@ -0,0 +1,281 @@
...
+}
+```
+
+## Design Proposal
+
+### Overview
+
+FCL will expose an [EIP-1193](https://eips.ethereum.org/EIPS/eip-1193) Ethereum provider which can be used to interact with a Cadence-aware wallet’s COA via the Ethereum JSON-RPC API.  This provider will behave as a middleware between EVM tooling and FCL-JS by translating Ethereum JSON-RPC requests to interactions with an FCL wallet provider.
+
+It will be assumed that the user’s COA is stored at the path `/storage/coa`.
```
sisyphusSmiling, line 79: I assume this should be the standard COA storage path which is `/storage/evm`

```suggestion
It will be assumed that the user’s COA is stored at the path `/storage/evm`.
```

application/20241223-fcl-ethereum-provider.md
```diff
@@ -0,0 +1,281 @@
...
+console.log("Cadence Address:", (await fcl.currentUser().snapshot()).addr)
+
+// Send EVM TX from COA
+await signer.sendTransaction(...)
+
+// Send FCL Transaction from Cadence account
+fcl.mutate({ cadence: ... })
+```
+
+In practice, application developers will typically interact directly with the provider, but instead through higher-level tooling where it exists behind layers of abstraction.
```
sisyphusSmiling, line 119: Based on the second half of the sentence, I'm wondering if this is meant to read "...will not typically..."

application/20241223-fcl-ethereum-provider.md
```diff
@@ -0,0 +1,281 @@
...
+    Replace `<<CONTRACT_ADDRESS>>` with the appropriate contract address before using the script.  
+    
+    - Call COA Transaction
+        
+        ```tsx
+        import EVM from <<CONTRACT_ADDRESS>>
+        
+        /// Executes the calldata from the signer's COA
+        ///
+        transaction(evmContractAddressHex: String, calldata: String, gasLimit: UInt64, value: UFix64) {
```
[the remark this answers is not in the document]
bluesign: yeah UInt256 is better choice there

bluesign:
( I don't know much of EVM wallet flow, may be totally wrong here. )

If I understood correctly, eth provider is injected after FCL login. But I think injecting provider before can be better option. 

The flow I see on kitty punch, If I click login, it is popping up me a wallet selection dialog. Now if I add updated FCL I have to first login user with non-evm Flow Wallet to have provider in this list. ( I may be wrong on this assumption, but it seems provider is created on current user ) 

I think it may be better, if provider is in this list ( as we detect in discovery ) let's say BlueWallet ( hypothetical non-erm flow wallet extension ) and another provider maybe something like ( Other Flow Wallets ) which redirects to discovery. 

So when I click BlueWallet it can directly authenticate with FCL ( extension pop up ) and I can use it transparently. ( via just initializing FCL )

bluesign:
thanks @jribbink for this is great explanation. It all makes sense now, I think this is a great idea and very useful for filling the current gap.

bjartek (commented):
Question: how do we ensure that evm developers are aware that they can perform evm method calls and use that result in a cadence transaction without the need to hit javascript in the middle.


Atm that is the world they are used to. Do some calls to get data then do one tx and then maybe some more calls.

bjartek:
Nice progress. How do you incentivice users to create a flow wallet and not use metamask like they are used to for flow on evm apps.

I think a demo of this would be very useful. Like a fully functional prototype where some of the actions has to be done in cadence but the entry point can be done in both metamask and flow.

chasefleming (approved):

Issue #264, opened 2024-04-24
Passkey Support


JeffreyDoyle:
Passkey Support FLIP Issue Tracker

KshitijChaudhary666:
https://github.com/onflow/flips/pull/250

PR #320, opened 2025-02-27
FLIP 264: WebAuthn Credential Support in Flow

tarakby:
FLIP to support WebAuthn/FIDO2 credentials, including passkeys (https://github.com/onflow/flips/issues/264), for transactions signing in Flow.

It includes:
 - changes to the protocol to validate WebAuthn-generated signatures:
     - Access API
     - FVM
 - wallet guidelines to generate WebAuthn signatures for Flow transactions


protocol/20250203-webauthn-credential-support.md
```diff
@@ -0,0 +1,326 @@
...
+
+## Scenarios
+
+### Account public key registration
+
+- A wallet wants to add a new account public key to an existing Flow account, or create a new Flow account with a new public key.
+- The wallet initiates the process by calling [`authenticatorMakeCredential`](https://www.w3.org/TR/webauthn-3/#sctn-op-make-cred). In most platforms this is invoked via the higher level web API [`navigator.credentials.create()`](https://www.w3.org/TR/webauthn-3/#sctn-createCredential). This requests the authenticator to create a new key pair credential with the options specified in [`PublicKeyCredentialCreationOptions`](https://www.w3.org/TR/webauthn-3/#dictionary-makecredentialoptions):
+    - The relying party ID `rpID` is a string set by the wallet to identify the Flow credentials. This is a constant defined by the wallet for all Flow usage (for example `"FLOW-WEBAUTHN-MYWALLET-V0.0"`). Note that further map indices like [user handles](https://www.w3.org/TR/webauthn-3/#user-handle) and [credential IDs](https://www.w3.org/TR/webauthn-3/#credential-id) would allow the look up of the Flow private credential in the authenticator internal map.
```
bluesign, line 87: is this valid rpID ? `FLOW-WEBAUTHN-MYWALLET-V0.0` ?

protocol/20250203-webauthn-credential-support.md
```diff
@@ -0,0 +1,326 @@
...
+- Unlike the traditional WebAuthn registration process, the attestation signature is not verified on the server side. It is highly recommended to verify it on the wallet side, against the newly created public key and the attested message. This step is important to validate the public key before finalizing its registration on-chain. It serves as a proof that the authenticator possesses the private key corresponding to the public credential. 
+- If the validation succeeds, the wallet builds a new account public key using the newly generated public key, the signature algorithm returned by the authenticator, and the hash algorithm SHA2-256. 
+- The wallet includes the new account public key into a transaction to create a new Flow account with the key, or to add the new account key to an existing Flow account. The transaction is then submitted by the wallet.
+- The wallet stores the credential data required to look up the user private key in future calls to the authenticator. This may be a combination of the rpID, user handle and credential ID depending on the wallet implementation.
```
tombeckenham, line 95: It would be a lot nicer if the full credential was stored on-chain. Otherwise, we'd likely have to store it in a secondary db on the flow server (FRW-web-next). If it was stored locally, it'd defeat one of the key purposes of passkeys - device loss. You wouldn't be able to login without the rest of the credential.

tombeckenham (commented):
This is great work Tarak. I have a few questions about the rpId, and challenge. I'd also like to see the entire credential stored on chain. I also suspect we're going to need an intermediate server to handle passkey authentication. It's going to be useful on the client side to have the rpId match a specific server.

protocol/20250203-webauthn-credential-support.md
```diff
@@ -0,0 +1,326 @@
...
+- Once the user confirms registration through an authorization gesture, the authenticator creates the key pair and returns an [AuthenticatorAttestationResponse](https://www.w3.org/TR/webauthn-3/#iface-authenticatorattestationresponse) to the wallet. The response contains all the data needed to verify and register the new account public key.
+- The wallet extracts the new public key and the attestation signature from the authenticator response. It also constructs the [attested signed message](https://www.w3.org/TR/webauthn-3/#sctn-attestation) from the response and validates it contains the expected constant `challenge` set earlier. The wallet should also check the rest of the settings are set as expected in the attestation response such as the `rpIDHash` (should be hashed correctly from the `rpID`) and the other flags.
```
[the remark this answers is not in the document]
bluesign: I think each wallet will use their own RP ( I think it is important to not share auth between wallets )

protocol/20250203-webauthn-credential-support.md
```diff
@@ -0,0 +1,326 @@
...
+- The wallet extracts the new public key and the attestation signature from the authenticator response. It also constructs the [attested signed message](https://www.w3.org/TR/webauthn-3/#sctn-attestation) from the response and validates it contains the expected constant `challenge` set earlier. The wallet should also check the rest of the settings are set as expected in the attestation response such as the `rpIDHash` (should be hashed correctly from the `rpID`) and the other flags.  
+- Unlike the traditional WebAuthn registration process, the attestation signature is not verified on the server side. It is highly recommended to verify it on the wallet side, against the newly created public key and the attested message. This step is important to validate the public key before finalizing its registration on-chain. It serves as a proof that the authenticator possesses the private key corresponding to the public credential.
```
bluesign, line 92: > attestation signature is not verified on the server side.

I think here server side ( on webauthn context )  is wallet side already ( as RP will be on native side )

protocol/20250203-webauthn-credential-support.md
bluesign: I think if we have rpID per wallet , rest can be queried from passkeys

protocol/20250203-webauthn-credential-support.md
```diff
@@ -0,0 +1,326 @@
...
+
+Some fields in the signature extension data have arbitrary size and can be made purposely long in malicious transactions.
+Access nodes and collection nodes already implement a size check when ingesting transaction into the protocol and such malicious transactions won't pass the check.
+
+### RP ID hash and the Flow DST
+
+The signable message in the WebAuthn scheme begins with 32 bytes of `rpIdHash`, which is in honest cases, computed by hashing the `rpID`.
+The signable message in the plain scheme also begins with 32 bytes of the [Flow Domain Separation Tag](https://github.com/onflow/flow-go/blob/bc341a46060ab2ee6c6b23d4dc5a6d4a262a9931/model/flow/constants.go#L85) (DST) constant.
+The DST scopes the signable message to the Flow transaction context: `scoped_message = Flow_DST || payload`.
+It would be an issue if a malicious/careless wallet chooses `rpIdHash` to match `Flow_DST`.
```
bluesign, line 210: I am not sure what is the risk here.

protocol/20250203-webauthn-credential-support.md
```diff
@@ -0,0 +1,326 @@
...
+The other reason is that `collectedClientData` is submitted as part of the transaction data. Setting the challenge to the unhashed payload would almost double the transaction size for large payloads.
+
+### Replay attacks
+
+The original WebAuthn scheme uses two layers of protection against replay attacks: 
+- The random challenge-response mechanism: The server initiates the process by generating a random challenge and only authenticate the user when the challenge is answered (i.e. signed).
+- The signing counter: This counter is privately stored on the server to track the signature count per authenticator. The authenticator must provide a valid incremented counter along with every signature (included in `authenticatorData`). The server compares the provided count in a signature against the server stored one. This counter also prevents authenticator cloning across sessions, where the local counters can get out of sync with the server's stored counter. 
+
+The proposal does not require random challenges (the challenge is the payload hash) and does not store or check the authenticator's signing counts.
```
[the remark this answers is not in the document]
bluesign: I think here common situation is something like this: `Flow - [RP - Authenticator]` 

as RP and Authenticator here ( in native wallet case ) in the same domain, they don't need to do challenge as they don't have anyone to eavesdrop on. 

as for web-wallet case; I think you can still use the challenge ( but also it is a bit meaningless ) as authentication does not allow you to run any mutable action on Flow account.

protocol/20250203-webauthn-credential-support.md
```diff
@@ -0,0 +1,326 @@
...
+An internal structure of the field would need to be defined to encode the scheme identifier byte and the rest of the data. 
+This would break with seeing the `signature` field as purely the cryptographic signature and would mix different data in the same field. Since the structure already cleanly abstracts the `address` and `key_id` from the signature bytes, adding a new separate field `extension_data` was preferred.
+
+```protobuf
+message Signature {
+    bytes address = 1;
+    uint32 key_id = 2;
+    bytes signature = 3; // field would contain the cryptographic signature + the extension data
+}
+```
```
bluesign, line 294: yeah this was my initial idea ( almost ) something like adding a `signature_type` field and encapsulating structure in signature field ( signature + the extension data )

bluesign: even we can derive `signature_type` from maybe from account_key ( that also prevents webauthn keys to be used as flow keys )

protocol/20250203-webauthn-credential-support.md
lmcmz: Yes, it need to the a domain. In additional, the mobile app also need to be registered with this domain by adding `webcredentials` in `apple-app-site-association` or `assetlinks.json` file.

protocol/20250203-webauthn-credential-support.md
lmcmz: Yeah, wallet need a database to store the UserId and publicKey mapping. Otherwise, wallet won't able to retrieve the public key again on local. Otherwise, if we can store extra data per key, it will be helpful. For example, while user register a new key in the flow address, we can also store userId or sth to lookup sth on chain.

protocol/20250203-webauthn-credential-support.md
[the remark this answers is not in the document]
lmcmz: There are 2 types of challenges, one is during key creation, another after key creation using signing tx. 

For the first one, the wallet server need store the publicKey and userInfo mapping, it need send challenge which can be a random hash to let wallet to sign. After wallet signed and sent it back, it verified the signature with publicKey and store the mapping and then create flow address for this user.

For the second one, we hash flow transaction payload into challenge and use it to let passkey to sign. And then using extensionData to verify the signature.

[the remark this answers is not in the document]
bluesign: I think here context is only for first one; I think what @tarakby means is: you don't need challenge, because there is no replay risk.

protocol/20250203-webauthn-credential-support.md
bluesign: > Wallet won't able to retrieve the public key again on local after creation. 
I don't think this is true. Assertion is signing with that private key afaik. 

not sure about the best UX; but what I imagined was more like: you login , select account, credential ID is temporarily saved, account is looked up with corresponding public key from chain, future signings are done with that credential id.

lmcmz: Correct me if I'm wrong. From my understanding, the credentialID and public key mapping is not on local, it need saved on wallet server. 
The passkey's public key only return in creation under `credential.create`, after that everything is using `credential.get` which won't contain the public key. Hence, if wallet lost the mapping between credential Id and public key or address, it's a risk of loss assets. 

Although, wallet have the private key but it can't find the associated flow account because it can't get the public key from the private key again.

bluesign: I can be wrong too, but I think `credentialID` should be saved temporarily like username. ( remember username stuff we see on web2) 

@tarakby knows better but as assertion ( credentials.get ) will give you signature ( which you can verify with message : authenticatorData || sha256(clientJson) ) can't you just extract public key from those ?

bluesign: here I made a quick and dirty example: https://gist.github.com/bluesign/425720a5d5afbf637b5b5469a98da83a

lmcmz: Nice work, let me double check it.

protocol/20250203-webauthn-credential-support.md
bluesign: quick note here: ( I just learned ) also domain has to reference application identifier with some file ( depending on the platform I guess) in  .well-known directory.

protocol/20250203-webauthn-credential-support.md
```diff
@@ -0,0 +1,326 @@
...
+
+The proposal does not require random challenges (the challenge is the payload hash) and does not store or check the authenticator's signing counts. 
+Instead, Flow relies on the [proposal key sequence number](https://developers.flow.com/build/basics/transactions#sequence-numbers) to prevent replay attacks. 
+The sequence number tracks the number of signatures per account public key, it is stored on-chain and FVM expects it to be incremented in each signed transaction (in a similar logic to WebAuthn's signing counts). 
+Flow sequence numbers are not effective against authenticator cloning because the expected sequence number is publicly stored on-chain.
+A cloned authenticator can look up the next valid sequence number before signing.
+
+### Server side attestation verification
+
+Unlike WebAuthn, Flow does not require a signature from the newly registered account key.
```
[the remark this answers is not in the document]
bluesign: nit: does webauthn require signature from newly registered account key ? I think in attestation context, device private key signs the attestation; so technically verifying with device public key should prove public key is valid and came from the authenticator.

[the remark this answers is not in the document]
tarakby: > Does this mean that technically we could add an account key that is invalid?

It's not possible to register an invalid public key on chain. The chain (specifically FVM) checks that the key is an elliptic curve public key corresponding to the signature algorithm indicated by the transaction.

What I'm trying to say in the FLIP is that when a transaction adds a new public key `B` to an existing account on-chain, the transaction is signed by the another account key `A`. `B` does not need to sign anything. `B` is registered successfully on chain as long as it has a valid cryptographic format and the signature against `A` is valid. This means that `B` can be registered, even when its corresponding private key isn't known to the end user.
I took adding an account key as an example, but the same applies to creating a new account. This can be risky of course but this is how FVM currently works (nothing to do with this FLIP).

> does webauthn require signature from newly registered account key ?

Webauthn registration of a new public key requires the a signature under the new key. The specs call it "attestation signature" (while the login signature is called assertion signature). The authenticator returns such attestation to the client, and the server verifies it before registering the key. 
Since in our case Flow chain does not natively verify this signature like webauthn servers, the client is encouraged do the extra step of verifying the attestation in order to protect its users (from unknown private keys).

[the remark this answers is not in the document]
bluesign: I think it is only the case in less secure `self-attestation`. https://www.w3.org/TR/webauthn-2/#self-attestation

protocol/20250203-webauthn-credential-support.md
tarakby: Attaching the scheme to the account key was also considered, it's mentioned in https://github.com/onflow/flips/blob/tarak/webauthn/protocol/20250203-webauthn-credential-support.md#encode-the-scheme-within-account-public-keys as an alternative.

> that also prevents webauthn keys to be used as flow keys

Why do you think this is important? 
In particular, isn't this a limitation rather than a protection?

protocol/20250203-webauthn-credential-support.md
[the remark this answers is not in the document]
tarakby: I updated the attestation verification parts in https://github.com/onflow/flips/pull/320/commits/56ced9862dc82b4feafa8603f6537ab10262c1d2

protocol/20250203-webauthn-credential-support.md
tarakby: I think the update of the attestation verification part fixes this, but let me know if there is still confusion in the text.

protocol/20250203-webauthn-credential-support.md
tarakby: I'll try to detail the risk:
- a user signs a webauthn transaction using a careless wallet. The wallet happens to use an `rpIdHash` equal to `Flow_DST` (the wallet skips hashing an `rpId`). The signed message would be `Flow_DST || other_webauthn_data`. The transaction is submitted to the chain and passes verification. 
- a malicious party reads that transaction and extracts the signature. It builds a plain-scheme transaction (non-webauthn) such that `Flow_DST || other_webauthn_data` would be the signable payload of the transaction, i.e. "other_webauthn_data" matches the payload encoding. 
- The malicious party submits the non-webauthn transaction by copying the signature from the webauthn transaction. The newly formed transaction has a different intent than the original webauthn transaction, but the user got tricked.

It's a highly unlikely risk but it's still possible. 

> I think only possible protection here is adding key with some flag to the account ( as webauthn key ) and disallow normal transaction signing with it.

Yes that's one way to do it, but I don't think it's the only one.

> but also they will also need to reverse sha256 the Flow_DST

Yes, that's the case where the wallet honestly and builds `rpIdHash` from `rpId`. From the chain point of view, that is not guaranteed, the chain only sees 32 bytes of `rpIdHash` within the transaction and can't verify it is indeed a hash.

protocol/20250203-webauthn-credential-support.md
```diff
@@ -0,0 +1,326 @@
...
+Losing the user's device often means losing access to their private keys.
+
+FIDO2 is already implemented by major platforms and operating systems. These implementations use passkeys as FIDO2 credentials, which are backed by secure hardware and are very convenient to use.
+On certain systems, passkey credentials are shared between the user's devices for easier access, and are securely backed up for later recovery in cases of loss.
+Wallets based on passkeys would resolve the current self-custody flaws. 
+
+# User Benefit
+
+Flow account users would be able to use self-custody wallets based on passkeys from major platforms (Apple's Password app, Google's Password Manager, etc.). 
+They would be able to register account public keys from their devices and authorize transaction signing via the device's biometric authentication.
```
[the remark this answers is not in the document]
tarakby: Good point!

I mentioned it in the following sentence but I'm happy to emphasize it further: 

> This combination provides strong security through hardware elements (no phishing, protection at rest and in use) and high usability through the portability between devices and recovery features in cases of loss.

protocol/20250203-webauthn-credential-support.md
bluesign: I don't think limitation;  if you think that keys are secure in authenticator ( which should be ) and it is not possible to use them to sign regular flow transactions anyway ( otherwise we would not need this flip ) 


btw my gut feeling is still, it is not possible to collude there but in any case it is also valuable information to have. ( when user support etc ) to know about one key is coming from enclave ( or hardware module )

protocol/20250203-webauthn-credential-support.md
bluesign: nit: I think `server side` is confusing term here for people. For traditional webauthn server side means RP, but I think here we use server side as Flow network.

protocol/20250203-webauthn-credential-support.md
tarakby: The reason it can be a limitation is the scenario where you'd like to export your private credential to a different context, and would like to use it to sign transactions (outside the webauthn context). I think the platform shouldn't prevent you from using your key in different contexts.
Now the wallet and tools you use may want to prevent you from exporting your key in the first place (to protect your key, along with other protections), and that's understandable, but I think the native platform (FVM and access API) should be as agnostic as possible, and support all cases. The platform can't detect private key thefts anyway.

This FLIP is written from the platform point of view, the account key has been so far agnostic to how the private key is used (self-custody, non-self custody, hardware wallet, software..). Adding info about where the key comes from is indeed important to the user, but I think storing it on-chain breaks the notion of a low level generic account key. 
That's why the proposal attaches the context/scheme info on the transaction signature and not the on-chain account key.
That's an opinion and it's debatable of course.

protocol/20250203-webauthn-credential-support.md
tarakby: > do you mean like without authenticator

yes a wallet without an authenticator, that emulates webauthn signatures. 

> so we need to reject it on AN level ( transaction should not be part of the chain at all ) even though maybe we can also prove it is not possible at all

Exactly, it's not possible to detect such transactions, unless we force all transactions to also include the used `rpId` so that FVM can verify the hash pre-image.

protocol/20250203-webauthn-credential-support.md
bluesign: actually this makes sense, I didn't think from this perspective before.

protocol/20250203-webauthn-credential-support.md
tarakby: You're right, I did remove all the point commented above, and other "server side" mentions. The [terminology](https://github.com/onflow/flips/blob/tarak/webauthn/protocol/20250203-webauthn-credential-support.md#terminology) section clarifies what's the relying party server in the case of Flow. 

There are still a few mentions of "server" that are used in the traditional webauthn context, but please let me know of the line # if I missed some in the Flow context.

protocol/20250203-webauthn-credential-support.md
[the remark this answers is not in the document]
tarakby: Good question about the random challenge. 

First it's important to recall that we don't want to implement the same webuathn client-server authentication. Webauthn is meant for servers to make sure they are talking to the rightful client, so that they can establish a secure connection and allow the client to access further services on that channel. There are no such client-server authenticated channels in blockchains, there are transactions that we want wallets to sign and the Flow chain to verify (the chain acts as the server). It's an atomic action, there is no back and forth between the wallet and the chain.

In Flow, we're only going to borrow SOME existing functions from the webauthn specs to implement our case of transaction signing/verification with webauthn credentials. We definitely don't need all the functions from webauthn specs (that's why the FLIP title is webauth credentials support only and not the full webauthn support). 

The random challenge role in webauthn client-server is clear, as @tombeckenham pointed out. It prevent copying and re-submitting the same answer again to get access to online services. 
Are replay attacks relevant in the Flow transaction context? Yes. Signed transactions could be copied and resubmitted to authorize the same on-chain action again. 
Flow already addresses this risk using sequence numbers (linked in the FLIP). That's a counter attached to the account key on-chain. It increases on-chain for every successful transaction verified by the key. Signers must specify the next expected sequence number and must sign it as part of the transaction payload. If the transaction is replayed, it would have an outdated sequence number and will be rejected on-chain. Given there is already a replay attack mitigation in place, Flow does not need the random challenge method, that's why we're proposing to use the challenge for a different purpose (not the original webauthn purpose), which is to encapsulate the action authorized by the signature, i.e the transaction payload.

Bonus:
Why in Flow we didn't use the same challenge-response mechanism for replay attacks instead of sequence numbers? The challenge has to remain confidential .Server-clients use a secure channel to communicate, while interactions with a chain are all public.
Moreover, the challenge-response mechanism is interactive, it involves a dialogue between the chain and clients for every transaction, which is expensive and hard to implement. Sending transactions is better done in a non-interactive way.

protocol/20250203-webauthn-credential-support.md
```diff
@@ -0,0 +1,326 @@
...
+    - The relying party ID `rpID` is a string set by the wallet to identify the Flow credentials. This is a constant defined by the wallet for all Flow usage (for example `"FLOW-WEBAUTHN-MYWALLET-V0.0"`). Note that further map indices like [user handles](https://www.w3.org/TR/webauthn-3/#user-handle) and [credential IDs](https://www.w3.org/TR/webauthn-3/#credential-id) would allow the look up of the Flow private credential in the authenticator internal map.
+    - Unlike the original WebAuthn registration case, the `challenge` is not sent by the server to the client. The wallet sets `challenge` to any constant string that should be stored temporarily during till receiving the authenticator response. It is not necessary to use a random challenge.
```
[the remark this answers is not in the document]
tarakby: The other thread https://github.com/onflow/flips/pull/320#discussion_r1978449082 addresses the same point about the random challenge. I answered on the other thread so I will resolve this thread to keep the challenge-related discussions in one place.

protocol/20250203-webauthn-credential-support.md
tarakby: > If it was stored locally, it'd defeat one of the key purposes of passkeys - device loss.

I believe Apple/Google device-sharing and recovery of passkeys don't only back up the private keys, they should also back up the way to look up private keys (the mapping credential IDs or other --> keys). So in case of device loss, there shouldn't be a problem.

> It would be a lot nicer if the full credential was stored on-chain.

If I understood correctly, I don't think storing the credential ID on-chain solves the issues for wallets that do not wish to store data locally: To look up the credential ID on-chain, the wallet needs to know the address. The wallet then needs the public key to query the address (via an off-chain service). However, the wallet needs the credential ID to query the public key from the authenticator (via @bluesign's key recover method).

If the wallet design uses credential IDs to query signatures, I think a wallet service will need to store some data off-chain (locally or on its server). 
The other way would be to rely on the RpID only. It seems that authenticators can look up private keys under a given rpID.

tarakby: > as assertion ( credentials.get ) will give you signature can't you just extract public key from those ?

Yes you can do that with ECDSA on any elliptic curve. Without `v` you just have 2 possible public keys as you mentioned. In some curves you can have 4 possible keys (though not the ones supported by Flow)

protocol/20250203-webauthn-credential-support.md
tarakby: > is this valid rpID ? 

Good point! That's not a valid rp ID. I made tests to confirm that browsers rejects arbitrary strings as rp ID. 
It needs to be at least a domain suffix and it has to match the current [web origin](https://html.spec.whatwg.org/multipage/browsers.html#origin) on the client (btw if a developer doesn't specify the rpID in `navigator.credentials.get`, it is read by the API from the origin).

I'm wondering what is the origin in the case of a wallet mobile app, or a wallet browser extension ? 🤔 
Whatever it is, I think it will be different for each client and can't be a Flow-wide constant.

protocol/20250203-webauthn-credential-support.md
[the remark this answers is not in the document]
tarakby: @tombeckenham, I think the details of the wallet registration depends on each wallet implementation. The FLIP should be written in a generic way to support any wallet integration. The current writeup calls "client" all the off-chain components of the wallet, including the user's device and the wallet servers. 
From the platform point of view, the off-chain "client" components would generate a new key pair, extract the public key and submit a transaction to the chain to register the new account key.

Now in the specific case of the FRW, I am not sure there is need to involve the challenge. I think the FRW server can receive the public key from the user's device and registers it to the chain.

protocol/20250203-webauthn-credential-support.md
bluesign: > Whatever it is, I think it will be different for each client and can't be a Flow-wide constant.

yeah if we want native support and decentralized, different rpId is way to go 
if we want control, I think it can be centralized via discovery. 

As we have to trust wallet 100% in any case ( currently this is the setup except ledger ) different rpId makes more sense. 

if we want to make passkey as secure as ledger, I think than we need RP to be something like `passkey.flow.com` 

Problem comes from 2 intend model we have flow instead of single intend of webauthn.

tarakby: > Problem comes from 2 intend model we have for flow instead of single intend of webauthn ( login vs login & sign )

Can you please explain further the "login & sign" case?
From the Flow platform perspective (on-chain perspective), it's only the user signing case right? Maybe the wallet server would like to make a login step with a user's device, but that's an off-chain step that does not involve the platform. 

> if we want to make passkey as secure as ledger, I think than we need RP to be something like passkey.flow.com

I also would like to make sure I understand your point here too, as in what is the risk of using different RP IDs?

bluesign: In the case user trusting the wallet, there is no issue.

Now I am looking at from user perspective ( arguably a bit paranoid one ): 

let's say I have a wallet ( wallet.dnz.dev ) using webauthn and implementing FCL compatible wallet. 

multi RpId case: 

- User clicks sign-in 
- I sent them a challenge 1 ( which they don't see ) they see a login popup with username, confirm with biometrics and login 
- I show them their account etc, later they want to sign a transaction, they have the exact same flow from UX perspective. 

So technically I can send them a transaction hash as initial challenge 1 and they would not know. 

single RpId case: 

- User clicks sign-in 
- I redirect them to wallet.flow.com , where wallet.flow.com gives them a challenge, they sign, and wallet.flow.com forwards me the signed response ( a bit like oauth ) 
- when they want to sign transaction, I send transaction data to wallet.flow.com , they build and show transaction to user , user approves via biometrics, flow forwards me the signature. 


Single rpId case seems a bit more secure but I think unnecessary as we already have 100% trust relationship between wallet and user

tarakby: Thanks for the clarifications!

But even if we allow multiple RP IDs on-chain, wallets (that want to be more transparent with paranoid users) can always use the trusted centralized service "wallet.flow.com" to validate signing transactions. It is just not enforced for all wallets.
Moreover, it seems odd for FVM to hardcode the RP ID of "wallet.flow.com" while it is a centralized server not run on-chain. The server (even when run by the foundation) does not present the same trust guarantees of the multi-node chain. 

Anyway I think that these cases point out two problems that the FLIP does not intend to solve:
 - how do users trust wallets that they are signing what they think they are signing (when there is no user-wallet trust)
 - how do wallets authenticate their users (the login step in the examples)
 
The second point in particular has been coming back in the discussions. I think it's important to clarify that the FLIP does not describe how a user authenticates into the wallet server (using webauthn, passwords or other). Maybe my write-up isn't clear in this sense. The FLIP only wants to describe the transaction signing part, which can be unrelated to how the login is made in the first place.

tarakby: We seem to agree about not enforcing FVM to hardcode a single RP ID, so I will keep that in the proposal, till we chat about the topic (in the next protocol working group)

bluesign: yeah I think multiple rpID is much better solution, actually what I was trying to show was ( as defender of multiple TP ) if we make single RP, then we need to be in the middle. ( which is not a good idea imo as you pointed out ) 

With multiple rpID,  as it has to come with assumption that trust between wallet - user should be 100%, there is no problem wallets also can use webauthn for login. ( extract public key etc )

janezpodhostnik:
Great FLIP!

One consideration I had is that it looks like the account keys are bound to the wallet that facilitated their creation.

This means that if I as a user create a Flow account with wallet A and wallet A goes out of business without me noticing, I can't (easily) continue using the same Flow account with some other wallet B.

Is there something we can do to address this?
- can we remove wallet specifics from key registration?
- can we offer a recovery tool that allows using keys from wallet A even if it no longer exist?
- should we enforce that the user must have an additional backup key on their account (no ideal)

tarakby:
Good question @janezpodhostnik!

The case of a service going out of business isn't an issue in traditional webauthn and is not addressed by the specs. I believe the reason is that if a service is down, there is no need to authenticate into it, and there is no need for its passkey anymore. 
In the wallet case, this is not true and the passkey is still needed to access the account. 

> can we remove wallet specifics from key registration?

Authenticators tie the private key to the relying party unfortunately, per the webauthn specs.
> can we offer a recovery tool that allows using keys from wallet A even if it no longer exist?

I believe that the tool would need to simulate/recreate A's relying party server with the same domain so that A's passkey can be accessed on the client. @tombeckenham and @lmcmz may know better.

I can add this as a risk in the FLIP.

protocol/20250203-webauthn-credential-support.md
```diff
@@ -0,0 +1,345 @@
+---
+status: Proposed
+flip: 264
+authors: Tarak Ben Youssef (tarak.benyoussef@flowfoundation.org)
+sponsor: Janez Podhostnik (janez.podhostnik@flowfoundation.org)
+updated: 2024-27-03
```
tim-barry, line 6:
```suggestion
updated: 2025-03-27
```
YYYY-MM-DD and use current year

bluesign:
> I believe that the tool would need to simulate/recreate A's relying party server with the same domain so that A's passkey can be accessed on the client. 

I think this is pretty hard with modern web. ( needs DNS spoofing , certificate spoofing  etc ) First one is easy, but certificate would be big pain ( don't know even possible ) 

I think backup is the only sane option tbh.

janezpodhostnik (approved):

PR #333, opened 2025-06-20
Adding FLIP 332:Enforcing Domain-Based Networking Addresses for Nodes

vishalchangrani:
Flip Issue: https://github.com/onflow/flips/issues/332

### DIFF
--- a/protocol/20250619-network-address-validation.md
+++ b/protocol/20250619-network-address-validation.md
@@ -32,6 +32,8 @@ The FLIP proposes to:
 
 To register a node on the Flow network, an operator submits a [node registration transaction](https://developers.flow.com/networks/staking/staking-collection#register-stakers), which includes the networking address of the node. This address is how other nodes establish peer connections with this node.
 
+The networking address has two parts - the TCP/IP address or domain name of the node and the port used by the node for peer-to-peer communication.q
+
 While [best practices](https://developers.flow.com/networks/node-ops/node-operation/node-bootstrap#generate-your-node-keys) recommend using a fully qualified domain name (FQDN) and including the port, the staking contract does not enforce these recommendations.
 
 ### Drawbacks of IP-Based Addresses
--- a/protocol/20250619-network-address-validation.md
+++ b/protocol/20250619-network-address-validation.md
@@ -1,5 +1,5 @@
 ---
-status: draft 
+status: Proposed
 flip: 332
 authors: Vishal Changrani
 sponsor:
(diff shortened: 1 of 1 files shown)


vishalchangrani:
> This is very convenient to have for sure, but think also need to think security side of things; slashing, routing etc. But maybe as we probably allow and suggest domain names already, those considerations are already done.

Hi @bluesign - do you mean slashing if someone stakes with an IP address?

bluesign:
hmm we are adding another layer of direction, maybe that can lead to some slashing/routing problems. 

imagine you are sending message to me ( node A to node B ) if I set my network address as your network address, maybe it can lead to some problems for example.

protocol/20250619-network-address-validation.md
```diff
@@ -0,0 +1,78 @@
+---
+status: draft 
+flip: NNN (set to the issue number)
+authors: My Name (me@example.org), AN Other (you@example.org) 
+sponsor: AN Expert (core-contributor@example.org) 
+updated: YYYY-MM-DD
```
joshuahannan, line 6: Still have the placeholders here

bjartek:
Hou should probably mention what port that this is. IE what role it has for the node.

bluesign:
> Just FYI, we already have a construct in the contract make sure that once a networking address has been used, it cannot be used again, so we don't need to worry about people setting their network address to the same as someone else

Yeah that was my point: like dnz.dev today can point to 1.1.1.1, but tomorrow I can point it to where dappernode.com pointing etc

vishalchangrani:
> Hou should probably mention what port that this is. IE what role it has for the node.

added a line about the port [here](https://github.com/onflow/flips/pull/333/files#diff-f596cdf1727c68abeeb9f065a94d5649a6c489f3ff748d86abe9c3cce8593a99R35) under Background.

vishalchangrani:
> > Just FYI, we already have a construct in the contract make sure that once a networking address has been used, it cannot be used again, so we don't need to worry about people setting their network address to the same as someone else
> 
> Yeah that was my point: like dnz.dev today can point to 1.1.1.1, but tomorrow I can point it to where dappernode.com pointing etc
> 
> but probably it is not problematic yet, as we have verified nodes for roles and trust.

That is a fair point but that is part of the larger protocol autonomy initiative to make the network more secure against byzantine nodes.

protocol/20250619-network-address-validation.md
vishalchangrani: fixed it.

Issue #338, opened 2025-07-31
FLIP 338: Flow Actions: Composable Standards for Protocols


sisyphusSmiling:
This proposal should introduce Flow Actions with DeFiActions (DFA) as the first implementation. DeFi Actions is a suite of standardized Cadence interfaces that enable developers to compose complex DeFi workflows by connecting small, reusable components. DFA will provide a "money LEGO" framework where each component performs a single DeFi operation (deposit, withdraw, swap, price lookup) while maintaining composability with other components to create sophisticated financial strategies. More broadly, Flow Actions will be a set of standards to create composable workflows for common operations by small individual components, acting as an abstraction layer atop new and existing protocols in the interest of enhancing ecosystem composability.

PR #339, opened 2025-07-31
FLIP 338: Flow Actions: Composable Standards for Protocols

sisyphusSmiling:
This FLIP is tracked by: #338


sisyphusSmiling:
Thanks for the initial feedback @bluesign! I don't remember seeing your FLIX alternative @bluesign, but I'm interested in hearing more about it.

> To be honest I would love to see DefiActions as a subset of Actions framework, I believe it would be more meaningful.

What do you mean "as a subset of Actions framework"? Is that your FLIX alternative or are you speaking to a broader potential standard?

bluesign:
> In fact, it's a hope that the mental model of an "action" as a composable, tightly scoped component with single responsibility enhances composable architectures

This is great idea, but I think the main issue I have here is Action and Interface used as interchangeable here. I feel action should be more generic to be honest. It is not possible to predict everything, and things can go ugly fast. 

Interfaces are lacking a lot to cover Actions, then you will have a lot of boilerplate. ( other structs interfaces implementing said interface etc )

sisyphusSmiling:
@bjartek sorry, not sure I follow. Correct me if I'm wrong - I think you're asking if all actions regardless of purpose can share a common interface. Tbh I'm not sure what that could be given the variety of use cases.

The idea of an "action" is more philosophical than anything - a building block that accomplishes some task with a tight scope of responsibility and graceful failure so that any given action can be interchanged as blocks in larger, more complex workflows.

Fundamentally, any given "action" may do very different things from each other - a price oracle fetches a price while a swapper swaps tokens and a sink accepts deposits. However, when interacting with two of the same connectors, you don't have to concern yourself with the underlying implementation. Without some degree of assurance that every different "action" connector will share at least some functionality, I'm not sure how a super interface comes into play.

This is complicated by the amount of abstraction that can be packaged under a single connector. One `Sink` may simply deposit tokens to some underlying Vault, while another could be funnel depositing tokens to a complex DeFi workflow (e.g. adding liquidity into an LP position; adding collateral to a loan & pushing any additional borrowable value to a yield-bearing strategy).

That's to say that even the initialization patterns between the same action could be very different. However, importantly using them can be simplified - I have tokens to deposit, I stick them in a `Sink`- I don't worry about where they go, providing a lot of flexibility in the logic beneath that surface.

Do you have a shared super interface in mind?

bluesign:
@sisyphusSmiling I think what @bjartek says is pretty similar to what I was saying.

Basically Action should just provide "introspection" on some superior interface. Rest is ( where defined here as action ) is actually optional. ( Even wrong without a superior interface )

This usually leads to assignable input/output connectors and generic execute method. More like an invocation pattern.

So in the end you can understand and execute action over a single basic interface. This is a basic requirement in my opinion.

bjartek:
We are currently brainstorming how to visualize and enhance actions in flowscan. 

We can analyze the source of a tx and find all vars that are DefiActions. But what about when there are other actions? Marketplace actions, nft actions etc?

The defining sink source et all structs live in DefiAcfiond not in Action. 

Also how can i in code know more metadata about an action? Views could work or something.

Does that help make this clearer @sisyphusSmiling

bjartek:
Another question this a bit more detail focused

```
 // Derive the path keys from the token types
    let flowKey = SwapConfig.SliceTokenTypeIdentifierFromVaultType(vaultTypeIdentifier: Type<@FlowToken.Vault>().identifier)
    let usdcFlowKey = SwapConfig.SliceTokenTypeIdentifierFromVaultType(vaultTypeIdentifier: Type<@USDCFlow.Vault>().identifier)

    // Minimal path Flow -> USDCFlow
    let swapper = IncrementFiSwapConnectors.Swapper(
      path: [
        flowKey,
        usdcFlowKey
      ],
      inVault: Type<@FlowToken.Vault>(),
      outVault: Type<@USDCFlow.Vault>(),
      uniqueID: nil
    )
    
    ```
    
To me this is the wrong level of abstraction. Why do i hae to know about this path and the swap config stuff in here? I know there are increment fi swaps that can have complex paths, but has anybody ever used them. And if that is the case this could be a MultiPathSwapper.

IMHO this code should be

```
    // Minimal path Flow -> USDCFlow
    let swapper = IncrementFiSwapConnectors.Swapper(
      inVault: Type<@FlowToken.Vault>(),
      outVault: Type<@USDCFlow.Vault>(),
      uniqueID: nil
    )
    ```

bjartek:
Sorry for beeing late to the party on actions, but I have been/am sick and have not had much time. 

I see that UniqueID is sent into structs and not when structs are used. I am not sure I like this implementation detail. This makes it very hard to reuse any action. I cannot now make a registry of predefined actions and fetch reuse them to make my code clean. 

Is it possible to look at this detail this late?

PR #342, opened 2025-08-15
FLIP 341: Add 128-bit fixed-point types to Cadence

SupunS:
This FLIP is tracked by: https://github.com/onflow/flips/issues/341

### DIFF
--- a/cadence/20250815-128-bit-fixed-point-types.md
+++ b/cadence/20250815-128-bit-fixed-point-types.md
@@ -1,9 +1,9 @@
 ---
-status: proposed 
+status: accepted 
 flip: 341
 authors: Supun Setunga (supun.setunga@flowfoundation.org)
 sponsor: Dieter Shirley (dete@flowfoundation.com)
-updated: 2025-08-15
+updated: 2025-08-29
 ---
 
 # FLIP 341: Add 128-bit Fixed-point Types to Cadence
--- a/cadence/20250815-128-bit-fixed-point-types.md
+++ b/cadence/20250815-128-bit-fixed-point-types.md
@@ -32,8 +32,13 @@ Developers can use fixed-point values in Cadence for high-precision arithmetics,
 
 Add `Fix128` and `UFix128` to Cadence.
 
-The proposal is to use a scale of 24,  and hence a scaling factor of 1e-24.
-Thus, the new fixed-point type can accommodate values in the following ranges:
+The proposal is to use a scale of 24, and therefore a scaling factor of 1e-24.
+This scaling factor was chosen because the main objective of adding 128-bit length types is to achieve higher precision,
+rather than being able to represent extremely large numbers.
+Thus, we think having a scaling factor of 24 provides a high-enough precision for fractional values,
+while also leaving large enough upper and lower bounds sufficient for most real-world use cases.
+
+The new fixed-point type can accommodate values in the following ranges:
 - `Fix128`: `-170141183460469.231731687303715884105728` through `170141183460469.231731687303715884105727`
 - `UFix128`: `0.0` through `340282366920938.463463374607431768211455`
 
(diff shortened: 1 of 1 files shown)


vishalchangrani:
lgtm

joshuahannan (approved):

cadence/20250815-128-bit-fixed-point-types.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Design Proposal
+
+Add `Fix128` and `UFix128` to Cadence.
+
+The proposal is to use a scale of 24, and therefore a scaling factor of 1e-24.
+This scaling factor was chosen because the main objective of adding 128-bit length types is to achieve higher precision,
+rather than being able to represent extremely large numbers.
+Thus, we think having a scaling factor of 24 provides a high-enough precision for fractional values,
+while also leaving large enough upper and lower bounds sufficient for most real-world use cases.
```
dete, line 39:
```suggestion
The proposal is to use a scale of 24, and therefore a scaling factor of 1e-24. Other blockchain 
ecosystems (most notably Ethereum) frequently use scale factors (usually called "decimals", per 
ERC-20) of 18 (used by ETH itself) 24 (used for internal calculations in MakerDAO) and 27 (used 
internally by Compound and Aave for interest calculations). Although 27 is the most commonly used 
value, the value of 24 has a precedent in Maker, and leads to a range that is close to, but 
slightly larger than the range of Fix64 and UFix64. The allows all UFix64 and Fix64 values to be 
converted to the equivalent 128-bit types without any loss of precision or range. (27 decimals 
wouldn't have allowed the full range of Fix64 and UFix64 to fit in the new types, and using 18 
decimals has been deemed to be insufficient by a number of defi protocols.) A scaling factor of 24 
provides a very high precision for fractional values, while also leaving large enough upper and 
lower bounds sufficient for most real-world use cases (which typically will involve internal 
calculations that convert back to the 64-bit types).
```

SupunS: Thanks for the detailed explanation!

cadence/20250815-128-bit-fixed-point-types.md
```diff
@@ -0,0 +1,93 @@
...
+
+Other blockchain ecosystems (most notably Ethereum) frequently use scale factors (usually called "decimals", per ERC-20)
+of 18 (used by ETH itself), 24 (used for internal calculations in MakerDAO), and 27 (used internally by Compound and Aave
+for interest calculations).
+Although 27 is the most commonly used value, the value of 24 has a precedent in Maker, and leads to a range that is
+close to, but slightly larger than the range of `Fix64` and `UFix64`.
+This allows all `UFix64` and `Fix64` values to be converted to the equivalent 128-bit types without any loss of precision
+or range.
+On the other hand, 27 decimals wouldn't have allowed the full range of `Fix64` and `UFix64` to fit in the new types, 
+and using 18 decimals has been deemed to be insufficient by a number of defi protocols.
```
dete, line 45:
```suggestion
Although 27 is the most commonly used value internally by defi protocols, the value of 24 has a precedent in Maker, and
introduces a type that can hold values in the trillions (extreme, but plausible values for very large financial calculations).
Note that all `UFix64` and `Fix64` values can be converted to the equivalent 128-bit types without any loss of precision or range.
```

I should have done more checking before that last comment, Supun: It turns out that 27 decimals WOULD still fit in 128-bits, but we actually kind of do need a fixed type that can support values over 187 billion...

dete: (The real reason I chose 24 was because I like round numbers... 😅 )

vishalchangrani:
hi folks - Please add any comments or questions you have about this FLIP. If not, this FLIP will be marked as `Accepted` by EOD tomorrow (8/29/2025).
Thank you

PR #344, opened 2025-08-28
FLIP 343: Fix numeric type rounding inconsistency

SupunS:


turbolent (approved):
+1 from me

Though this is a backward incompatible change, the current behaviour is neither documented nor intended. I think it is more important to make the behaviour consistent than to keep backward compatibility for undefined behaviour.

Issue #348, opened 2025-10-14
FLIP 348: EVM scheduled transactions


devbugging:
Flow's Cadence environment supports scheduled transactions through the `FlowTransactionScheduler` contract, allowing smart contracts to autonomously execute logic at specified future times. 

However, Solidity contracts deployed in Flow EVM cannot currently access this scheduling capability. This proposal introduces a bridge design to expose Flow's native scheduling feature to EVM contracts through:
- **EVM Scheduler (Cadence)**: A bridge contract that integrates with the existing `FlowTransactionScheduler`, managing gas limits and fund withdrawal from EVM to Cadence
- **EVM Scheduler Proxy (Solidity)**: An EVM-facing contract that provides the scheduling API, stores transaction metadata, emits events for observability, and handles cancellation
- **Cadence Arch Functions**: New precompile functions (schedule, estimate) to facilitate cross-environment calls
- **Standard Handler Interface**: `IScheduledTransactionHandler` defining a single `executeTransaction(uint64 id, bytes calldata data)` function that all handler contracts must implement

Key Benefits:
- Reuses existing Flow scheduler infrastructure (single source of truth)
- Non-intrusive design with no changes to existing contracts
- Enables autonomous DeFi and automation use cases without off-chain keepers
- Lazy execution with no overhead when no EVM transactions are scheduled

PR #349, opened 2025-10-14
FLIP 348: EVM Scheduled Transactions

devbugging:
Issue: #348

### DIFF
--- a/protocol/20251014-evm-scheduled-transactions.md
+++ b/protocol/20251014-evm-scheduled-transactions.md
@@ -152,9 +152,9 @@ The implementation of the **EVM Scheduler** introduces new components that integ
             uint64 gasLimit
         ) external view returns (uint64 fee, uint64 timestamp, bool ok);
 
-    		/// @notice Cancel existing scheduled transaction
-    		/// @param id ID of the scheduled transaction
-    		function cancel(uint64 id) external;
+    	/// @notice Cancel existing scheduled transaction, msg.sender must match the caller of schedule()
+    	/// @param id ID of the scheduled transaction
+    	function cancel(uint64 id) external;
 
         /// @notice Get the status of a transaction
         /// @param id Transaction ID


protocol/20251014-evm-scheduled-transactions.md
```diff
@@ -0,0 +1,465 @@
...
+    
+
+---
+
+## Scheduling Workflow
+
+Scheduling involves coordination between all the above components and should be initiated from EOA using the EVM Scheduler proxy Solidity contract.
+
+1. **EOA**
+    1. Submits an EVM transaction that calls the EVM Scheduler Proxy with all required arguments as well as the amount needed to pay for execution.
```
joshuahannan: Maybe add a first step here for EOA having to deploy a contract that contains the logic to execute for the scheduled transaction

protocol/20251014-evm-scheduled-transactions.md
```diff
@@ -0,0 +1,465 @@
...
+    4. Calls the handler's `executeTransaction(uint64 id, bytes cdata)` function with the transaction ID and stored data.
+    5. If the execution fails it stores failed status for the ID.
+    6. Emits an executed event.
+    7. Garbage collects the transaction data. 
+
+## Cancellation Workflow
+
+Cancellation follows a circuit-breaker pattern. Execution is still initiated on the Cadence side, but the EVM Scheduler Proxy drops it, preventing the EVM callback from running. 
+
+This effectively cancels execution on the EVM side but without a refund, since Cadence has already executed its portion. Because Cadence Arch cannot authorize calls or forward environment data (such as `msg.sender`), it cannot enforce authorization for cancellation at the Cadence side. As a result, a scheduled transaction cannot be fully canceled.
```
bluesign: > Because Cadence Arch cannot authorize calls or forward environment data (such as `msg.sender`), it cannot enforce authorization for cancellation at the Cadence side.

that does this mean, can you explain a bit on it

protocol/20251014-evm-scheduled-transactions.md
```diff
@@ -0,0 +1,465 @@
...
+
+To guarantee execution integrity, the EVM Scheduler Proxy must consistently maintain the mapping between transaction IDs and their metadata. IDs must only be created through the proxy and registered only by the COA, ensuring that no transaction can be executed unless it was scheduled through the intended flow.
+
+# Alternatives
+
+There were multiple alternatives considered, some of which proved to be unviable, most elegant ones due to limitation of Cadence Arch authorization mechanism. The Cadence Arch is a precompile that must follow the [EVM precompile interface](https://github.com/ethereum/go-ethereum/blob/1487a8577d1566497e161a04f8cee3204d4b3d36/core/vm/contracts.go#L52) and unless we fork the EVM we can not introduce more arguments to it. Forking of EVM is not acceptable at this point.
```
bluesign: ah this makes sense. I think it may be possible without forking EVM (as `Call` will precede with transfer from `msg.sender` only issue maybe delegatecall there but I don't think it is too important in this context )

protocol/20251014-evm-scheduled-transactions.md
devbugging: Yes, this is a bit confusing. 

So, one idea was, that we could cancel directly from EVM calling the Cancel on the Cadence Transaction Scheduler contract via Cadence Arch. The problem with this is that you have to somehow authorize the call, meaning sine Cadence Arch are currently callable by anyone, this also mean anyone could cancel your scheduled transaction by ID. So we had an idea for Cadence Arch to be able to authorize the call, by including the msg.sender as the parameter to cancel, so we could check that the account calling cancel is the same as the one that called schedule, but this unfortunately is not possible because the interface definition for the EVM Precompiles doesn't include the msg data, so we would have to fork it, which I don't think justifies it. 

That way, we can just on the EVM side drop the execution, during execution if it was canceled.

protocol/20251014-evm-scheduled-transactions.md
devbugging: What do you mean? can you elaborate

PR #353, opened 2025-12-04
FLIP 353: Automatic Restaking

### LINKED ISSUE #354: FLIP 353: Automatic Restaking
Issue to track FLIP 353: Automatic Restaking

joshuahannan:
Closes https://github.com/onflow/flips/issues/354

Adds a FLIP that proposes automatically restaking all stakers rewards every epoch

### DIFF
--- a/governance/20251205-enable-automatic-restaking.md
+++ b/governance/20251205-enable-automatic-restaking.md
@@ -47,9 +47,19 @@ In addition to being a convenience for users, this will also ensure that there i
 
 ### Opt-in/Opt-out automatic restaking
 
-One option considered was making this feature opt-in, but this was decided against because the changes required for it would be cumbersome because of Cadence's upgrade restrictions with fairly small benefit. Also, because users still have a week to withdraw their rewards, they still have the opportunity to "opt-out" every week by withdrawing their rewards if they want.
+One option considered was making this feature opt-in, so users could choose whether they want this. 
 
-Another option could be to make this automatic by default, but allow users to opt-out by indicating that they don't want their rewards to be automatically restaked every week. This would still have most of the benefits described above, but would allow some users that maybe have unique tax considerations to have control over their rewards. It is also being considered as an option.
+Another option could be to make this automatic by default, but allow users to opt-out.
+
+Both of these options would still have most of the benefits described above, but would allow some users that maybe have unique tax considerations to have control over what is done with their rewards. It is also being considered as an option.
+
+One potential problem with allowing users to choose is that Cadence's upgrade restrictions make including this quite cumbersome. We would have to create a separate record from the existing staking records that are accessed separately from the other records, which, with the large number of existing stakers, could significantly affect the performance and cost of existing epoch operations. We would like to avoid this for the protocol and users if possible.
+
+Additionally, users still have a week to withdraw their rewards, they still have the opportunity to "opt-out" every week by withdrawing their rewards if they want.
+
+### Longer Period between Restakes
+
+If one week feels like it is not enough time to give users to decide whether or not to restake or withdraw, it is possible to extend this to any arbitrary number of weeks between restaking. If a month or more is preferable, this could be included and would not affect performance and cost like making the feature opt-in or opt-out would.
 
 ### Scheduled Transactions for Restaking
 
--- a/governance/20251205-enable-automatic-restaking.md
+++ b/governance/20251205-enable-automatic-restaking.md
@@ -1,11 +1,11 @@
 ---
 status: Proposed
-flip: GOV-30
+flip: GOV-353
 authors: Joshua Hannan (joshua.hannan@flowfoundation.org)
 editors:
 ---
 
-# FLIP GOV-30: Enable Automatic Restaking of Staking Rewards
+# FLIP GOV-353: Enable Automatic Restaking of Staking Rewards
 
 ## Abstract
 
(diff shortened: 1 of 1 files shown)


governance/20251205-enable-automatic-restaking.md
```diff
@@ -0,0 +1,40 @@
...
+
+# FLIP GOV-30: Enable Automatic Restaking of Staking Rewards
+
+## Abstract
+
+Enable core contract feature to automatically restake rewarded tokens for nodes and delegators each epoch instead of requiring stakers to do it manually.
+
+## Background
+
+Rewards to node operators and delegators are paid at the end of every epoch to each staker's reward bucket. Currently, these rewards will sit there until the user manually does something with them, either withdrawing or restaking them. This can be cumbersome to most stakers who simply want to continue staking their tokens every week. Many users over the years have asked for functionality to automatically restake their rewards for them.
```
AlexHentschel: I think it would significantly improve the clarity of this document if you explain the "buckets" you are mentioning in the next section. I would suggest to explain the buckets at the end of the background section. For the _start_ of the background section, I would keep it more general 

```suggestion
## Background

Rewards to node operators and delegators are paid at the end of every epoch. Currently, these rewards will sit there until the user manually does something with them, either withdrawing or restaking them. This can be cumbersome to most stakers who simply want to continue staking their tokens every week. Many users over the years have asked for functionality to automatically restake their rewards for them.
```
[continue the background section as is:]
> ...  enabling this feature following community discussion and voting.

and then we explain the buckets.
```
Under the hood, the staking smart contract maintains three "buckets" of tokens for each staker and delegator. 
... <explain buckets and introduce specific names for the relevant buckets, use those names in the later section>
```

governance/20251205-enable-automatic-restaking.md
```diff
@@ -0,0 +1,40 @@
...
+
+Flow originally decided to not include this feature out of an abundance of caution about crypto regulations. Given recent developments in the past five years related to crypto regulations, and especially considering that most other Proof-of-Stake blockchains support this feature with no negative repercussions, the Flow foundation is proposing enabling this feature following community discussion and voting.
+
+## Proposal
+
+When tokens are moved between buckets at the end of every epoch, for any given staker, transfer all their rewarded tokens to their staked bucket and update all the relevant staking metadata for that user.
+
+Because rewards are paid every epoch after tokens are moved between buckets, this means that staking rewards that are paid to users will still stay in their rewards bucket after they are paid every week. This will allow users who want to withdraw rewards every week instead of restaking them to still be able to do without having to wait for the 1-2 week unstaking period. We believe that this will be the least disruptive to any existing stakers who maybe doing something different than restaking every week.
```
AlexHentschel: I had to read this section 3 times before I understood, and I would guess that I have pretty good background information 😅 . I think you can significantly improve clarity by emphasizing key aspects and spelling it out more explicitly step by step. 

```suggestion
It is important to emphasize that rewards are paid every epoch _after_ tokens are moved between buckets. This means that staking rewards that are paid to users will still **stay in their rewards bucket for one week** after they have been paid. This means:
* Reward tokens that the staker's / delegator's does not deposit into their account within one week will automatically be staked. Once the tokens are staked, they are subject to the standard 1-2 week unbonding period. 
* Users can withdraw (a portion of) rewards immediately within one week after the tokens have been rewarded. This moves the tokens out of the rewards bucket. Only the tokens remaining in the rewards bucket will be automatically restaked at the next epoch switchover. 

We believe that this will be the least disruptive to any existing stakers who maybe doing something different than restaking every week.
```

AlexHentschel (approved):

joshuahannan:
Thanks for the feedback, @AlexHentschel! I addressed your comments and made some other changes to improve clarity

JeffreyDoyle:
I'm unsure if auto-restaking by default is desired by all users, as there are potential tax implications when being awarded tokens. Have we considered an opt-in approach?

chasefleming:
Yeah, I agree with @JeffreyDoyle. This should probably be a flag. Not everyone checks weekly, and they’ll have to unstake and wait weeks if they miss the window. Even if it’s not a taxable event, some users wait to move funds so they can make fewer transactions.

joshuahannan:
I'm definitely open to changing it if there is enough pushback, but I would argue that even if you want to make fewer transactions and only withdraw once a month or once every few months or something like that, this system is still better because instead of just sitting in your rewards bucket not accruing any additional rewards, your tokens were restaked and gave you some additional rewards while you were waiting.

Also, about taxes, I am not a tax expert, so this is not legal advice, but as far as I know, your rewards are not taxable until you withdraw them from the staking protocol, so I believe automatic restaking would likely not be a taxable event, especially since it is something that the protocol would do automatically with no input from the user. But we'll need to talk to some tax experts

briandoyle81:
I'm 1000% for this and agree with a flag, but making it opt out.  I think that the users who would benefit from choosing to defer rewards to defer taxes etc. will find the option they need.

vishalchangrani:
Issue to track this FLIP: https://github.com/onflow/flips/issues/354

bluesign:
IANAL, not tax advice

I think this depends on jurisdiction, some regard "withdraw of reward" as taxable event, others regard "calculation of reward" as taxable event.

But I think also the biggest problem here, it changes your cost basis when restake, in second case. ( Not sure about the first case ) If I had 1000 flow with 1$ each, and staked , when I got 10 flow reward  ( let's say flow was $0.5) and restake, income tax on $5 and my cost basis will be ~0.9950. So went I unstake & sell 100 flow at $1.5 I should calculate ((1.5-0.9950)*100) for taxes.

Opt in seems safer.

AlexHentschel:
Disclaimer: none of the following is tax advice. 

I chatted with our tax person a while back. She was hesitant to provide a conclusive statement because crypto regulations are still developing and she felt she didn't know all the details yet that factor into the equation. Anyway, she referred me to some documents and explained how the CRA typically approaches the question whether something like staking / unstaking is a taxable event in Canada. TLDR: 

The canadian tax system uses the notion of ["Identical Properties"](https://www.canada.ca/en/revenue-agency/services/forms-publications/publications/it387r2-consolid/archived-meaning-identical-properties.html), which _very_ broadly says that any two assets are considered different kinds, if they for example [quote]:
> do _not_ have the same interests, rights and privileges

This is clearly the case for liquid vs staked tokens (one produces rewards the other doesn't; one needs to be unstaked and only then is transferable, the other is immediately transferable). Therefore, in agreement with my tax advisor, I have been treating any conversion between liquid and staked tokens as a taxable event. 

### From a Canadian tax perspective, the change seems subtle but important:

The detailed means don't matter, fact is that when they are received, reward tokens are transferable and don't produce revenue. Hence, I think they are "identical properties" as liquid tokens (in Canada).

> your rewards are not taxable until you withdraw them from the staking protocol

To the best of my understanding, the canadian tax system considers the point in time when you receive the rewards and _could_ withdraw them as a taxable event, not when they are actually withdrawn.

So receiving the rewards is a taxable event anyway. What changes with this flip is whether I have an _additional_ taxable events every week converting reward tokens to staked tokens.

Status Quo: No taxable events from converting rewarded to staked tokens, unless I explicitly re-stake (taxable event) some of the reward tokens manually. Also, if selling some of the rewarded tokens to cover taxes due on the rewards (to the best of my understanding count as income in canada), one can just use the reward tokens that have accumulated and sell some of those from time to time. 

With the Flip: Unless I withdraw reward tokens manually _every week_, there will be automatically an additional taxable event every week that I need to include in my accounting. This is major additional effort. If I wanted to sell a portion of the reward tokens from time to time to cover taxes due on the rewards, I need to unstake some of the auto-restaked rewards, which is another taxable event. 

### My perspective as a user:

⚠️ **Automatic restaking** would **significantly increase my bookkeeping overhead**. This is **_not_** a **favourable proposal** for my case.

### Recommendation

Can we implement this with an opt-in or opt-out? I understand that the benefit of the current proposal is that it would be uniform for _all_ accounts, i.e. it could be implemented in the staking smart contract.  In contrast, what I am asking for is an account-individual solution. 

However, as it stands right now, for me personally (only speaking as a user, _not_ as a developer of the flow platform), I would rather have no change to the status quo than the proposed "one-size fits all" approach. Of course, the ideal would be a solution where users could pick.

joshuahannan:
I'm coming back to this after the holidays. I can see that there is a lot of pushback about the lack of flexibility, but there is a mix of opinions about it being opt-in or opt-out. So we can either make it opt-in or opt-out. The implementation for each is pretty similar. My vote would be for opt-out. This way, anyone can choose whether or not to participate, but we can still have the benefit of getting all the stale rewarded tokens that haven't been restaked yet into the protocol in the first week for those who haven't opted out. But if enough people think it should be opt-in instead if we want it to be as non-disruptive as possible, I'm open to that too

AlexHentschel:
In my opinion, it needs to be an important consideration what helps the broader Flow ecosystem to gain more attention and market share. The current web3 market environment is very challenging and TVL and tokens staked are important metrics, by which protocols compete for attention. Personally, I disagree with approximating a protocol's soundness and potential future market impact by a few defi-related metrics. But that's the reality. And given that reality, I think Josh made a good point **in favour of opt-out**.

Issue #357, opened 2026-02-09
FLIP 357: Add Comparison Functions to Cadence


turbolent:
- FLIP PR: https://github.com/onflow/flips/pull/358
- Implementation PR: https://github.com/onflow/cadence/pull/4430

PR #358, opened 2026-02-09
FLIP 357: Add Comparison Functions to Cadence

turbolent:
Work towards #357


cadence/20260209-add-comparison-functions.md
```diff
@@ -0,0 +1,225 @@
...
+## Drawbacks
+
+**Limited to Two Arguments**: The functions only accept two arguments. Finding the minimum or maximum
+of more values requires chaining:
+
+```cadence
+let min = minOf(minOf(a, b), c)
+```
+
+A future enhancement could add variadic versions that accept more than two arguments.
```
SupunS, line 147: This would be a backward compatible change too 👍

cadence/20260209-add-comparison-functions.md
```diff
@@ -0,0 +1,225 @@
...
+```
+
+**Pros:**
+- More discoverable through type-specific documentation
+- No naming conflict
+
+**Cons:**
+- More verbose
+- Requires many separate implementations (one per comparable type)
+- Users must remember type-specific methods instead of a single global function
```
SupunS, line 202: Alternatively we could introduce a new interface type, e.g: `Comparable`, and make these static functions of that type. Then there would be only one function.

Also it would allow adding more functionalities/functions in future (I can't think of a concrete example though)

SupunS (approved):
+1

cadence/20260209-add-comparison-functions.md
turbolent: Good idea! Like you mentioned on the PR, we could add a `ComparableStruct` interface (cf. `HashableStruct`). I quite like that

janezpodhostnik (approved):

cadence/20260209-add-comparison-functions.md
jordanschalm: I think it is good for Cadence to provide plenty of utility functions like this, and also to minimize how much of the global namespace it takes up, so +1 for adding these utility functions to a shared interface type.

jordanschalm (approved):

bluesign:
yeah this is much needed. I love it.  ( I would also love `clamp(value, min, max)` )

cadence/20260209-minof-maxof-functions.md
```diff
@@ -0,0 +1,225 @@
...
+or accidentally comparing the wrong variables.
+
+**Type Safety**: The functions work with any comparable type and ensure both arguments have the same type,
+catching type mismatches at compile time.
+
+**Consistency**: Provides a standard way to perform these operations across all Cadence codebases.
+
+## Design Proposal
+
+Add two generic functions to the Cadence standard library that work with any comparable type.
```
joshuahannan: Is it possible to compare two values of different types like a UInt8 and UInt64 or UFix64 and UFix128 do they have to be the same exact type?

joshuahannan (approved):
this will be great to have!

cadence/20260209-minof-maxof-functions.md
turbolent: Comparison is only defined for values of the same type, just like we require the same type for binary operations like `+`. When two values of different types should be compared, they should be first converted to the same type.

cadence/20260209-add-comparison-functions.md
turbolent: Would it also suffice to just move the functions and add future comparison functionality in a new built-in `Comparison` contract? Similar to how we already have built-in `RLP` and `BLS` contracts, and similar to how e.g. JS namespaces this in `Math`, but generalized for all comparable types.

Compared to adding a new `ComparableStruct`, this would be easier/less code to implement, and would also not prohibit eventually introducing a `ComparableStruct` type later.

cadence/20260209-minof-maxof-functions.md
bluesign: I think we can make Number as Comparable. ( not sure if it will be a foot-gun )

cadence/20260209-add-comparison-functions.md
jordanschalm: That seems fine to me. From my perspective as a language user, the difference between "utility functions defined on a built-in interface type" and "utility functions defined on a built-in contract" is pretty minimal, so happy to go with whatever is easier to implement

turbolent: I've updated the proposal in [a8a2a4d](https://github.com/onflow/flips/pull/358/commits/a8a2a4d05c8569941b699c3a9c1c73cef9e2e48a). PTAL

turbolent:
I've updated the proposal in [a8a2a4d](https://github.com/onflow/flips/pull/358/commits/a8a2a4d05c8569941b699c3a9c1c73cef9e2e48a):
- Renamed the functions to `min`/`max`, so they have familiar names, like in most other languages
- Define the functions in the new contract `Comparison`, instead globally. This avoids any naming conflicts and guarantees backward compatibility for existing code

turbolent:
Added a `clamp` function in [c05e5f4](https://github.com/onflow/flips/pull/358/commits/c05e5f4be73dce9823c96c6119330f5238acd735). cc @bluesign

turbolent:
There has been no negative feedback, this FLIP has been accepted

PR #369, opened 2026-06-11
FLIP: Execution effort calibration III

janezpodhostnik:
Proposes an update to Flow's per-kind execution-effort weights — the successor to [FLIP-346: Execution Effort Calibration II](https://github.com/onflow/flips/blob/main/protocol/20251008-execution-effort-2.md).

Recalibrates the weights for FVM and runtime changes since FLIP-346, using a joint non-negative least-squares fit against the live mainnet scrape (3.39M user transactions) and the migrationtestnet synthetic-load scrape (0.22M labelled transactions). The proposed fit lifts mainnet per-transaction R² from 0.15 to 0.80.

FLIP file: `protocol/20260611-execution-effort-3.md`


bluesign:
little strange with set_value and get_value, do you have any insight there?

janezpodhostnik:
Are you referring to the value change of them? @bluesign

bluesign:
Yeah set being cheaper then get vs existing behaviour ( opposite ) 

I think some caching behavior etc can be in play, but if not seems little strange reading being more expensive than writing.

janezpodhostnik:
Whats going on underneath is that setting also does a get, so the cost of a set is actually get+set.

bluesign:
> Whats going on underneath is that setting also does a get, so the cost of a set is actually get+set.

Yeah but now isn't get is much more expensive than set? With new weights. That little confused me 

I understand the point a bit, if there is existing data, set can be like delta ( set-get cost ) , but is it common case? 

Deleting slab -> get(old size)  OK 
Writing new slab - > get (0) + set (data size)  = set(data size) -> (set_cost-get_cost) * data_size -> not OK 
Modifying existing slab -> get(old size) + set(new size)  -> this may be not OK

It feels like we are undermetering new slab writes with this logic somehow. ( putting cost to get instead ) 

Probably nothing so major btw ( as we also block flow for storage fees ) , just at first sight felt little weird.

janezpodhostnik:
I suspect that most of the cost of `set` gets folded into individual atree operations.

`get` is generally more of an isolated operation, where as set leads to calls to other weights. 


I will see if I can run an experiment for this, to make it somewhat more apparent whats going on.

vishalchangrani (approved):

